PR #311, opened 2021-09-02
fix: utilize endpoint ports

### LINKED ISSUE #310: StaticEndpointResolver with custom port does not connect with that port.
## Describe the bug

When setting up a port with the StaticEndpointResolver it does not connect with port passed to the EndPoint

## Expected Behavior

Connects to the server on expected port

## Current Behavior

Does not connect to server on expected port.

```
Exception in thread "main" software.amazon.awssdk.crt.http.HttpException: socket connection refused.
	at software.amazon.awssdk.crt.http.HttpClientConnection.onConnectionAcquired(HttpClientConnection.java:85)
```

## Steps to Reproduce

Set up a local DynamoDB
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.DownloadingAndRunning.html

```
val ddb = DynamoDbClient {
        region = "eu-west-2"

        this.endpointResolver = StaticEndpointResolver(
            Endpoint("localhost", "HTTP", 8000)
        )
    }
```

## Possible Solution

The port is not passed to the Http Engine. (Checked it with a manual implementation of HttpClientEngine)

## Context

I cannot connect to local running DynamoDB
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.DownloadingAndRunning.html

## Your Environment

* AWS Kotlin SDK version used: 0.4.0-alpha
* Operating System and version: macOS 11.5.2

ianbotsf:
## Issue \#

Closes #310 

## Description of changes

Currently, neither the presigning logic nor the service endpoint resolver correctly utilize endpoint ports. This is most readily apparent when using the `StaticEndpointResolver` or a test mock, but may also have negative implications for generated endpoint resolvers in the future.

Companion PR: awslabs/aws-crt-kotlin#28

## Scope

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/aws-runtime/http-client-engine-crt/common/src/aws/sdk/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
+++ b/aws-runtime/http-client-engine-crt/common/src/aws/sdk/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt
@@ -82,7 +82,7 @@ public class CrtHttpEngine(public val config: HttpClientEngineConfig) : HttpClie
     }
 
     private suspend fun getManagerForUri(uri: Uri): HttpClientConnectionManager = mutex.withLock {
-        connManagers.getOrPut(uri.host) {
+        connManagers.getOrPut(uri.authority) {
             HttpClientConnectionManager(options.apply { this.uri = uri }.build())
         }
     }
--- a/aws-runtime/http-client-engine-crt/jvm/test/aws/sdk/kotlin/runtime/http/engine/crt/AsyncStressTest.kt
+++ b/aws-runtime/http-client-engine-crt/jvm/test/aws/sdk/kotlin/runtime/http/engine/crt/AsyncStressTest.kt
@@ -48,6 +48,7 @@ class AsyncStressTest : TestWithLocalServer() {
                 scheme = Protocol.HTTP
                 method = HttpMethod.GET
                 host = testHost
+                port = serverPort
                 path = "/largeResponse"
             }
         }
@@ -82,6 +83,7 @@ class AsyncStressTest : TestWithLocalServer() {
                 scheme = Protocol.HTTP
                 method = HttpMethod.GET
                 host = testHost
+                port = serverPort
                 path = "/largeResponse"
             }
         }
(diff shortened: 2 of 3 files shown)


kggilmer (approved):
LGTM

aajtodd (changes requested):
These changes look ok but I think we're missing part of the puzzle.

We need to ensure each HTTP client engine respects the port as well.

Glancing at the [CrtHttpClientEngine](https://github.com/awslabs/aws-sdk-kotlin/blob/9dcd33450ff654a07d2d21011b7edc68704b85a7/aws-runtime/http-client-engine-crt/common/src/aws/sdk/kotlin/runtime/http/engine/crt/CrtHttpEngine.kt#L85) it looks like we still wouldn't do the right thing because we would be caching client managers by only the hostname (as an aside probably should have named that property `hostname` rather than `host` which usually implies "hostname:port"). 

It [looks like](https://github.com/awslabs/smithy-kotlin/blob/5b4b37e98b55e5aab63df70517a6626b52ff79fe/runtime/protocol/http-client-engines/http-client-engine-ktor/jvm/src/aws/smithy/kotlin/runtime/http/engine/ktor/KtorUtils.kt#L33) the KTOR based engine we started with would be ok.


I found a semi related issue with how we handle ports in the [AsyncStressTest](https://github.com/awslabs/aws-sdk-kotlin/blob/78a671cb72d9ec59c5fce1e430c52c5227b0a75e/aws-runtime/http-client-engine-crt/jvm/test/aws/sdk/kotlin/runtime/http/engine/crt/AsyncStressTest.kt#L50) and [TestWithLocalServer](https://github.com/awslabs/aws-sdk-kotlin/blob/78a671cb72d9ec59c5fce1e430c52c5227b0a75e/aws-runtime/http-client-engine-crt/jvm/test/aws/sdk/kotlin/runtime/http/engine/crt/TestWithLocalServer.kt#L17) work. The host is set as `hostname:port` which only works because of how it gets passed into crt-java later [here](https://github.com/awslabs/aws-crt-kotlin/blob/8daf1da6d8fc122105bb03c35517239541b8aa70/src/jvm/src/aws/sdk/kotlin/crt/http/HttpClientConnectionManagerJVM.kt#L56).

ianbotsf:
> We need to ensure each HTTP client engine respects the port as well.

You're right, didn't go deep enough.

> as an aside probably should have named that property `hostname` rather than `host` which usually implies "hostname:port").

RFC 3986 names the combination of **userinfo**, **host**, and **port** as the [**authority**](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2) of the URI. Logically, is that the right scope for caching connection managers? Or is **userinfo** irrelevant?

aajtodd:
> RFC 3986 names the combination of **userinfo**, **host**, and **port** as the [**authority**](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2) of the URI. Logically, is that the right scope for caching connection managers? Or is **userinfo** irrelevant?

Userinfo for an AWS SDK is _likely_ always going to be irrelevant but that is probably as good as anything to cache on.

aajtodd (approved):

PR #318, opened 2021-09-17
refactor(rt)!: split auth and signing packages; expose a method to sign requests

aajtodd:
## Issue \#

N/A

support PR(s):
* https://github.com/awslabs/smithy-kotlin/pull/479
* https://github.com/awslabs/aws-crt-kotlin/pull/30

## Description of changes

* move signing related things to `signing` subpackage of `auth`
* move credential related things to `credentials` subpackage of `auth`
* expose a top level `sign()` method that can be used to sign HttpRequests independent of middleware
* refactor middleware to re-use `AwsSigningConfig` for it's feature configuration
* expose SigV4A algorithm enum value

## Scope

This is a breaking change for users that import a specific credentials provider. They will need to update their code to the new import location.


ianbotsf (approved):

aws-runtime/auth/common/src/aws/sdk/kotlin/runtime/auth/signing/AwsSigning.kt
```diff
@@ -0,0 +1,49 @@
...
+import aws.sdk.kotlin.crt.auth.signing.AwsSigner
+import aws.sdk.kotlin.runtime.crt.toSignableCrtRequest
+import aws.sdk.kotlin.runtime.crt.update
+import aws.smithy.kotlin.runtime.http.request.HttpRequest
+import aws.smithy.kotlin.runtime.http.request.toBuilder
+
+/**
+ * Container for signed output and signature
+ */
+public data class SigningResult<T>(val output: T, val signature: ByteArray) {
```
kggilmer, line 17: question

why is equals/hashCode impl required for this type?

aws-runtime/auth/common/src/aws/sdk/kotlin/runtime/auth/signing/AwsSigning.kt
```diff
@@ -0,0 +1,49 @@
...
+
+package aws.sdk.kotlin.runtime.auth.signing
+
+import aws.sdk.kotlin.crt.auth.signing.AwsSigner
+import aws.sdk.kotlin.runtime.crt.toSignableCrtRequest
+import aws.sdk.kotlin.runtime.crt.update
+import aws.smithy.kotlin.runtime.http.request.HttpRequest
+import aws.smithy.kotlin.runtime.http.request.toBuilder
+
+/**
```
kggilmer, line 14: suggestion

I assume the intent is that customers may use this type since there is no internal markers.  the KDoc should probably be a bit more thorough in this case, specifying details for parameters for example.

aws-runtime/auth/common/src/aws/sdk/kotlin/runtime/auth/signing/AwsSigning.kt
```diff
@@ -0,0 +1,49 @@
...
+        result = 31 * result + signature.contentHashCode()
+        return result
+    }
+}
+
+/**
+ * Sign [HttpRequest] using the given signing [config]
+ * @return signing result
+ */
+public suspend fun sign(request: HttpRequest, config: AwsSigningConfig): SigningResult<HttpRequest> {
```
kggilmer, line 41: same as above regarding kdoc

aws-runtime/auth/common/src/aws/sdk/kotlin/runtime/auth/signing/Presigner.kt
```diff
@@ -1,10 +1,17 @@
-package aws.sdk.kotlin.runtime.auth
+/*
```
kggilmer, line 1: my bad

aws-runtime/auth/common/src/aws/sdk/kotlin/runtime/auth/signing/AwsSigV4SigningMiddleware.kt
```diff
@@ -18,68 +16,30 @@ import aws.smithy.kotlin.runtime.http.*
...
 
-        override fun create(block: Config.() -> Unit): AwsSigV4SigningMiddleware {
-            val config = Config().apply(block)
+        override fun create(block: AwsSigningConfig.Builder.() -> Unit): AwsSigV4SigningMiddleware {
+            val builder = AwsSigningConfig.Builder().apply(block)
+
+            // region is required when using a standalone signing config but the middleware takes the signing region
+            // (and other settings) from the operation execution attributes and combines them with the middleware
+            // defaults
+            if (builder.region == null) { builder.region = "" }
```
kggilmer, line 38: question

It's not clear to me that this mapping from null to empty-string is better than allowing `region` to be a nullable property of `AwsSigningConfig` and deal with handling null based on context at points in code where the property is accessed.  From my understanding, here `""` means `null`, is this correct?

kggilmer (changes requested):
Overall LGTM but would like to think a bit more about the modeling of `AwsSigningConfig.region`

aws-runtime/auth/common/src/aws/sdk/kotlin/runtime/auth/signing/AwsSigning.kt
aajtodd: Data classes don't handle arrays as expected. What you want usually is to compare the contents of the array not whether they are the same instance (which is what it will do by default). Intellij even warns you of this and generates the more appropriate one for you.

aws-runtime/auth/common/src/aws/sdk/kotlin/runtime/auth/signing/AwsSigV4SigningMiddleware.kt
aajtodd: Yeah basically. I changed this such that the middleware could just re-use the signing config rather than re-declaring an equivalent config type.

The issue I ran into is we (rightfully so) designed `AwsSigningConfig` to be general purpose and throw errors if not [configured correctly](https://github.com/awslabs/aws-sdk-kotlin/blob/b8a6bc72932f32425b4abef07ef160ef265eb70d/aws-runtime/auth/common/src/aws/sdk/kotlin/runtime/auth/signing/AwsSigningConfig.kt#L32). 

It's didn't seem necessarily the right thing to do to remove the validation since doing so will move the need to validate into each use of it. I opted to just abuse the type rather than remove the validation but if you see a cleaner way I'm all ears.

kggilmer: IMO it would be cleaner to model the difference between presigning and signer middleware via different types which accurately capture their required inputs.  So based on your breakdown the middleware would have a type in which `region` is nullable and the presigner would have a `region` member that was non-null.

aajtodd: That's how it was. I changed it because the signing config has a lot of knobs and it was a lot of duplication. If anything I would just remove the `requireNotNull()` on the signing config and enforce it elsewhere. The only reason I left is to try and give customers an early indication that their signing config isn't valid.

kggilmer: I would prefer that the types used in signing accurately modeled the required inputs.  If you and @ianbotsf prefer this approach however I'm fine with that.

kggilmer (approved):

aws-runtime/auth/common/src/aws/sdk/kotlin/runtime/auth/signing/AwsSigV4SigningMiddleware.kt
ianbotsf: I agree with @kggilmer on this one. Having separate config classes which use different types for `region` seems more correct and more expected to me.

aajtodd: Reverted to using a dedicated type for middleware config

PR #326, opened 2021-09-28
Use mockk to mock Platform without having to supply reference

kggilmer:
## Description of changes

In the AWS configuration PR there was some discussion regarding the testability of the Platform type, which is an `Object`.  Due faciliate testing, callers were required to supply an instance (of the singleton, yes) to the function.  This allowed the function to be tested based on mocked Platform variations.  However it also generated "noise" as the "reference" to the singleton was passed from the top down to all the various internal functions that pulled state from it.

It turns out that the Mockk library supports mocking singletons without having to reference anything.  This updates the tests to use that approach and removes all of the passing of `Platform` in various places.

NOTE: I also snuck in a change I meant to add in the PR, returning the alias `AwsConfiguration` over `Map<String, Map<String, String>>` in `loadAwsConfiguration()`.

## Scope

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/aws-runtime/aws-core/common/src/aws/sdk/kotlin/runtime/AwsSdkSetting.kt
+++ b/aws-runtime/aws-core/common/src/aws/sdk/kotlin/runtime/AwsSdkSetting.kt
@@ -82,9 +82,8 @@ public sealed class AwsSdkSetting<T>(
  * Read the [AwsSdkSetting] by first checking JVM property, environment variable, and default value.
  * Property sources not available on a given platform will be ignored.
  *
- * @param platform A singleton that provides platform-specific settings.  Exposed as a parameter for testing.
  * @return the value of the [AwsSdkSetting] or null if undefined.
  */
 @InternalSdkApi
-public inline fun <reified T> AwsSdkSetting<T>.resolve(platform: Platform): T? =
-    (platform.getProperty(jvmProperty) ?: platform.getenv(environmentVariable) ?: defaultValue) as T?
+public inline fun <reified T> AwsSdkSetting<T>.resolve(): T? =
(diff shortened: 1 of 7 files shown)


aajtodd (approved):
Looks fine, only concern is if `mockk` supports multiplatform (as in all targets, some libraries are multiplatform but lack support for all targets). Just wondering if we will run into future issues.

ianbotsf (commented):
Honestly, I don't care for this change.

Removing the parameterized dependency doesn't change the relationship of `Platform` to any of these functions—it's still a dependency but now a hidden one. We could remove `Platform` altogether from these functions and the tests _wouldn't fail_ to compile (a valuable characteristic of the existing code). Moreover, this style of dependency management makes it easier to _forget_ to mock the `Platform` in future/modified tests, introducing the possibility of false negatives when tests pass on one machine but fail on another.

I favor code where dependencies are explicit and the relationships between them are strongly modeled in APIs. A function signature that takes a dependency makes very clear to calling code (tests, runtime business logic, etc.) the relationship to the dependency and allows for a language-native way of providing an alternate implementation.

ianbotsf:
> Looks fine, only concern is if `mockk` supports multiplatform (as in all targets, some libraries are multiplatform but lack support for all targets). Just wondering if we will run into future issues.

**mockk** appears to be JVM-only at the moment. While it looks like [support for JS is developing](https://github.com/mockk/mockk/tree/master/mockk/js), those subprojects are [commented-out in the build scripts](https://github.com/mockk/mockk/blob/master/settings.gradle#L7).

kggilmer:
Opting not to merge this PR after internal discussion regarding the consequences of this style of mock as it applies to complexity of testing in general.

PR #328, opened 2021-09-30
feat: implement basic retry support in runtime

ianbotsf:
## Issue \#

Addresses #20 and smithy-kotlin#224.

## Description of changes

This change adds support for retries in the runtime (but does not yet integrate them into codegen—separate PR coming). See the included design doc (**docs/design/retries.md**) for more details.

Companion PR: smithy-kotlin#487

## Scope

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)


aws-runtime/aws-core/common/src/aws/sdk/kotlin/runtime/retries/SdkRetryPolicy.kt
```diff
@@ -0,0 +1,41 @@
...
+    internal val knownErrorTypes = mapOf(
+        "BandwidthLimitExceeded" to Throttling,
+        "EC2ThrottledException" to Throttling,
+        "LimitExceededException" to Throttling,
+        "PriorRequestNotComplete" to Throttling,
+        "ProvisionedThroughputExceededException" to Throttling,
+        "RequestLimitExceeded" to Throttling,
+        "RequestThrottled" to Throttling,
+        "RequestThrottledException" to Throttling,
+        "RequestTimeout" to Timeout,
```
aajtodd, line 23: correctness

Aren't both `RequestTimeout` and `RequestTImeoutException` classified as `Transient`?

aws-runtime/aws-core/common/src/aws/sdk/kotlin/runtime/retries/SdkRetryPolicy.kt
```diff
@@ -0,0 +1,41 @@
...
+        "RequestTimeoutException" to Timeout,
+        "SlowDown" to Throttling,
+        "ThrottledException" to Throttling,
+        "Throttling" to Throttling,
+        "ThrottlingException" to Throttling,
+        "TooManyRequestsException" to Throttling,
+        "TransactionInProgressException" to Throttling,
+    )
+
+    override fun evaluateOtherExceptions(ex: Throwable): RetryDirective? = when (ex) {
```
aajtodd, line 33: correctness

Don't we need to classify raw HTTP status codes `500`, `502`, `503`,  and `504` as `Transient`?

aajtodd (changes requested):
Looks good, couple questions

aws-runtime/aws-core/common/src/aws/sdk/kotlin/runtime/retries/SdkRetryPolicy.kt
ianbotsf: The spec calls these kinds of errors "transient" but our code calls them "timeout" errors. Transient isn't a good name for these since that's a class of error that also encompasses throttling and many server-side error conditions.

aws-runtime/aws-core/common/src/aws/sdk/kotlin/runtime/retries/SdkRetryPolicy.kt
ianbotsf: Hmm, seems I missed that part. In order to accomplish that, we'll need access to the status code of the response which is only available in the metadata when casting the [`protocolResponse`](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/Exceptions.kt#L92-L93) to an [`HttpProtocolResponse`](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/response/HttpResponse.kt#L23-L27), which is defined in **smithy-kotlin:runtime:protocol:http**. Given that `SdkRetryPolicy` is currently located in **aws-sdk-kotlin:aws-runtime:aws-core**, it seems we have two options:

* Make **aws-sdk-kotlin:aws-runtime:aws-core** depend on **smithy-kotlin:runtime:protocol:http**. Seems odd to have a "core" package depend on a protocol implementation but maybe that's okay for AWS SDK since all the services run on HTTP?
* Move the retry policy to **aws-sdk-kotlin:aws-runtime:protocols:http**, which already depends on **smithy-kotlin:runtime:protocol:http**. Seems odd to have a retry policy defined somewhere protocol-specific but again maybe that's okay for AWS SDK since all the services run on HTTP?

Which is least odd?

aajtodd: > Make `aws-sdk-kotlin:aws-runtime:aws-core` depend on `smithy-kotlin:runtime:protocol:http`

I don't love this. I just moved config out of `aws-core` for similar reasons. It's an option though.


> Move the retry policy to aws-sdk-kotlin:aws-runtime:protocols:http, which already depends on smithy-kotlin:runtime:protocol:http. Seems odd to have a retry policy defined somewhere protocol-specific but again maybe that's okay for AWS SDK since all the services run on HTTP?

This is closer to what makes sense to me. I could also see defining the core logic in `aws-core` (basically what you have now) and having it be an `open` class that can be further customized for `HTTP` protocols? 

FWIW this looks to be where Rust defined theirs (`aws-http`). 

I don't love (3) either because it defeats the entire point of the generic interface. At that point we should maybe reconsider the design/relationships.

aws-runtime/protocols/http/common/src/aws/sdk/kotlin/runtime/http/retries/AwsDefaultRetryPolicy.kt
```diff
@@ -31,12 +34,23 @@ public object AwsDefaultRetryPolicy : StandardRetryPolicy() {
         "TransactionInProgressException" to Throttling,
     )
 
+    internal val knownStatusCodes = mapOf(
```
aajtodd, line 36: correctness

Aren't these `Transient` errors?

aajtodd (approved):
Fix and ship!

aws-runtime/protocols/http/common/src/aws/sdk/kotlin/runtime/http/retries/AwsDefaultRetryPolicy.kt
aajtodd: Also don't we need to define error codes for `RequestTimeout` and `RequestTimeoutException` as `Transient` error codes?

ianbotsf: Right you are, they should be `Timeout`.

PR #329, opened 2021-09-30
chore: bump snapshot versions

aajtodd:
## Description of changes

Bump versions to next snapshot

### DIFF
--- a/examples/build.gradle.kts
+++ b/examples/build.gradle.kts
@@ -4,7 +4,7 @@ plugins {
 
 allprojects {
     group = "aws.sdk.kotlin.example"
-    version = "0.5.0-alpha"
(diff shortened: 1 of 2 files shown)


ianbotsf (changes requested):
The **gradle.properties** changes look good. Let's also bump the **examples** versions too.

aajtodd:
> The gradle.properties changes look good. Let's also bump the examples versions too.

You would think I would have learned. Really need that automation. But also we need to fix our gradle setup. I firmly believe we need to have the default mode be "build the world" such that all service projects can be directly depended on for examples, tests, benchmarks, etc. As well as to prevent accidental changes in names, packages, etc

ianbotsf (approved):

PR #351, opened 2021-10-19
refactor!: require a resolved configuration

aajtodd:
## Issue \#

fixes smithy-kotlin#484

sibling: https://github.com/awslabs/smithy-kotlin/pull/504

## Description of changes

This PR updates the way AWS SDK clients are constructed such that configuration *MUST* be resolved at instantiation. This means that properties like `region`, `credentialsProvider`, `endpointResolver`, etc are required and non-nullable. This PR adds several ways to get to this resolved state:
* Provide an explicit config (does not require `suspend`)
* Load a shared configuration and re-use across multiple clients (`suspend`)
* Load a client from the environment (`suspend`)

Example of constructing a client:

```kotlin
// fluent builder
val config = S3Client.Config.fluentBuilder()
    .region("us-east-2")
    .build()

val c1 = S3Client(config)

// explicit
val c2 = S3Client {
    region = "us-east-2"
    credentialsProvider = DefaultChainCredentialsProvider()
}

runBlocking {
    // load/resolve automatically
    val c3 = S3Client.loadFromEnvironment()

    // load/resolve with options for controlling the loading process
    val c4 = S3Client.loadFromEnvironment { region = "us-east-1" }

    val sconfig = AwsClientConfig.loadFromEnvironment()
    
    // shared config
    val c5 = S3Client(sconfig)

    // shared config + overrides
    val c6 = S3Client(sconfig) {
        sdkLogMode = SdkLogMode.LogRequest
    }
}

```

### DIFF
--- a/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CrtCredentialUtils.kt
+++ b/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CrtCredentialUtils.kt
@@ -26,3 +26,11 @@ internal fun Credentials.toCrt(): CredentialsCrt = CredentialsCrt(accessKeyId, s
  * Convert CRT credentials into SDK equivalent
  */
 internal fun CredentialsCrt.toSdk(): Credentials = Credentials(accessKeyId, secretAccessKey, sessionToken)
+
+/**
+ * Adapt or convert a SDK credentials provider into CRT equivalent
+ */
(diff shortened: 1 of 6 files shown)


kggilmer:
> The "substantial work" possibility is not the majority use case is it?

IMO it doesn't matter as long as it's possible.  The idea is to eliminate unwanted surprises.  That this call may read from disk or network predicated on factors outside of the program itself I think warrants the extra word.  I am not sure `load` is it but I can't think of anything better atm.

Regarding suspension, we have it everywhere and IMO that in itself is not a clear indicator.

PR #379, opened 2021-10-28
Glacier checksum

### LINKED ISSUE #245: Glacier Request Body Checksum
Compute X-Amz-Sha256-Tree-Hash, and X-Amz-Content-Sha256 checksum for all request bodies

*internal id*: 176558901

ianbotsf:
## Issue \#

Closes #245 

## Description of changes

Adds support for calculating/sending Glacier request body checksums.

Companion PR: smithy-kotlin#512

**Note**: The currently-ignored protocol tests for Glacier will pass once the latest version of smithy is released. Those changes will be added later (or in a separate PR).

## Scope

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/codegen/protocol-tests/build.gradle.kts
+++ b/codegen/protocol-tests/build.gradle.kts
@@ -145,12 +145,18 @@ open class ProtocolTestTask : DefaultTask() {
     @get:Input
     var plugin: String = ""
 
+    /**
+     * The build directory for the task
+     */
+    val generatedBuildDir: File
+        @OutputDirectory
+        get() = project.buildDir.resolve("smithyprojections/${project.name}/$protocol/$plugin")
+
     @TaskAction
     fun runTests() {
         require(protocol.isNotEmpty()) { "protocol name must be specified" }
         require(plugin.isNotEmpty()) { "plugin name must be specified" }
 
-        val generatedBuildDir = project.file("${project.buildDir}/smithyprojections/${project.name}/$protocol/$plugin")
         println("[$protocol] buildDir: $generatedBuildDir")
         if (!generatedBuildDir.exists()) {
             throw GradleException("$generatedBuildDir does not exist")
@@ -169,20 +175,21 @@ open class ProtocolTestTask : DefaultTask() {
     }
 }
 
-
-
 enabledProtocols.forEach {
     val protocolName = it.projectionName
-    tasks.register<ProtocolTestTask>("testProtocol-$protocolName") {
+
+    val protocolTestTask = tasks.register<ProtocolTestTask>("testProtocol-$protocolName") {
         dependsOn(tasks["generateSdk"])
         group = "Verification"
         protocol = protocolName
         plugin = "kotlin-codegen"
-    }
+    }.get()
 
+    // FIXME This is a hack to work around how protocol tests aren't in the actual service model and thus codegen
+    // separately from service customizations.
     tasks.create<Copy>("copyStaticFiles-$protocolName") {
-        from(projectDir.resolve("../../services/$protocolName/common/src"))
-        into(buildDir.resolve("smithyprojections/${project.name}/$protocolName/kotlin-codegen/src/main/kotlin/"))
+        from(rootProject.projectDir.resolve("services/$protocolName/common/src"))
+        into(protocolTestTask.generatedBuildDir.resolve("src/main/kotlin/"))
         tasks["generateSdk"].finalizedBy(this)
     }
 }
--- a/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/AwsRuntimeTypes.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/AwsRuntimeTypes.kt
@@ -66,16 +66,8 @@ object AwsRuntimeTypes {
         val withPayload = runtimeSymbol("withPayload", AwsKotlinDependency.AWS_HTTP)
         val setAseErrorMetadata = runtimeSymbol("setAseErrorMetadata", AwsKotlinDependency.AWS_HTTP)
 
-        object Middleware {
-            val GlacierBodyChecksum = runtimeSymbol("GlacierBodyChecksum", AwsKotlinDependency.AWS_HTTP, "middleware")
-        }
-
         object Retries {
             val AwsDefaultRetryPolicy = runtimeSymbol("AwsDefaultRetryPolicy", AwsKotlinDependency.AWS_HTTP, "retries")
-        }
-
-        object Util {
-            val TreeHasherImpl = runtimeSymbol("TreeHasherImpl", AwsKotlinDependency.AWS_HTTP, "util")
         }
     }
 
--- a/services/build.gradle.kts
+++ b/services/build.gradle.kts
@@ -40,7 +40,6 @@ subprojects {
             kotlin.srcDir("generated-src/test")
 
             dependencies {
-                implementation(kotlin("test"))
                 implementation(kotlin("test-junit5"))
                 implementation(project(":aws-runtime:testing"))
             }
--- a/services/glacier/common/src/aws/sdk/kotlin/services/glacier/internal/TreeHasher.kt
+++ b/services/glacier/common/src/aws/sdk/kotlin/services/glacier/internal/TreeHasher.kt
@@ -5,43 +5,38 @@
 
 package aws.sdk.kotlin.services.glacier.internal
 
-import aws.sdk.kotlin.runtime.InternalSdkApi
 import aws.smithy.kotlin.runtime.http.HttpBody
 import aws.smithy.kotlin.runtime.util.HashFunction
 import kotlinx.coroutines.flow.*
 import kotlin.math.min
 
-@InternalSdkApi
-public typealias HashSupplier = () -> HashFunction
+internal typealias HashSupplier = () -> HashFunction
 
 /**
  * The result of a [TreeHasher] calculation.
  * @param fullHash A full hash of the entire byte array (taken at once)
  * @param treeHash A composite hash of the byte array, taken in chunks.
  */
-@InternalSdkApi
-public class Hashes(public val fullHash: ByteArray, public val treeHash: ByteArray)
+internal class Hashes(public val fullHash: ByteArray, public val treeHash: ByteArray)
 
 /**
  * A hash calculator that returns [Hashes] derived using a tree. See
  * [Computing Checksums](https://docs.aws.amazon.com/amazonglacier/latest/dev/checksum-calculations.html) in the Glacier
  * service guide for more details.
  */
-@InternalSdkApi
-public interface TreeHasher {
+internal interface TreeHasher {
     /**
      * Perform the hash calculation.
      * @param body The [HttpBody] over which to calculate hashes.
      * @return A [Hashes] containing the results of the calculation.
      */
-    public suspend fun calculateHashes(body: HttpBody): Hashes
+    suspend fun calculateHashes(body: HttpBody): Hashes
 }
 
 /**
  * The default implementation of a [TreeHasher].
  */
-@InternalSdkApi
-public class TreeHasherImpl(private val chunkSizeBytes: Int, private val hashSupplier: HashSupplier) : TreeHasher {
+internal class TreeHasherImpl(private val chunkSizeBytes: Int, private val hashSupplier: HashSupplier) : TreeHasher {
     override suspend fun calculateHashes(body: HttpBody): Hashes {
         val full = hashSupplier()
         val hashTree = ArrayDeque<ByteArray>()


aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/middleware/GlacierBodyChecksum.kt
```diff
@@ -0,0 +1,93 @@
...
+        override fun create(block: Config.() -> Unit): GlacierBodyChecksum {
+            val config = Config().apply(block)
+            return GlacierBodyChecksum(config)
+        }
+    }
+
+    override fun <I, O> install(operation: SdkHttpOperation<I, O>) {
+        operation.execution.finalize.intercept { req, next ->
+            val body = req.subject.body
+            require(body !is HttpBody.Streaming || body.isReplayable) {
```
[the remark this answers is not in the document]
ianbotsf: `GlacierException` isn't available from **aws-runtime** so went with `ClientException`.

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/middleware/GlacierBodyChecksum.kt
```diff
@@ -0,0 +1,93 @@
...
+            next.call(req)
+        }
+    }
+}
+
+private val chunkSizeBytes = 1024 * 1024 // 1MB
+
+private class GlacierHashes(val sha256Full: ByteArray, val sha256Tree: ByteArray)
+
+private suspend fun HttpBody.calculateHashes(): GlacierHashes {
```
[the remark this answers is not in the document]
ianbotsf: This inspired me to refactor out the tree hash implementation from the middleware itself, making unit/integration testing easier and the middleware more modular.

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/middleware/GlacierBodyChecksum.kt
```diff
@@ -0,0 +1,93 @@
...
+        override val key: FeatureKey<GlacierBodyChecksum> = FeatureKey("GlacierBodyChecksum")
+
+        override fun create(block: Config.() -> Unit): GlacierBodyChecksum {
+            val config = Config().apply(block)
+            return GlacierBodyChecksum(config)
+        }
+    }
+
+    override fun <I, O> install(operation: SdkHttpOperation<I, O>) {
+        operation.execution.finalize.intercept { req, next ->
```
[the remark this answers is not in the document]
aajtodd: Ahh right, working around that is doable but would be a pain. Finalize seems reasonable

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/middleware/GlacierBodyChecksum.kt
```diff
@@ -0,0 +1,48 @@
...
+            val body = req.subject.body
+            if (body is HttpBody.Streaming && !body.isReplayable) {
+                throw ClientException("This operation requires a byte array or replayable stream")
```
aajtodd, line 37: suggestion

We know the operation name, it's [in the context](https://github.com/awslabs/smithy-kotlin/blob/dc4e5949342157d0a598f076152d5eab2052e1af/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/client/SdkClientOption.kt#L19)

```kotlin
val operation = req.context[SdkClientOption.OperationName]
```

You can add null handling around it you want but IIRC we [require](https://github.com/awslabs/smithy-kotlin/blob/af83d7cd4e30e6f68bfa9bde9351ce56cb4584ae/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/operation/HttpOperationContext.kt#L72) operation name to be set in the context as part of `SdkHttpOperation` construction

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/middleware/GlacierBodyChecksum.kt
```diff
@@ -0,0 +1,48 @@
...
+            }
+            val hashes = treeHasher.calculateHashes(body)
+            req.subject.headers.apply {
```
aajtodd, line 40: style

We already have convenience [extensions](https://github.com/awslabs/smithy-kotlin/blob/d8872592b7ff9adb03b3cbbb2db5b1ae6ebe2bf7/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/request/HttpRequestBuilder.kt#L62) to do this:

```kotlin
req.subject.headers {
    // same
}
```

OR

```kotlin
req.subject.header("X-Amz-Content-Sha256", ...)
req.subject.header("X-Amz-Sha256-Tree-Hash", ...)
```

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/middleware/GlacierBodyChecksum.kt
```diff
@@ -0,0 +1,48 @@
...
+import aws.smithy.kotlin.runtime.util.encodeToHex
+
+private const val defaultChunkSizeBytes = 1024 * 1024 // 1MB
+
+@Suppress("UNUSED_PARAMETER")
+public class GlacierBodyChecksum(config: Config) : Feature {
+    private val treeHasher = config.treeHasher
+
+    public class Config {
+        public var treeHasher: TreeHasher = TreeHasherImpl(defaultChunkSizeBytes) { Sha256() }
```
aajtodd, line 21: concern/style

Why does this need to be configurable?

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/util/TreeHasher.kt
```diff
@@ -0,0 +1,94 @@
+package aws.sdk.kotlin.runtime.http.util
```
aajtodd, line 1: question

Why doesn't his just live with glacier source? Unless I'm missing something I see no reason for this to live in `aws-http`. It could all just be `internal` visibility in glacier (custom) source set.

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/util/TreeHasher.kt
```diff
@@ -0,0 +1,94 @@
...
+@InternalSdkApi
+public class Hashes(public val fullHash: ByteArray, public val treeHash: ByteArray)
+
+/**
+ * A hash calculator that returns [Hashes] derived using a tree. See
+ * [Computing Checksums](https://docs.aws.amazon.com/amazonglacier/latest/dev/checksum-calculations.html) in the Glacier
+ * service guide for more details.
+ */
+@InternalSdkApi
+public interface TreeHasher {
```
aajtodd, line 26: comment

I'm not sure why we need the interface and `Impl` distinction. If this was internal to glacier all of this abstraction could be removed IMO. I don't think we'll end up needing a tree hash algorithm anywhere else

aws-runtime/aws-http/common/test/aws/sdk/kotlin/runtime/http/util/TreeHasherTest.kt
```diff
@@ -0,0 +1,82 @@
...
+        assertEquals("a1c6d421d75f727ce97e6998ab79e6a1cc08ee9502f541f9c5748b462c4dc83f", hashes.treeHash.encodeToHex())
+    }
+}
+
+/**
+ * Calculates a rolling sum for a hash. In this algorithm:
+ * * hash size = chunk size
+ * * every [update] call adds the positional input bytes to a rolling hash plus 1 (differentiates full from tree hashes)
+ */
+class RollingSumHashFunction(private val chunkSize: Int) : HashFunction {
```
aajtodd, line 71: question

Why not just use the sha256 algorithm specified for glacier tree hash? Isn't that what we are trying to verify works correctly?

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/util/TreeHasher.kt
```diff
@@ -0,0 +1,94 @@
...
+     * @return A [Hashes] containing the results of the calculation.
+     */
+    public suspend fun calculateHashes(body: HttpBody): Hashes
+}
+
+/**
+ * The default implementation of a [TreeHasher].
+ */
+@InternalSdkApi
+public class TreeHasherImpl(private val chunkSizeBytes: Int, private val hashSupplier: HashSupplier) : TreeHasher {
```
aajtodd, line 39: style

I don't love `Impl`. If there can only be one of these things couldn't we just remove the interface and call this thing `TreeHasher`? 

Typically if you're going to hide the implementation you would not make it public and instead provide a factory function (e.g.) off the interface. This seems like it doesn't need an interface though in the first place since a tree hash would be a very specific calculation right?

aajtodd (commented):
Couple concerns about tests and where things live but overall looks good

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/middleware/GlacierBodyChecksum.kt
ianbotsf: Good point, I can enhance the error message.

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/middleware/GlacierBodyChecksum.kt
ianbotsf: Missed those, they seem preferable.

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/middleware/GlacierBodyChecksum.kt
ianbotsf: In `Config`? It could probably be a `val` instead.

In the `TreeHasherImpl` constructor? I'd like to leave it configurable since that enables sane unit tests.

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/util/TreeHasher.kt
ianbotsf: It probably could. Let me play with that and make sure nothing breaks (e.g., accessing the Glacier source set from the middleware in **aws-http**).

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/util/TreeHasher.kt
ianbotsf: I broke it out to facilitate DI for unit testing...but then never wound up writing unit tests for the usage site in `GlacierBodyChecksum`. I can replace it with a single class.

aws-runtime/aws-http/common/test/aws/sdk/kotlin/runtime/http/util/TreeHasherTest.kt
ianbotsf: That's pretty painful to debug inside of a consuming hash algorithm. The purpose here was to test the TreeHasher independently of SHA256. I did use SHA256 for the full integration test but it's pretty opaque. If anything gets broken I imagine the unit test will be more helpful in debugging than the integration test.

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/middleware/GlacierBodyChecksum.kt
aajtodd: Ya I meant on the middleware. I guess it depends on whether we keep `TreeHasher` as an interface and a generic concept. I think I've been sort of arguing for just making it a glacier detail rather than something generic in `aws-http`. I don't care _that_ much I'm just having a hard time fathoming where else we would use this specific hashing functionality

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/util/TreeHasher.kt
aajtodd: > It probably could. Let me play with that and make sure nothing breaks (e.g., accessing the Glacier source set from the middleware in aws-http).

I don't follow this. I was suggesting moving this out of `aws-http` and into the "custom" source set for glacier.

aws-runtime/aws-http/common/test/aws/sdk/kotlin/runtime/http/util/TreeHasherTest.kt
aajtodd: > I imagine the unit test will be more helpful in debugging than the integration test

For sure.

> That's pretty painful to debug inside of a consuming hash algorithm.

Is the rolling sum hash easier? 

I'm not really against it per say it just surprised me since on first read through. Mainly because I think I would have a hard time knowing if a rolling hash is correct anymore so than a sha256

ianbotsf: > Is the rolling sum hash easier?

I believe so. The complete implementation is 10 lines of code and is trivially executable mentally. I could use a mocking style here to verify expectations instead if that would be clearer.

aajtodd: No it's fine, I believe you

services/build.gradle.kts
```diff
@@ -38,6 +38,12 @@ subprojects {
         sourceSets.getByName("test") {
             kotlin.srcDir("common/test")
             kotlin.srcDir("generated-src/test")
+
+            dependencies {
+                implementation(kotlin("test"))
```
aajtodd, line 43: I believe Ken added registration of these in certain integrations like S3 and Polly ([example](https://github.com/awslabs/aws-sdk-kotlin/blob/main/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3ErrorMetadataIntegration.kt#L44-L45)). 

I think I prefer this though rather than registering them in codegen, just add it to the overall configuration for all services.

In any case you might want to make sure this doesn't break those service builds.

@kggilmer thoughts

services/glacier/common/src/aws/sdk/kotlin/services/glacier/internal/TreeHasher.kt
```diff
@@ -0,0 +1,99 @@
+/*
```
aajtodd, line 1: fix

commenting on whole file, almost all of these types can be made `internal` visibility (which means we can also remove the `InternalSdkApi` annotation)

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/glacier/GlacierBodyChecksum.kt
```diff
@@ -0,0 +1,57 @@
...
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean =
+        model.expectShape<ServiceShape>(settings.service).sdkId.equals("Glacier", ignoreCase = true)
+
+    override fun customizeMiddleware(
+        ctx: ProtocolGenerator.GenerationContext,
+        resolved: List<ProtocolMiddleware>,
+    ): List<ProtocolMiddleware> = resolved + glacierBodyChecksumMiddleware
+
+    private val glacierBodyChecksumMiddleware = object : HttpFeatureMiddleware() {
+        override val order: Byte = 127 // Must come after AwsSignatureVersion4
```
kggilmer, line 35: comment

not related to PR but it would be cool if we could express these dependency relationships in this relative way.  something like

```kotlin
 override val order: Byte = AwsSignatureVersion4.order.after()
```

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/AwsRuntimeTypes.kt
```diff
@@ -70,9 +70,17 @@ object AwsRuntimeTypes {
         val withPayload = runtimeSymbol("withPayload", AwsKotlinDependency.AWS_HTTP)
         val setAseErrorMetadata = runtimeSymbol("setAseErrorMetadata", AwsKotlinDependency.AWS_HTTP)
 
+        object Middleware {
```
aajtodd, line 69: fix

Looks like all of these runtime symbols can be removed now that they are internal

kggilmer (approved):

codegen/protocol-tests/build.gradle.kts
```diff
@@ -106,7 +106,7 @@ fun generateSmithyBuild(tests: List<ProtocolTest>): String {
             "kotlin-codegen": {
               "service": "${test.serviceShapeId}",
               "package": {
-                "name": "aws.sdk.kotlin.protocoltest.${test.packageName}",
+                "name": "aws.sdk.kotlin.services.${test.packageName}",
```
aajtodd, line 109: nit

Why? It's not a service really.

aajtodd: Oh I see it's probably because of the copy to allow custom source package names to be correct

codegen/protocol-tests/build.gradle.kts
```diff
@@ -172,12 +172,19 @@ open class ProtocolTestTask : DefaultTask() {
...
+    tasks.register<ProtocolTestTask>("testProtocol-$protocolName") {
         dependsOn(tasks["generateSdk"])
         group = "Verification"
-        protocol = it.projectionName
+        protocol = protocolName
         plugin = "kotlin-codegen"
     }
+
+    tasks.create<Copy>("copyStaticFiles-$protocolName") {
+        from(projectDir.resolve("../../services/$protocolName/common/src"))
```
aajtodd, line 184: suggestion

you can use `rootProject.resolve("services/$protocolName/common/src")`

codegen/protocol-tests/build.gradle.kts
```diff
@@ -172,12 +172,19 @@ open class ProtocolTestTask : DefaultTask() {
...
+        into(buildDir.resolve("smithyprojections/${project.name}/$protocolName/kotlin-codegen/src/main/kotlin/"))
```
aajtodd, line 185: suggestion

we should make this a property of the `ProtocolTestTask` which [resolves](https://github.com/awslabs/aws-sdk-kotlin/blob/cfab41de857c15b68f4a297d3fdba54e53153ae1/codegen/protocol-tests/build.gradle.kts#L153) the same root directory (everything up to `kotlin-codegen`)

codegen/protocol-tests/build.gradle.kts
```diff
@@ -172,12 +172,19 @@ open class ProtocolTestTask : DefaultTask() {
...
+    val protocolName = it.projectionName
+    tasks.register<ProtocolTestTask>("testProtocol-$protocolName") {
         dependsOn(tasks["generateSdk"])
         group = "Verification"
-        protocol = it.projectionName
+        protocol = protocolName
         plugin = "kotlin-codegen"
     }
+
+    tasks.create<Copy>("copyStaticFiles-$protocolName") {
```
aajtodd, line 183: suggestion

Maybe add a FIXME comment to this that we shouldn't be doing this and it's a reflection of how the protocol tests for services aren't off the actual service model.

aajtodd (approved):
Fix and ship!

services/build.gradle.kts
kggilmer: Using a static declaration here rather than in codegen seems better to me too.  Note that we only need the `test-junit5` dependency, and the `test` dependency causes spurious KMP errors on my local machine.

Issue #387, opened 2021-10-29
Generate presigners that return signed urls rather than headers


kggilmer:
* Convert to use the query-string as location for signing to make it easier to use presigned urls in browsers and other use cases.

ianbotsf:
Will handle this in [smithy-kotlin#755](https://github.com/awslabs/smithy-kotlin/issues/755)

PR #388, opened 2021-10-29
Refactor presigner querystring

kggilmer:
## Issue \#

https://github.com/awslabs/aws-sdk-kotlin/issues/387

## Description of changes

* Updates presigner codegen to use querystring option 
* Adds presigner e2e tests for supported services
* Remove unused parameter introduced in original presigner PR

### DIFF
--- a/codegen/smithy-aws-kotlin-codegen/src/test/kotlin/aws/sdk/kotlin/codegen/PresignerGeneratorTest.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/test/kotlin/aws/sdk/kotlin/codegen/PresignerGeneratorTest.kt
@@ -176,7 +176,7 @@ class PresignerGeneratorTest {
                     httpRequestBuilder.url.path,
                     httpRequestBuilder.url.parameters.build(),
                     durationSeconds.toLong(),
-                    true,
+                    false,
                     SigningLocation.HEADER
                 )
             }
@@ -214,7 +214,7 @@ class PresignerGeneratorTest {
                     httpRequestBuilder.url.path,
                     httpRequestBuilder.url.parameters.build(),
                     durationSeconds.toLong(),
-                    true,
+                    false,
                     SigningLocation.HEADER
                 )
             }
@@ -233,7 +233,7 @@ class PresignerGeneratorTest {
                 companion object {
                     @JvmStatic
                     fun fluentBuilder(): FluentBuilder = BuilderImpl()
-
+            
                     operator fun invoke(block: DslBuilder.() -> kotlin.Unit): ServicePresignConfig = BuilderImpl().apply(block).build()
                 }
             


codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/PresignerGenerator.kt
```diff
@@ -116,6 +115,20 @@ class PresignerGenerator : KotlinIntegration {
...
+    // Service: S3 signedBodyValue: UNSIGNED-PAYLOAD hasBody: false location: QUERY_STRING (getobject) (GET)
+    // Service: S3 signedBodyValue: UNSIGNED-PAYLOAD hasBody: false location: QUERY_STRING (putobject) (PUT)
+    // Service: STS signedBodyValue: null hasBody: true location: QUERY_STRING (GET)
+    // Service: Polly signedBodyValue: null hasBody: true location: QUERY_STRING (GET)
+    private fun shouldSignBody(serviceId: String, operationId: String): Boolean =
+        when (Pair(serviceId, operationId)) {
+            "com.amazonaws.s3#AmazonS3" to "com.amazonaws.s3#GetObject" -> false
+            "com.amazonaws.s3#AmazonS3" to "com.amazonaws.s3#PutObject" -> false
+            "com.amazonaws.s3#AmazonS3" to "com.amazonaws.s3#UploadPart" -> false
+            "com.amazonaws.sts#AWSSecurityTokenServiceV20110615" to "com.amazonaws.sts#GetCallerIdentity" -> true
```
ianbotsf, line 127: **Comment**: I wonder if this could be derived from whether a body is supplied in the request. Take `PutObject` for instance...most of the time I expect no body will be supplied since [the feature exists to allow callers without a reference to the SDK or credentials to upload a file](https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html). But suppose a caller did provide a body and merely wanted to defer execution until later...under those conditions I'd imagine `shouldSignBody` should return `true`.

kggilmer: Good call, I'll take a look.  Also sharing the link Aaron provided relating to this topic:

services/s3/e2eTest/S3TestUtils.kt
```diff
@@ -94,4 +98,44 @@ object S3TestUtils {
             throw ex
         }
     }
+
+    fun httpResponseFromGetRequest(presignedRequest: HttpRequest): Pair<Int, String?> {
```
aajtodd, line 102: suggestion

If this is general purpose it probably should have a parameter name of `request` rather than `presignedRequest`. You also probably don't need one for both `GET` and `PUT` requests. Just copy the original request (e.g. `req.copy(method=HttpMethod.GET)` or pass in the method as a parameter.

Personally I'd like us to eat our own dog food though rather than using the JVM `HttpsUrlConnection` by using `SdkHttpClient` with our own HTTP engine. We shouldn't really even need these two methods.

services/sts/e2eTest/STSTest.kt
```diff
@@ -0,0 +1,47 @@
...
+import java.net.URL
+import javax.net.ssl.HttpsURLConnection
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+/**
+ * Tests for presigner
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class STSIntegrationTest {
```
aajtodd, line 16: nit

`StsPresignTest` or `StsPresignIntegrationTest`

services/sts/e2eTest/STSTest.kt
```diff
@@ -0,0 +1,47 @@
...
+        val req = GetCallerIdentityRequest { }
+        val pr0 = req.presign(c.config, 60)
+
+        val rc = httpResponseCodeFromGetPresignedUrl(pr0)
+
+        assertEquals(200, rc)
+    }
+}
+
+fun httpResponseCodeFromGetPresignedUrl(presignedRequest: HttpRequest): Int {
```
aajtodd, line 30: nit

This now exists in 3 places which furthers my inclination to just do away with it and use our own machinery.

```kotlin
fun testPutObjectPresigner(): Unit = runSuspendTest {
     val engine = CrtHttpEngine()
     val client = StsClient { region = "us-east-2"; httpEngine = engine }
     
    ...
     
     val presignedRequest = req.presign(...)
     
     // make any assertions on the request and/or copy to a new request if needed
     // e.g. to change the method
     // val presignedGet = presignedRequest.copy(method = HttpMethod.GET)
     assertTrue(presignedRequest.body is HttpBody.Empty)
     
     // use our own internal client which knows how to consume `HttpRequest` already
     val httpClient = sdkHttpClient(engine)
     
     val call = httpClient.call(presignedRequest)
     call.complete()

     assertEquals(200, call.response.status.value)
     engine.close()
}

```

aajtodd (changes requested):
Overall looks good. Couple signing config related issues to address

services/polly/e2eTest/PollyTest.kt
```diff
@@ -0,0 +1,55 @@
...
+import java.net.URL
+import javax.net.ssl.HttpsURLConnection
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+/**
+ * Tests for presigner
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class STSIntegrationTest {
```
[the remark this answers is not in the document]
kggilmer: whoops, thanks for catching the copypasta

ianbotsf (approved):
No comments beyond what @aajtodd said.

aws-runtime/aws-signing/common/src/aws/sdk/kotlin/runtime/auth/signing/Presigner.kt
```diff
@@ -90,7 +90,7 @@ public suspend fun createPresignedRequest(serviceConfig: ServicePresignConfig, r
         credentials = crtCredentials
         signatureType = if (requestConfig.signingLocation == SigningLocation.HEADER) AwsSignatureType.HTTP_REQUEST_VIA_HEADERS else AwsSignatureType.HTTP_REQUEST_VIA_QUERY_PARAMS
         signedBodyHeader = AwsSignedBodyHeaderType.X_AMZ_CONTENT_SHA256
-        signedBodyValue = if (requestConfig.hasBody) AwsSignedBodyValue.UNSIGNED_PAYLOAD else null
+        signedBodyValue = if (requestConfig.signBody) null else AwsSignedBodyValue.UNSIGNED_PAYLOAD
```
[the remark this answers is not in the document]
kggilmer: What you're saying here makes sense to me but it's not clear what is incorrect with line 93.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/PresignerGenerator.kt
```diff
@@ -116,6 +111,14 @@ class PresignerGenerator : KotlinIntegration {
         }
     }
 
+    // Determine if body should be signed.  If body is to be signed, null is passed to signer for signedBodyValue
```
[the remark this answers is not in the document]
kggilmer: updated comment

services/polly/e2eTest/PollyTest.kt
```diff
@@ -0,0 +1,55 @@
...
+        }
+        urlConnection.connect()
+
+        if (urlConnection.errorStream != null) {
+            error("PUT failed: ${urlConnection.errorStream?.bufferedReader()?.readText()}")
+        }
+
+        return urlConnection.responseCode to urlConnection.inputStream?.bufferedReader()?.readText()
+    } finally {
+        urlConnection!!.inputStream.close()
```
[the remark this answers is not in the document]
kggilmer: removed in favor of dogfood

services/s3/e2eTest/S3IntegrationTest.kt
```diff
@@ -98,4 +102,47 @@ class S3BucketOpsIntegrationTest {
         val contents = tempFile.readText()
         assertEquals(contents, roundTrippedContents)
     }
+
+    @Test
+    fun testPutObjectPresigner() = runSuspendTest {
```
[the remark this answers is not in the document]
kggilmer: tests cleaned up

services/s3/e2eTest/S3TestUtils.kt
kggilmer: Opted to use the CRT for GETs, continue to use HttpUrlConnection for PUT.  IMO in this case dogfooding is a mixed bag, because the nature of the feature is that the presigned request should be usable outside of the context of the SDK.  So, we should ensure that some CRT HTTP client specific behavior isn't being relied upon for test pass in some future set of refactorings.

services/sts/e2eTest/STSTest.kt
kggilmer: cleaned up tests

services/sts/e2eTest/STSTest.kt
kggilmer: updated, see above

aws-runtime/aws-signing/common/src/aws/sdk/kotlin/runtime/auth/signing/Presigner.kt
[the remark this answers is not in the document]
aajtodd: If the config/user is requesting `signBody` we are setting `null` AND if there is in fact a body, CRT will sign it. Thus you don't end up with an unsigned body.

As written I'm very confused on what we are trying to do here and with the way it is written it looks like a bug. What are we _trying_ to do here?

kggilmer:
I took another look at the STS presigner based on the discussion in the morning standup.  It turns out that the op is a GET and there is in fact no body.  So there is nothing to append to the querystring.  It remains true in testing though that when a presigned request is generated on the qs, CRT must read the body to calculate the signature.

aws-runtime/aws-signing/common/src/aws/sdk/kotlin/runtime/auth/signing/Presigner.kt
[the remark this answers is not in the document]
kggilmer: I updated the property name to be more descriptive of what CRT does based on the value.

services/sts/e2eTest/STSPresignerTest.kt
```diff
@@ -0,0 +1,33 @@
...
+import kotlin.test.assertEquals
+
+/**
+ * Tests for presigner
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class StsPresignerTest {
+
+    @Test
+    fun testPutObjectPresigner() = runSuspendTest {
```
aajtodd, line 19: fix

more copy pasta, should be `testGetCallerIdentityPresigner` or something

services/sts/e2eTest/STSPresignerTest.kt
```diff
@@ -0,0 +1,33 @@
...
+        val c = StsClient { region = "us-east-2" }
+        val req = GetCallerIdentityRequest { }
+        val presignedRequest = req.presign(c.config, 60)
```
aajtodd, line 22: question

Are there any assertions to make about a presigned request here or for s3/polly? e.g. should a particular protocol/method be asserted, or no http body, etc?

aajtodd (approved):
Fix and ship

services/sts/e2eTest/STSPresignerTest.kt
kggilmer: fixed

services/sts/e2eTest/STSPresignerTest.kt
kggilmer: Essentially the tests here simply verify that the signed request is not rejected by the service.  A deeper level of testing would include verifying the request values, yes.  IMO this is not important here because it seems unlikely that an invalid request (eg unexpected body provided, wrong method) would result in a valid service request.  Please LMK if you see it differently

PR #405, opened 2021-11-04
API service docs workaround

kggilmer:
Build job that generates SDK API docs individually and link with a top index.  The job generates an `api-docs` release with the entire document content as a zip artifact.  No publishing to ghpages or hosting anything on branches.

## Description of changes
* Provides API service docs for all generatable AWS services, in HTML
* Includes `smithy-kotlin` docs in-line for runtime types there
* Provides a static, top-level index of all service modules

### Demo
* Successful Job Run: https://github.com/kggilmer/aws-sdk-kotlin/runs/4146227584
* Doc Artifacts: https://github.com/kggilmer/aws-sdk-kotlin/releases/tag/api-docs-1636417393

### Known Issues
* Some cross-repo links are broken, due to https://github.com/Kotlin/dokka/issues/2211
* No links to source yet

### DIFF
--- a/docs/howto/README.md
+++ b/docs/howto/README.md
@@ -6,3 +6,4 @@ This documentation supplements the official [AWS SDK for Kotlin Developer Guide]
 ## Configuration
 
 * [Endpoints](./configuring/endpoints.md) - How to configure a custom endpoint resolver
+* [Http Clients](./configuring/http-clients.md) - How to specify an alternative HTTP client
--- a/docs/howto/configuring/http-clients.md
+++ b/docs/howto/configuring/http-clients.md
@@ -1,0 +1,33 @@
+# Configuring HTTP Clients
+
+By default the AWS SDK for Kotlin uses an HTTP client known as the AWS Common Runtime (CRT) HTTP client. This 
+client was written by AWS to ensure the best experience with AWS services. However customers may choose to override 
+the default HTTP client by specifying an existing [HttpClientEngine](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngine.kt) 
+implementation or implementing their own, and referencing that implementation in the service client configuration at 
+the time of client construction.
+
+The SDK provides an additional client, `KtorEngine` via the `aws.smithy.kotlin:http-client-engine-ktor` dependency.
+
+[![Maven][maven-badge]][maven-url]
+
+[maven-badge]: https://img.shields.io/maven-central/v/aws.smithy.kotlin/http-client-engine-ktor.svg?label=Maven
+[maven-url]: https://search.maven.org/search?q=g:aws.smithy.kotlin+a:http-client-engine-ktor
+
+## Example
+
+The following code snippet demonstrates constructing an S3 client using the [Ktor OkHttp HTTP client](https://ktor.io/docs/http-client-engines.html#okhttp):
+
+`build.gradle.kts`:
+```kotlin
+dependencies {
+    implementation("aws.smithy.kotlin:http-client-engine-ktor:<version>")
+}
+```
+
+Application code:
+```kotlin
+val sharedConfig = AwsClientConfig.fromEnvironment()
+val client = S3Client(sharedConfig) {
+    httpClientEngine = KtorEngine()
+}
+```


.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,221 @@
...
+                 ./gradlew --no-parallel -PdokkaOutSubDir=$SERVICE_NAME -PsmithyKotlinPackageListUrl=file://$smithy_kotlin_package_list_path dokkaHtmlMultiModule && break
+                 n=$((n+1))
+                 sleep 15
+              done
+
+              echo ::set-output name=docs-generated::"true"
+            else # SDK was excluded
+              echo "Ignoring excluded service $SERVICE_NAME"
+              echo ::set-output name=docs-generated::"false"
+            fi
```
ianbotsf, line 88: **Question**: I don't understand this. Won't this set `docs-generated` to whatever the last result is in the loop? Meaning if an excluded service is last in the loop, `docs-generated` will be false even if other services' docs were generated.

.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,221 @@
...
+          done
+      - name: Compress Docs
+        if: steps.gen-docs.outputs.docs-generated == 'true'
+        run: |
+          tar cJvf api-docs-${{ matrix.service-prefix }}.tar.xz build/dokka/*
```
ianbotsf, line 93: **Question**: The `v` in this `tar` command will significantly increase the logs. Is it strictly necessary?

.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,221 @@
...
+        with:
+          name: smithy-kotlin-docs
+          path: smithy-kotlin-docs
+
+  api-doc-gen-parallel:
+    runs-on: ubuntu-latest
+    needs: smithy-kotlin-docs
+    strategy:
+      matrix:
+        service-prefix: [ a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p,q,r,s,t,u,v,w,x,y,z ]
```
ianbotsf, line 36: **Comment**: To get a more even distribution, you could consider using modular arithmetic instead of a prefix. With this method, some instances will be building 30+ services while others will build 1.

aajtodd: We already have a way to do this in the codebuild service check job FYI:

```shell
      - echo "Bootstrapping AWS services"
      - find codegen/sdk/aws-models -name '*.json' | sort | awk "NR % $NUM_JOBS == $JOB" > services.txt
      - cat services.txt
      - includes=$(cat services.txt | cut -d"/" -f4 | cut -d"." -f1 | sed 's/^\(.*\)$/+\1/' | paste -sd "," -)
      - echo $includes
      - ./gradlew -Paws.services=$includes :codegen:sdk:bootstrap
```

.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,221 @@
+name: API Docs
+
+on: # TODO: determine best event hook (release, etc)
+  workflow_dispatch:
```
aajtodd, line 4: it would be nice to have this run off release as well and just upload artifacts to a release OR have inputs to the workflow dispatch such that it updates the release with the artifact (i.e. use this release and tag to run with)

Creating independent releases is sort of confusing and creates extra tags for only this purpose. I suppose if we only plan on this being a stop gap for a very short time it won't matter but something to consider doing if it isn't too much work.

kggilmer: At this point the job is best manually run due to time it takes the job to complete and some flakeyness observed in the past few weeks.  I've seen weird behavior where jobs are simply cancelled...some process returns 1 and the job halts.  I've also seen a case where a step simply froze.  The runner may have died or something.  In any case, beause of this random flakiness, I do not think we should tie this job in a critical-path to the weekly release.  if this job fails w/ some weirdness, it can be re-run without impacting other release process steps.

After a few weeks of job runs go by ok (and we see that the need for this job will not be super short term) I totally agree with what you've said here and can make followup work to integrate it into the SDK release.

.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,221 @@
...
+
+env:
+  DOC_BRANCH_NAME: api-docs # the branch from which github pages content is sourced
+  PUBLISH_GH_PAGES: false # toggle publishing on github pages branch
```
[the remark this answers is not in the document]
kggilmer: sure

kggilmer:
* Responded to PR feedback
* Remove unneeded variables, configuration
* Assume/clone same branch/tag exists on smithy-kotlin rather than cloning head of default branch

Test from latest revision: https://github.com/kggilmer/aws-sdk-kotlin/runs/4158355794

.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,196 @@
...
+              until [ "$n" -ge 5 ]
+              do
+                 ./gradlew --no-parallel -PdokkaOutSubDir=$SERVICE_NAME -PsmithyKotlinPackageListUrl=file://$smithy_kotlin_package_list_path dokkaHtmlMultiModule && break
+                 n=$((n+1))
+                 sleep 15
+              done
+            else # SDK was excluded
+              echo "Ignoring excluded service $SERVICE_NAME"
+              docs_generated=false
+            fi
```
ianbotsf, line 85: **Correctness**: I think this is a regression from what it was before. Now _any_ exclusion will set the entire step's `docs_generated` to `false`...even if the excluded service wasn't last. I think you want to initially set `docs_generated=false` and conditionally set it to `true` in the happy path of `if [[ $excluded == 0 ]]` (line 69-81). That'll ensure it's `true` if _at least one_ service generated docs.

ianbotsf (changes requested):

.github/workflows/api-docs.yml
kggilmer: Yes.  test for this is still running.  Rather than save time by pushing untested updates I'm going to wait for jobs to complete.  If you care I'm pushing test changes here first: https://github.com/kggilmer/aws-sdk-kotlin/tree/feat-api-docs-codex

.github/workflows/api-docs.yml
aajtodd: > I do not think we should tie this job in a critical-path to the weekly release

I don't believe it will block anything it would just trigger off the release event and then upload artifacts to that release. 

At any rate in the short run we can keep it manual but could still have it tied to an existing release such that tag/release are inputs to the workflow dispatch, see:
https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#example-workflow-configuration

That way we don't have to create separate tags/releases just for docs, they can be tied to the same tags/release as an actual release.

aajtodd (approved):

.github/workflows/api-docs.yml
aajtodd: Just saw this as well FYI: https://cli.github.com/manual/gh_release_upload

.github/workflows/api-docs.yml
kggilmer: @ianbotsf the awslabs branch is now up to date with my private repo test updates.  Please LMK if you still see an issue here.

ianbotsf (approved):

.github/workflows/api-docs.yml
kggilmer: The "critical path" part is how the release process is altered if the api-docs job has a spurious failure.  Re-release?  Run the job manually?  

In any case I realized it would be pretty easy to test it working and so added it now rather than later.  Latest version builds from (pre)release hook.  See here for (presently running) demo: https://github.com/kggilmer/aws-sdk-kotlin/actions/runs/1445251321

kggilmer: > the workflow dispatch takes as inputs the release/tag to work on and upload to

I had something like this in an earlier version of the workflow.  Now it simply uses the branch which you select when launching any workflow via workflow_dispatch.  However I now realize this isn't going to work for us.  In "PR mode" we have a convention of consistent branch names across repos, but this association stops there.  For release builds we need to pull the version of `smithy-kotlin` from the gradle properties in `aws-sdk-kotlin` to ensure to pull down the correct upstream dependency.  This implies that smithy-kotlin is released before aws-sdk-kotlin, which I believe we already are working this way.

> you could then cut a release and manually go run the job 

I've updated the `workflow_dispatch` to require specifying the tag.  I assumed that tags and branches were the same here but turns out that's not the case so an explicit field is required.

Testing manual launch: https://github.com/kggilmer/aws-sdk-kotlin/actions/runs/1445980129
Testing release launch: https://github.com/kggilmer/aws-sdk-kotlin/actions/runs/1445987751

kggilmer: Eventually completed test run: https://github.com/kggilmer/aws-sdk-kotlin/actions/runs/1446809624

Release: https://github.com/kggilmer/aws-sdk-kotlin/releases/tag/v0.9.0-alpha

.github/workflows/api-docs.yml
kggilmer: I'll come back to this great optimization idea if it ends up that the time savings will be of benefit.  As it stands the job runs between 1 and 2 hours.  A flatter distribution will certainly cause the job to complete faster but don't expect it to make a game-changing dent in the total job time.

This job is fairly complex (for my limited gh workflow skill set at least) and even minor changes take lots of wall clock time to tweak and test.

gradle.properties
```diff
@@ -13,7 +13,7 @@ sdkVersion=0.9.2-SNAPSHOT
 smithyVersion=1.13.1
 smithyGradleVersion=0.5.3
 # smithy-kotlin codegen and runtime are versioned together
-smithyKotlinVersion=0.7.2-SNAPSHOT
+smithyKotlinVersion=0.7.1-alpha
```
aajtodd, line 16: do these versions need reverted?

.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,228 @@
...
+        run: |
+          owner=$(echo ${GITHUB_REPOSITORY} | cut -d/ -f1)
+          echo ::set-output name=repo-owner::"$owner"
+          smithy_kotlin_version=$(cat aws-sdk-kotlin/gradle.properties | grep smithyKotlinVersion | cut -d= -f2)
+          echo ::set-output name=smithy_kotlin_version::"v$smithy_kotlin_version"
+      - name: Checkout smithy-kotlin sources
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ steps.repo-owner.outputs.repo-owner }}/smithy-kotlin
+          ref: ${{ env.GITHUB_REF }}
```
aajtodd, line 42: correctness/question

Won't this use the ref from aws-sdk-kotlin? Will this be the correct version of `smithy-kotlin`?

aajtodd (approved):

.github/workflows/api-docs.yml
kggilmer: Yes this is wrong, will fix.

kggilmer: Fix verified

image

gradle.properties
kggilmer: yep, reverting before push

PR #424, opened 2021-11-16
fix(codegen): render default endpoint resolver for machinelearning

### LINKED ISSUE #423: Machine learning customization is generating client without a default resolver
The machine learning customization introduced by #244 is causing the default client to be generated _without_ an endpoint resolver middleware for all but the `Predict` operation.

This is because the customization replaces the default resolver middleware for all operations but the customized protocol middleware only applies to the `Predict` operation.

aajtodd:
## Issue \#

fix #423 

## Description of changes

Ensure a default endpoint resolver is still registered for machine learning. The custom resolver for predict now runs after the default resolver to override the endpoint. We don't currently have a way to replace/override middleware per/operation in codegen. 

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -14,7 +14,7 @@ labels: bug, needs-triage
 <!--- Tell us what happens instead of the expected behavior -->
 
 <!--- Include full errors, uncaught exceptions, stack traces, and relevant logs -->
-<!--- To turn on SDK logging, follow instructions here: http://github.com/aws-sdk-kotlin/tree/main/docs/debugging.md -->
+<!--- To turn on SDK logging, follow instructions here: https://github.com/awslabs/aws-sdk-kotlin/tree/main/docs/debugging.md -->
 <!--- If service responses are relevant, please include wirelogs (after removing any sensitive content!) -->
 
 ## Steps to Reproduce
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -50,13 +50,13 @@ jobs:
           printf "\norg.gradle.jvmargs=-Xmx6g -XX:MaxPermSize=6g" >> gradle.properties
 
           ./gradlew dokkaHtmlMultiModule
-          mkdir smithy-kotlin-docs
-          mv build/dokka/htmlMultiModule/* smithy-kotlin-docs/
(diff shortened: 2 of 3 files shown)


codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/machinelearning/MachineLearningEndpointCustomization.kt
```diff
@@ -18,21 +17,22 @@ import software.amazon.smithy.model.shapes.OperationShape
...
+    // the default resolver middleware will still be present in the operation's execution stack, we just
+    // need to ensure that the custom resolver runs _after_ the default
     override fun customizeMiddleware(
         ctx: ProtocolGenerator.GenerationContext,
         resolved: List<ProtocolMiddleware>
-    ): List<ProtocolMiddleware> =
-        super
-            .customizeMiddleware(ctx, resolved)
-            .replace(endpointResolverMiddleware) { it is ResolveAwsEndpointMiddleware }
+    ): List<ProtocolMiddleware> = resolved + endpointResolverMiddleware
```
kggilmer, line 25: question

> the custom resolver runs _after_ the default

is enforced here by the fact that `endpointResolverMiddleware` is the last term here?

kggilmer (commented):
Can you briefly mention testing steps here as I believe we don't have test coverage for this?

ianbotsf (approved):

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/machinelearning/MachineLearningEndpointCustomization.kt
aajtodd: Yes, the endpoint resolver is there by default so this will always run after. The functionality is covered by a protocol test

aajtodd:
> Can you briefly mention testing steps here as I believe we don't have test coverage for this?

I generated the machine learning client and noticed not endpoint resolver was registered in any of the operations except `Predict`. I regenerated after the change and verified that (1) every operation has an endpoint resolver and (2) `Predict` has the customized resolver registered after (again also covered by a protocol test)

kggilmer (approved):

PR #469, opened 2021-12-10
feat(rt): credential providers redux

### LINKED ISSUE #252: SSO Credential Provider
* /seps/accepted/shared/sso-provider/
* /seps/accepted/shared/sso-login-flow/

*internal id*: 174880365

### LINKED ISSUE #298: Construct DefaultCredentialsChain manually
The [DefaultChainCredentialsProvider](https://github.com/awslabs/aws-sdk-kotlin/blob/main/aws-runtime/auth/common/src/aws/sdk/kotlin/runtime/auth/DefaultChainCredentialsProvider.kt) directly wraps the CRT implementation. 

This makes it difficult to respect JVM system properties that control credentials and roles (which should be resolved before environment variables in the chain). It also makes it hard to support overriding the default profile via JVM system property `aws.profile`. The default chain has no way to support that behavior, it only respects the `AWS_PROFILE` environment variable (see #297).

We should implement the default chain provider ourselves once all of the individual providers we need are available.

### LINKED ISSUE #355: CRT based credential providers leak implementation
## Describe the bug

All of the implementations that implement [CrtCredentialsProvider](https://github.com/awslabs/aws-sdk-kotlin/blob/v0.7.0-alpha/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CrtBasedCredentialsProvider.kt#L10) expose the `crtProvider` property publicly:

Screen Shot 2021-10-20 at 10 25 21 AM

Discovered while looking at the generated documentation for the runtime. 

This is an implementation detail and should be internal or private.

### LINKED ISSUE #357: Address runBlocking in nested STS credential provider translation
There is [PR feedback](https://github.com/awslabs/aws-crt-kotlin/pull/36#discussion_r732766506) to change how `CompletableFuture` is handled in a nested credential provider.  This issue is to track updating the alternative approach until the test case mentioned in the notes is able to pass.

### LINKED ISSUE #419: Difficult to diagnose CredentialProvider failures
## Describe the bug

When I instantiate and attempt to consume a service client, and there is some issue with the credential providers that causes authentication to fail, the failure is opaque and difficult to diagnose.

## Expected Behavior

In the Java SDKs, if there is an auth failure in the CredentialProvider chain, an exception is thrown with a very detailed error message indicating which providers it attempted to use to authenticate and why each of them failed.  This behavior is pretty crucial for debugging purposes.

## Current Behavior

Even with logging at `TRACE`, this is all the output I see when my auth fails:

```
02:09:12.668 PM TRC sdkRequestId: 3f91daca-a33e-47d8-8e12-1b4be01a186d; service: EC2; operation: DescribeLaunchTemplates; - request serialized in 11.929645ms
Exception in thread "main" java.lang.RuntimeException: Failed to get a valid set of credentials
	at software.amazon.awssdk.crt.auth.credentials.CredentialsProvider.onGetCredentialsComplete(CredentialsProvider.java:46)
```

This ^^ link is a 404.

## Steps to Reproduce

```
// Make sure you don't have any default credentials set up, either via
//  ~/.aws/config or ~/.aws/credentials or env vars.
val ec2Client = Ec2Client { region = "us-west-2" }
ec2Client.describeAvailabilityZones(DescribeAvailabilityZonesRequest {})
```

## Possible Solution

Add log messages indicating which providers were attempted and why they failed.

Related question: how do I know which providers are currently supported?  Is SSO supported?
What about profiles with the `credential_process` configured?

## Context

I am trying to do dev work where I run some SDK code on my laptop and have it make API calls
up to my AWS account.  We do not use static credentials nor default profiles, so my auth info
is currently all provided via sso/credential_process config inside of profiles in `~/.aws/config`.

## Your Environment

* AWS Kotlin SDK version used: 0.8.0-alpha
* Operating System and version: OSX Big Sur (x86)

### LINKED ISSUE #461: Add SSO credentials provider to default chain
### Describe the feature

Integrate the SSO credentials provider into the `DefaultCredentialsProvider` chain (by way of `ProfileCredentialsProvider`)

Customers have to manually instantiate and configure an SSO provider. They cannot configure it from a profile and have it work out of the box:

Example,customers would normally configure a profile for SSO and it would be picked up by the default chain:

```ini
[profile my-sso-profile]
sso_region = us-east-2
sso_account_id = 123456789
sso_role_name = MySSORole
sso_start_url = https://my-portal-start-url/start
```

This will not work until the SSO provider is integrated into the default chain, for now they must manually instantiate a provider:

```kotlin

    val ssoProvider = SsoCredentialsProvider(
        accountId = "123456789",
        roleName = "MySSORole",
        startUrl = "https://my-portal-start-url/start",
        ssoRegion = "us-east-2"
    )
    val provider = CachedCredentialsProvider(ssoProvider)
```

### Proposed Solution

_No response_

_No response_

### Acknowledge

- [X] I may be able to implement this feature request

0.9.4-beta

### Platform (JVM/JS/Native)

JVM

N/A

aajtodd:
## Issue \#

closes #355
closes #298
closes #461 
closes #419
closes #357
closes #252 

## Description of changes

* Implement SSO provider and add to default chain
* Refactor the credential providers to no longer require CRT

### DIFF
--- a/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CachedCredentialsProvider.kt
+++ b/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CachedCredentialsProvider.kt
@@ -12,7 +12,6 @@ import aws.smithy.kotlin.runtime.logging.Logger
 import aws.smithy.kotlin.runtime.time.Clock
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
-import kotlin.time.ExperimentalTime
 
 private const val DEFAULT_CREDENTIALS_REFRESH_BUFFER_SECONDS = 10
 /**
@@ -42,7 +41,6 @@ public const val DEFAULT_CREDENTIALS_REFRESH_SECONDS: Int = 60 * 15
  *
  * @return the newly-constructed credentials provider
  */
-@OptIn(ExperimentalTime::class)
 public class CachedCredentialsProvider(
     private val source: CredentialsProvider,
     private val expireCredentialsAfter: Duration = DEFAULT_CREDENTIALS_REFRESH_SECONDS.seconds,
--- a/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/DefaultChainCredentialsProvider.kt
+++ b/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/DefaultChainCredentialsProvider.kt
@@ -12,7 +12,6 @@ import aws.smithy.kotlin.runtime.http.engine.HttpClientEngine
 import aws.smithy.kotlin.runtime.io.Closeable
 import aws.smithy.kotlin.runtime.util.Platform
 import aws.smithy.kotlin.runtime.util.PlatformProvider
-import kotlin.time.ExperimentalTime
 
 /**
  * Default AWS credential provider chain used by most AWS SDKs.
@@ -78,7 +77,6 @@ public class DefaultChainCredentialsProvider constructor(
  * Wrapper around [StsWebIdentityCredentialsProvider] that delays any exceptions until [getCredentials] is invoked.
  * This allows it to be part of the default chain and any failures result in the chain to move onto the next provider.
  */
-@OptIn(ExperimentalTime::class)
 private class StsWebIdentityProvider(
     val platformProvider: PlatformProvider,
     val httpClientEngine: HttpClientEngine? = null
--- a/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/StsAssumeRoleCredentialsProvider.kt
+++ b/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/StsAssumeRoleCredentialsProvider.kt
@@ -17,7 +17,7 @@ import aws.smithy.kotlin.runtime.time.epochMilliseconds
 import aws.smithy.kotlin.runtime.util.Platform
 import aws.smithy.kotlin.runtime.util.PlatformEnvironProvider
 import kotlin.time.Duration
-import kotlin.time.ExperimentalTime
+import kotlin.time.Duration.Companion.seconds
 
 private const val GLOBAL_STS_PARTITION_ENDPOINT = "aws-global"
 private const val PROVIDER_NAME = "AssumeRoleProvider"
@@ -41,14 +41,13 @@ private const val PROVIDER_NAME = "AssumeRoleProvider"
  * @param duration The expiry duration of the STS credentials. Defaults to 15 minutes if not set.
  * @param httpClientEngine The [HttpClientEngine] to use when making requests to the STS service
  */
-@OptIn(ExperimentalTime::class)
 public class StsAssumeRoleCredentialsProvider(
     private val credentialsProvider: CredentialsProvider,
     private val roleArn: String,
     private val region: String? = null,
     private val roleSessionName: String? = null,
     private val externalId: String? = null,
-    private val duration: Duration = Duration.seconds(DEFAULT_CREDENTIALS_REFRESH_SECONDS),
+    private val duration: Duration = DEFAULT_CREDENTIALS_REFRESH_SECONDS.seconds,
     private val httpClientEngine: HttpClientEngine? = null
 ) : CredentialsProvider {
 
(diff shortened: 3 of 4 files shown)


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -0,0 +1,279 @@
...
+    else -> this
+}
+
+/**
+ * Return current result if not null, otherwise use the result from calling [fn]
+ */
+private inline fun LeafProviderResult?.orElse(fn: () -> LeafProviderResult?): LeafProviderResult? = when (this) {
+    null -> fn()
+    else -> this
+}
```
ianbotsf, line 179: **Comment**: I don't quite understand why these functions are necessary over the built-in Elvis operator. Looking at the call site, I see:

```kotlin
return webIdentityTokenCreds()
    .orElse(::ssoCreds)
    .unwrapOrElse(::staticCreds)
    .unwrap()
```

Can't this be simplified to:

```kotlin
return (webIdentityTokenCreds() ?: ssoCreds() ?: staticCreds()).unwrap()
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CachedCredentialsProvider.kt
```diff
@@ -19,41 +30,41 @@ import aws.sdk.kotlin.crt.auth.credentials.CachedCredentialsProvider as CachedCr
...
+    private val source: CredentialsProvider,
+    private val expireCredentialsAfter: Duration = Duration.seconds(DEFAULT_CREDENTIALS_REFRESH_SECONDS),
+    refreshBufferWindow: Duration = Duration.seconds(DEFAULT_CREDENTIALS_REFRESH_BUFFER_SECONDS),
+    private val clock: Clock = Clock.System
+) : CredentialsProvider, Closeable {
 
-        // FIXME - note this won't work until https://github.com/awslabs/aws-crt-java/issues/252 is resolved
-        source = builder.source?.let { CredentialsProviderCrtProxy(it) }
-    }
+    private var cachedCredentials = CachedValue<Credentials>(null, bufferTime = refreshBufferWindow, clock)
```
ianbotsf, line 52: **Nit**: Could be `val`.

build.gradle.kts
```diff
@@ -83,9 +83,15 @@ fun Project.prop(name: String): Any? =
...
     subprojects {
-        tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
-            kotlinOptions.allWarningsAsErrors = true
+        if (name !in optOutWErrorProjects) {
+            // FIXME - this is NOT lazily evaluated
+            tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
+                kotlinOptions.allWarningsAsErrors = true
+            }
         }
     }
```
ianbotsf, line 96: Rather than maintaining a denylist, wouldn't it be simpler to turn off `allWarningsAsErrors` directly in **aws-config**'s build file similar to [what we do for **dokka-aws**](https://github.com/awslabs/aws-sdk-kotlin/blob/c05916b4cfa8ee1bc8bfa3416694ef02b15ead50/dokka-aws/build.gradle.kts#L20)?

services/build.gradle.kts
```diff
@@ -10,10 +10,13 @@ plugins {
...
 
 val sdkVersion: String by project
+val kotlinVersion: String by project
+val coroutinesVersion: String by project
+val kotestVersion: String by project
 
 val optinAnnotations = listOf(
     "aws.smithy.kotlin.runtime.util.InternalApi",
-    "aws.sdk.kotlin.runtime.InternalSdkApi",
+    "aws.sdk.kotlin.runtime.InternalSdkApi"
```
ianbotsf, line 19: **Nit**: Removed trailing comma.

ianbotsf (approved):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
aajtodd: I don't recall other than it was actually you that liked this change originally. It's a matter of taste really as changing it back to elvis operator will require throwing the configuration exceptions in each function vs returning an error. 

I'm going to leave it since it's already been tested and we've already reviewed. If we add more leaf providers at some point we can revisit

PR #475, opened 2021-12-15
refactor(rt)!: implement kmp ecs provider

aajtodd:
## Issue \#

N/A

## Description of changes

* Reimplement KMP compatible ECS provider based on SDK runtime primitives
* Adds a CDK test app that can be used to test out the provider (or really any ECS related task) going forward

### DIFF
--- a/aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProviderTest.kt
+++ b/aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProviderTest.kt
@@ -40,7 +40,7 @@ class EcsCredentialsProviderTest {
         expectedExpiration
     )
 
-    private fun ecs_response(): HttpResponse {
+    private fun ecsResponse(): HttpResponse {
         val payload = """
         {
             "Code" : "Success",
@@ -55,7 +55,7 @@ class EcsCredentialsProviderTest {
         return HttpResponse(HttpStatusCode.OK, Headers.Empty, ByteArrayContent(payload))
     }
 
-    private fun error_response(): HttpResponse {
+    private fun errorResponse(): HttpResponse {
         val payload = """
         {
             "Code" : "TestError",
@@ -66,7 +66,7 @@ class EcsCredentialsProviderTest {
         return HttpResponse(HttpStatusCode.BadRequest, Headers.Empty, ByteArrayContent(payload))
     }
 
-    private fun ecs_request(url: String, authToken: String? = null): HttpRequest {
+    private fun ecsRequest(url: String, authToken: String? = null): HttpRequest {
         val resolvedUrl = Url.parse(url)
         val builder = HttpRequestBuilder().apply {
             method = HttpMethod.GET
@@ -85,8 +85,8 @@ class EcsCredentialsProviderTest {
     fun testRelativeUri(): Unit = runSuspendTest {
         val engine = buildTestConnection {
             expect(
-                ecs_request("http://169.254.170.2/relative?foo=bar"),
-                ecs_response()
+                ecsRequest("http://169.254.170.2/relative?foo=bar"),
+                ecsResponse()
             )
         }
 
@@ -105,8 +105,8 @@ class EcsCredentialsProviderTest {
         val uri = "http://127.0.0.1/full?foo=bar"
         val engine = buildTestConnection {
             expect(
-                ecs_request(uri),
-                ecs_response()
+                ecsRequest(uri),
+                ecsResponse()
             )
         }
 
@@ -140,8 +140,8 @@ class EcsCredentialsProviderTest {
         val uri = "https://amazonaws.com/full"
         val engine = buildTestConnection {
             expect(
-                ecs_request(uri),
-                ecs_response()
+                ecsRequest(uri),
+                ecsResponse()
             )
         }
 
@@ -171,8 +171,8 @@ class EcsCredentialsProviderTest {
         val token = "auth-token"
         val engine = buildTestConnection {
             expect(
-                ecs_request("http://169.254.170.2/relative", token),
-                ecs_response()
+                ecsRequest("http://169.254.170.2/relative", token),
+                ecsResponse()
             )
         }
 
@@ -193,8 +193,8 @@ class EcsCredentialsProviderTest {
     fun testErrorResponse(): Unit = runSuspendTest {
         val engine = buildTestConnection {
             expect(
-                ecs_request("http://169.254.170.2/relative"),
-                error_response()
+                ecsRequest("http://169.254.170.2/relative"),
+                errorResponse()
             )
         }
 


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
```diff
@@ -5,27 +5,198 @@
...
+ *
+ */
+public class EcsCredentialsProvider internal constructor(
+    private val platform: PlatformEnvironProvider,
+    private val httpClientEngine: HttpClientEngine
+) : CredentialsProvider, Closeable {
+
+    public constructor() : this(Platform, CrtHttpEngine())
+
+    private val retryMiddleware = run {
```
kggilmer, line 57: nice

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
```diff
@@ -5,27 +5,198 @@
...
+        // full URI requires verification either https OR that the host resolves to loopback device
+        val url = try {
+            Url.parse(uri)
+        } catch (ex: Exception) {
+            throw ProviderConfigurationException("Invalid fullUri `$uri`", ex)
+        }
+
+        if (url.scheme == Protocol.HTTPS) return url
+
+        // TODO - validate loopback via DNS resolution instead of fixed set. Custom host names (including localhost) that
```
kggilmer, line 137: question 

will this require some KMP work?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
```diff
@@ -5,27 +5,198 @@
...
+            builder.header("Authorization", authToken)
+        }
+        return builder
+    }
+}
+
+internal class EcsCredentialsRetryPolicy : RetryPolicy<Any?> {
+    override fun evaluate(result: Result<Any?>): RetryDirective = when {
+        result.isSuccess -> RetryDirective.TerminateAndSucceed
+        else -> evaluate(result.exceptionOrNull()!!)
```
kggilmer, line 187: nit

it seems a little weird to call an `...orNull()` function and then `!!` it.  Is there a reason this is needed?

tests/integration-tests/ecs-credentials/README.md
```diff
@@ -0,0 +1,70 @@
+# CDK stack for ECS credentials provider testing
```
kggilmer, line 1: wow nice

tests/integration-tests/ecs-credentials/app/build.gradle.kts
```diff
@@ -0,0 +1,34 @@
+plugins {
```
kggilmer, line 1: nit

pull the version strings out property references (log4j, coroutines)

tests/integration-tests/ecs-credentials/bin/ecs-credentials.ts
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env node
+import 'source-map-support/register';
+import * as cdk from 'aws-cdk-lib';
+import { EcsCredentialsStack } from '../lib/ecs-credentials-stack';
+
+const app = new cdk.App();
+new EcsCredentialsStack(app, 'EcsCredentialsStack', {
```
kggilmer, line 7: question

does this reference any cloud things that may change over time that we may need to manage?  I'm not familiar with CDK myself.

tests/integration-tests/ecs-credentials/lib/ecs-credentials-stack.ts
```diff
@@ -0,0 +1,43 @@
...
+    const cluster = new ecs.Cluster(this, 'SdkTestCluster');
+    cluster.enableFargateCapacityProviders();
+
+    const logging = new ecs.AwsLogDriver({
+      streamPrefix: "ecs-test-app"
+    });
+
+    const taskDefinition = new ecs.FargateTaskDefinition(this, 'TaskDef', {
+      memoryLimitMiB: 512,
+      cpu: 256,
```
kggilmer, line 20: question

is this right?

tests/integration-tests/ecs-credentials/test/ecs-credentials.test.ts
```diff
@@ -0,0 +1,17 @@
+// import * as cdk from 'aws-cdk-lib';
```
kggilmer, line 1: nit

is most of this commented out for a reason?

kggilmer (approved):
Impressive work

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
aajtodd: Yeah that's why it isn't implemented. We need the ability to actually resolve a host to something like `List<InetAddress>` and verify that each one is in fact the loopback device (e.g. `InetAddres.isLoopBackDevice(): Boolean`). 

ECS itself only uses the relative URI, this only comes into play when the full URI is set (and not using `https`). I believe greengrass is the primary user of this for _something_ but I don't recall what. Seems safe to wait until it's needed. I created #476 to track it.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
aajtodd: Ya I agree it's a little weird but there is no function on `Result` that provides the exception or throws. The success case has functions like `getOrThrow()` but nothing like that for the error case. No smart casting to be had either :(

Seems like something is missing from `Result` to me but this should be safe since we know it's the error case.

tests/integration-tests/ecs-credentials/bin/ecs-credentials.ts
aajtodd: CDK is just a wrapper around cloudformation templates. To answer your question though no there is nothing to manage here. The expected usage of this is for one off testing. Developers would deploy, run whatever tests they need, and tear down the stack when they are done.

tests/integration-tests/ecs-credentials/lib/ecs-credentials-stack.ts
aajtodd: Yeah it's a bit of an odd one but it's correct. `256` maps to `.25 vCPU`

See docs [here](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_size)

tests/integration-tests/ecs-credentials/test/ecs-credentials.test.ts
aajtodd: This can be deleted, it's part of `cdk init` scaffolding 🤷

PR #478, opened 2021-12-21
feat(rt)!: implement kmp profile credentials provider

aajtodd:
## Issue \#

N/A

## Description of changes

* Implements KMP compatible profile credentials provider in pure Kotlin. 
    * The underlying representation (`ProfileChain`) and tests are mostly ported from Rust (thanks Russell)
* Adds provider name field to `Credentials` for easier debugging

## Testing
Tested with assuming a role:

```kotlin
    StsClient{
        region = "us-east-2"
        credentialsProvider = ProfileCredentialsProvider()
    }.use { client ->
        val resp = client.getCallerIdentity {}

        println("Account: ${resp.account}")
        println("UserID: ${resp.userId}")
        println("ARN: ${resp.arn}")
    }

```

Profile:
```ini
[profile foo]
role_arn = arn:aws:iam::123456789:role/RoleA
region = us-east-2
source_profile = baz

[profile baz]
# static credentials

```

NOTE: Additional test cases that cover more involved scenarios will be added when we replace the default chain.

## Questions
1. What should the behavior be in the code snippet above? `region` is provided to the `StsClient` but not the profile provider. There is no good way to currently plumb this information through. The profile provider requires region to be set still in either the profile or env/system property. I _think_ this is correct but wanted to bring it up for discussion.

### DIFF
--- a/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
+++ b/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
@@ -93,7 +93,7 @@ public class ProfileCredentialsProvider(
     override suspend fun getCredentials(): Credentials {
         val logger = Logger.getLogger<ProfileCredentialsProvider>()
         val source = resolveConfigSource(platform, profileName)
-        logger.debug { "loading credentials from profile `${source.profile}`" }
+        logger.debug { "Loading credentials from profile `${source.profile}`" }
         val profiles = loadAwsProfiles(platform, source)
         val chain = ProfileChain.resolve(profiles, source.profile)
 
@@ -102,16 +102,16 @@ public class ProfileCredentialsProvider(
         val region = region ?: profileOverride?.get("region") ?: resolveRegion(platform)
 
         val leaf = chain.leaf.toCredentialsProvider(region)
-        logger.debug { "resolving credentials from ${chain.leaf.description()}" }
+        logger.debug { "Resolving credentials from ${chain.leaf.description()}" }
         var creds = leaf.getCredentials()
 
         chain.roles.forEach { roleArn ->
-            logger.debug { "assuming role `${roleArn.roleArn}`" }
+            logger.debug { "Assuming role `${roleArn.roleArn}`" }
             val assumeProvider = roleArn.toCredentialsProvider(creds, region)
             creds = assumeProvider.getCredentials()
         }
 
-        logger.debug { "obtained credentials from profile; expiration=${creds.expiration?.format(TimestampFormat.ISO_8601)}" }
+        logger.debug { "Obtained credentials from profile; expiration=${creds.expiration?.format(TimestampFormat.ISO_8601)}" }
         return creds
     }
 


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
```diff
@@ -46,13 +48,20 @@ private const val AWS_CONTAINER_SERVICE_ENDPOINT = "http://169.254.170.2"
...
 public class EcsCredentialsProvider internal constructor(
     private val platform: PlatformEnvironProvider,
-    private val httpClientEngine: HttpClientEngine
+    httpClientEngine: HttpClientEngine? = null
 ) : CredentialsProvider, Closeable {
 
-    public constructor() : this(Platform, CrtHttpEngine())
+    public constructor() : this(Platform)
+
+    private val manageEngine = httpClientEngine == null
```
kggilmer, line 63: style

IMO `owned` is more descriptive than `manage` as to the intent of this property.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
```diff
@@ -5,31 +5,159 @@
...
+            },
+            platformProvider = platform
+        ),
+        "EcsContainer" to EcsCredentialsProvider(platform, httpClientEngine)
+    )
+
+    override suspend fun getCredentials(): Credentials {
+        val logger = Logger.getLogger<ProfileCredentialsProvider>()
+        val source = resolveConfigSource(platform, profileName)
+        logger.debug { "loading credentials from profile `${source.profile}`" }
```
kggilmer, line 96: nit

title case for log entry, for all log emissions in this file

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/LeafProvider.kt
```diff
@@ -0,0 +1,69 @@
...
+
+package aws.sdk.kotlin.runtime.auth.credentials.profile
+
+import aws.sdk.kotlin.runtime.auth.credentials.Credentials
+
+/**
+ * A standalone member of the profile chain. Leaf providers do not require
+ * input credentials to provide their own credentials (e.g. IMDS, ECS, Environment, etc)
+ */
+internal sealed class LeafProvider {
```
kggilmer, line 14: question

it's not connecting for me, what meaning does 'leaf' mean for this type?  Is it that it's terminal, as in a leaf in a tree?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -0,0 +1,245 @@
...
+                        "could not find source profile $sourceProfileName"
+                    } else {
+                        "could not find source profile $sourceProfileName referenced from ${visited.last()}"
+                    }
+                    ProviderConfigurationException(msg)
+                }
+
+                if (visited.contains(sourceProfileName)) {
+                    // we're in a loop, break out
+                    throw ProviderConfigurationException("profile formed an infinite loop: ${visited.joinToString(separator = " -> ")} -> $sourceProfileName")
```
kggilmer, line 53: question

assuming a customer would see this error message, is there some additional text we could add to explain what the customer should do to resolve it?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * Load [LeafProvider.AccessKey] from the current profile or throw an exception if the profile does not contain
+ * credentials
+ */
+private fun AwsProfile.staticCreds(): LeafProvider {
+    val accessKeyId = get(AWS_ACCESS_KEY_ID)
+    val secretKey = get(AWS_SECRET_ACCESS_KEY)
+    return when {
+        accessKeyId == null && secretKey == null -> throw ProviderConfigurationException("profile ($name) did not contain credential information")
```
kggilmer, line 188: nit

"profile ($name) did not contained neither an access key nor secret key" may be more actionable to a user

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -0,0 +1,245 @@
...
+        sourceProfile != null && credSource != null -> throw ProviderConfigurationException("profile ($name) contained both `source_profile` and `credential_source`. Only one or the other can be defined.")
+        sourceProfile == null && credSource == null -> throw ProviderConfigurationException("profile ($name) must contain `source_profile` or `credential_source` but neither were defined")
+        sourceProfile != null && credSource == null -> if (sourceProfile == name) {
+            NextProfile.SelfReference
+        } else {
+            NextProfile.Named(sourceProfile)
+        }
+        // loop back into this profile and pick up the credential source
+        sourceProfile == null && credSource != null -> NextProfile.SelfReference
+        else -> error("shouldn't be able to get here, this is a bug please file a ticket")
```
kggilmer, line 230: suggestion

link to gh issues so the customer knows where to file the ticket

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/ProfileChainTest.kt
```diff
@@ -0,0 +1,420 @@
...
+import aws.sdk.kotlin.runtime.auth.credentials.profile.LeafProvider
+import aws.sdk.kotlin.runtime.auth.credentials.profile.ProfileChain
+import aws.sdk.kotlin.runtime.auth.credentials.profile.RoleArn
+import aws.sdk.kotlin.runtime.config.profile.FileType
+import aws.sdk.kotlin.runtime.config.profile.parse
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.fail
+
+class ProfileChainTest {
```
kggilmer, line 17: wow nice :star_struck:

kggilmer (approved):
Quite clean and easy to follow.  i think every log emission needs the first word capitalized otherwise no concerns found.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/LeafProvider.kt
aajtodd: yeah the provider that doesn't depend on any other providers. In the Rust implementation they called it `BaseProvider`. I didn't find that descriptive enough for my tastes.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
aajtodd: Yeah true, I made it `manageEngine` to match the corresponding parameter that [sdkHttpClient](https://github.com/awslabs/smithy-kotlin/blob/e7ee95701221911780bbd7fb99a2792d3b9e9d32/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/SdkHttpClient.kt#L27) uses.

I kind of expect this is going to change when we decide on our ownership direction

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
aajtodd: did we document this convention somewhere. I remember discussing it and promptly forgot and I will probably continue to forget until we get it written down in a style guide

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -0,0 +1,245 @@
...
+private fun AwsProfile.webIdentityTokenCredsOrNull(): LeafProvider? {
+    val roleArn = get(ROLE_ARN)
+    val tokenFile = get(WEB_IDENTITY_TOKEN_FILE)
+    val sessionName = get(ROLE_SESSION_NAME)
+    return when {
+        tokenFile == null -> null
+        roleArn == null -> throw ProviderConfigurationException("profile ($name) missing `$ROLE_ARN`")
+        else -> {
+            LeafProvider.WebIdentityTokenRole(roleArn, tokenFile, sessionName)
+        }
```
ianbotsf, line 160: **Nit**: Unnecessary braces

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -0,0 +1,245 @@
...
+            chain.reverse()
+            return ProfileChain(checkNotNull(leaf), chain)
+        }
+    }
+}
+
+private inline fun ProfileMap.getOrThrow(name: String, lazyException: () -> Throwable): AwsProfile {
+    val props = get(name) ?: throw lazyException()
+    return AwsProfile(name, props)
+}
```
ianbotsf, line 94: **Comment**: This seems slightly strange to me. The function accepts a `() -> Throwable` whereas other similar functions (e.g., `require`, `check`, etc.) accept a `() -> Any` and handle the exception initialization themselves. Making this change could also simplify the call site.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -0,0 +1,245 @@
...
+ * Load [LeafProvider.AccessKey] from the current profile or throw an exception if the profile does not contain
+ * credentials
+ */
+private fun AwsProfile.staticCreds(): LeafProvider {
+    val accessKeyId = get(AWS_ACCESS_KEY_ID)
+    val secretKey = get(AWS_SECRET_ACCESS_KEY)
+    return when {
+        accessKeyId == null && secretKey == null -> throw ProviderConfigurationException("profile ($name) did not contain credential information")
+        accessKeyId == null -> throw throw ProviderConfigurationException("profile ($name) missing `aws_access_key_id`")
+        secretKey == null -> throw throw ProviderConfigurationException("profile ($name) missing `aws_secret_access_key`")
```
ianbotsf, line 190: **Question**: `throw throw`? Does this compile?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * Attempt to load [LeafProvider.AccessKey] from the current profile or `null` if the current profile does not contain
+ * credentials
+ */
+private fun AwsProfile.staticCredsOrNull(): LeafProvider? = try {
+    staticCreds()
+} catch (_: ProviderConfigurationException) {
+    null
+}
```
ianbotsf, line 206: **Comment**: At least in JVM, exceptions are slow and thus generally avoided for flow control. This may be more performant by using some other type to encapsulate the result of getting static credentials (e.g., an ADT) and then throwing or returning `null` as appropriate at a higher level:

```kt
sealed class LeafProviderResult {
    data class Success(provider: LeafProvider) : LeafProviderResult()
    data class Failure(errorMessage: String) : LeafProviderResult()
}

private fun AwsProfile.staticCredsResult(): LeafProviderResult {
    val accessKeyId = get(AWS_ACCESS_KEY_ID)
    val secretKey = get(AWS_SECRET_ACCESS_KEY)
    return when {
        accessKeyId == null && secretKey == null -> LeafProviderResult.Failure("profile ($name) did not contain credential information")
        accessKeyId == null -> LeafProviderResult.Failure("profile ($name) missing `aws_access_key_id`")
        secretKey == null -> LeafProviderResult.Failure("profile ($name) missing `aws_secret_access_key`")
        else -> {
            val sessionToken = get(AWS_SESSION_TOKEN)
            LeafProviderResult.Success(LeafProvider.AccessKey(Credentials(accessKeyId, secretKey, sessionToken)))
        }
    }
}

private fun AwsProfile.staticCredsOrNull(): LeafProvider? = when (val result = staticCredsResult()) {
    is LeafProviderResult.Success -> result.provider
    else -> null
}

private fun AwsProfile.staticCredsOrThrow(): LeafProvider = when (val result = staticCredsResult()) {
    is LeafProviderResult.Success -> result.provider
    else -> throw ProviderConfigurationException(result.errorMessage)
}
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -0,0 +1,245 @@
...
+        sourceProfile == null && credSource == null -> throw ProviderConfigurationException("profile ($name) must contain `source_profile` or `credential_source` but neither were defined")
+        sourceProfile != null && credSource == null -> if (sourceProfile == name) {
+            NextProfile.SelfReference
+        } else {
+            NextProfile.Named(sourceProfile)
+        }
+        // loop back into this profile and pick up the credential source
+        sourceProfile == null && credSource != null -> NextProfile.SelfReference
+        else -> error("shouldn't be able to get here, this is a bug please file a ticket")
+    }
```
ianbotsf, line 231: **Comment**: I'm surprised the compiler doesn't detect `else` as dead code. Should be able to avoid the `else -> error` case by rewriting this as nested `if`/`else` statements:

```kt
return
    if (sourceProfile == null) {
        if (credSource == null) {
            throw ProviderConfigurationException(...)
        } else { // credSource != null
            NextProfile.SelfReference
        }
    } else { // sourceProfile != null
        if (credSource == null) {
            if (sourceProfile == name) NextProfile.SelfReference else NextProfile.Named(sourceProfile)
        } else { // credSource != null
            throw ProviderConfigurationException(...)
        }
    }
```

ianbotsf (approved):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
```diff
@@ -38,6 +38,8 @@ import aws.smithy.kotlin.runtime.util.PlatformEnvironProvider
  */
 private const val AWS_CONTAINER_SERVICE_ENDPOINT = "http://169.254.170.2"
 
+private const val PROVIDER_NAME = "EcsContainer"
+
```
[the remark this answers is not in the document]
aajtodd: Provider names are actually called out in an SEP

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
aajtodd: Yeah true but the exception type would be wrong. I want a `ProviderConfigurationException`. What I really want is `check(condition, MyErrorType::class) { "lazy message" }` I guess but not sure that works

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
aajtodd: oof yes but I will fix it, must be copy pasta, good eye

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
aajtodd: Ya I can look into this.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
aajtodd: true but I find the `when` cleaner personally, I was surprised as well

aws-runtime/aws-types/common/src/aws/sdk/kotlin/runtime/auth/credentials/Credentials.kt
```diff
@@ -17,4 +17,5 @@ public data class Credentials(
     val secretAccessKey: String,
     val sessionToken: String? = null,
     val expiration: Instant? = null,
+    val providerName: String? = null,
```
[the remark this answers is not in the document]
aajtodd: We don't really but it seems very useful for debugging purposes. I could be convinced to remove it though

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -0,0 +1,245 @@
...
+
+                if (visited.contains(sourceProfileName)) {
+                    // we're in a loop, break out
+                    throw ProviderConfigurationException("profile formed an infinite loop: ${visited.joinToString(separator = " -> ")} -> $sourceProfileName")
+                }
+
+                visited.add(sourceProfileName)
+
+                // when chaining assume role profiles, SDKs MUST terminate the chain as soon as they hit a profile with static credentials
+                if (visited.size > 1) {
```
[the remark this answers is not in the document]
aajtodd: I'm not sure I follow. As the comment suggests the chain terminates as soon as it hits a profile with static creds. In the case of a single profile with static creds it would break out at L71

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -0,0 +1,245 @@
...
+
+    /**
+     * The list of roles to assume (in-order). The first role will be assumed with credentials from [leaf].
+     * Every role after that should be assumed with the prior role credentials.
+     */
+    val roles: List<RoleArn>
+) {
+    companion object {
+        internal fun resolve(profiles: ProfileMap, profileName: String): ProfileChain {
+            val visited = mutableListOf<String>()
```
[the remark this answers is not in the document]
aajtodd: It can't be a set because the order matters for the error message we form on L53. If we don't use that formatting we I think we could make it a set. Although I didn't go look and see if there is an ordered set available...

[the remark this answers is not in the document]
ianbotsf: [`mutableSetOf`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/mutable-set-of.html) returns a set with order-insertion guaranteed.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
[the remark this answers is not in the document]
ianbotsf: The `if` condition is `visited.size > 1`. Why _shouldn't_ we run the code inside the `if`-block if we've only visited one element so far?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
ianbotsf: Right sorry, I wasn't suggesting using `require` or `check`...I was suggesting keeping this function but constructing the `ProviderConfigurationException` inside of it rather than in calling code.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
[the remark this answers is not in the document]
aajtodd: Perfect. Will update.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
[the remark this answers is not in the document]
aajtodd: Ahh so as the comment on L64 indicates the only indication that multiple profiles are chained is the existince of `role_arn`. This check exists only to terminate a chain at the first profile with credentials, e.g.

```ini
[profile A]
role_arn = arn::test::foo::blah
source_profile = B

[profile B]
access_key_id = AKID
secret_access_key= secret
source_profile = C
# profile A chain must terminate here and not look at C for creds

[profile C]
access_key_id = incorrect_key
secret_access_key= incorrect_secret
```

The chain must terminate at the first profile that has static creds when multiple are involved. The singular profile with static creds that has no chain:

```ini
[profile A]
access_key_id = AKID
secret_access_key= secret
```

is handled by the else branch starting on L68 (just below).

As I type all this I'm now realizing your actual question. I'll leave all the above as context but the answer is in this [test case](https://github.com/awslabs/aws-sdk-kotlin/blob/feat-profile-provider/aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/ProfileChainTest.kt#L103-L115)

ianbotsf (approved):

Issue #485, opened 2022-01-02
 Get S3 object via presign with Unicode name fails with signature mismatch


vcanuel:
### Describe the bug

I generate a presign Request containing a key with an Unicode character. Then I got an error when retrieve it via an HTTP client : 
`The request signature we calculated does not match the signature you provided. Check your key and signing method`

### Expected behavior

I should be able to retrieve the S3 file.

### Current behavior

I got an error : `The request signature we calculated does not match the signature you provided. Check your key and signing method`

### Steps to Reproduce

```
fun main(): Unit = runBlocking {
    val c = S3Client { region = "us-east-2" }

    try {
        println("Creating obj")
        c.putObject(PutObjectRequest.invoke {
            bucket = "test-bucket"
            key = "nameé.txt"
            body = ByteStream.fromString("abody")
        })

        println("Getting obj")
      val getRequest = GetObjectRequest {
            bucket = "test-bucket"
            key = "nameé.txt"
        }
        val signRequest =   getRequest.presign(c.config, 60L)

   CrtHttpEngine().use { engine ->
        val httpClient = sdkHttpClient(engine)
        val call = httpClient.call(signRequest)
        call.complete()
        println("status =  ${call.response.status}") // this returns  status =  403: Forbidden
    }

    } finally {
        c.close()
    }
}
```

### Possible Solution

I think this is the same root cause as this issue https://github.com/awslabs/aws-sdk-kotlin/issues/200

### Context

We have lot of files with Unicode characters it it.

### AWS Kotlin SDK version used

0.9.5-beta

### Platform (JVM/JS/Native)

JVM

### Operating System and version

Mac OS 12.0.1

kggilmer:
Thank you for the clear and detailed bug report @vcanuel .  I've confirmed the issue in a test based on your work.  I'll update this ticket when a fix is available.

kggilmer:
Hi @vcanuel , the fix for this issue was merged into the SDK today.  You will be able to get the fix in the next release of the SDK.

PR #487, opened 2022-01-04
Fix usage of unicode in bucket names of s3 presigner

kggilmer:
## Issue \#

https://github.com/awslabs/aws-sdk-kotlin/issues/485

## Description of changes

* Update presigner service config to contain 2 additional signing parameters needed for unicode bucket names in s3.  
* Update codegen to specify values based on service.  
* Update S3 service tests for unicode bucket names.
* Misc cleanup

### DIFF
--- a/docs/design/tenets.md
+++ b/docs/design/tenets.md
@@ -1,0 +1,7 @@
+# AWS SDK for Kotlin Design Tenets
+
+These are the tenets that drive the development of the AWS SDK for Kotlin in order of priority.
+
+1. **Make simple things simple and complex things possible** - The AWS SDK for Kotlin should provide access to AWS services with obvious, flexible, and convenient APIs.  Where possible the SDK should reduce unneeded complexity in support of common use cases, but only in ways that do not prevent customers from realizing more complex use cases.
+2. **Kotlin idiomatic** - A Kotlin SDK should be geared towards developers working in Kotlin. While Kotlin has bindings and interoperability with other languages, the primary use case will be Kotlin applications.
+3. **Multi-platform** - One of the key benefits of Kotlin is that it enables the ability to “write-once, run in many places”. The AWS SDK for Kotlin should support customers use cases across the set of stable Kotlin target platforms.


ianbotsf (approved):

services/s3/e2eTest/S3PresignerTest.kt
```diff
@@ -43,49 +43,817 @@ class S3PresignerTest {
...
         val contents = "presign-test"
-        val keyName = "put-obj-from-memory-presigned.txt"
 
-        client.putObject {
-            bucket = testBucket
-            key = keyName
-            body = ByteStream.fromString(contents)
-        }
+        testKeyNames.reversed().forEach { keyName ->
+            println("testing $keyName")
```
aajtodd, line 70: can probably remove

services/s3/e2eTest/S3PresignerTest.kt
```diff
@@ -43,49 +43,817 @@ class S3PresignerTest {
...
-        }.presign(client.config, 60)
+            val presignedRequest = GetObjectRequest {
+                bucket = testBucket
+                key = keyName
+            }.presign(client.config, 60)
 
-        CrtHttpEngine().use { engine ->
-            val httpClient = sdkHttpClient(engine)
+            CrtHttpEngine().use { engine ->
+                val httpClient = sdkHttpClient(engine)
```
aajtodd, line 83: style/suggestion

`sdkHttpClient` will close the engine by default so this could be written as:

```kotlin
sdkHttpClient(CrtHttpEngine()).use { httpClient ->
    ...
}
```

aajtodd (approved):

services/s3/e2eTest/S3PresignerTest.kt
kggilmer: I get the following error when changing to your suggested code: `Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
public inline fun <T : Closeable?, R> TypeVariable(T).use(block: (TypeVariable(T)) -> TypeVariable(R)): TypeVariable(R) defined in kotlin.io`

I'm not seeing anywhere where `SdkHttpClient` implements `Closable`..

aajtodd: Hmm indeed it does not implement `Closeable` yet has a `close()` method. Probably a miss (or likely because SdkHttpClient predates our addition of `Closeable`). ignore for now then

PR #491, opened 2022-01-05
feat(rt)!: kmp default credentials provider chain

aajtodd:
## Issue \#

N/A

## Description of changes

* Re-implement the default chain in Kotlin
* Add unit tests adapted from the Rust test suite covering e2e behavior of the default chain
    * The tests are all a directory in a particular convention:
         * `fs/*` is the resulting test filesystem rooted at `fs`
         * `test-case.json` is the expected result of the test (specific to default chain tests)
         * `env.json` is the test environment variables to set
         * `http-traffic.json` is the request/response pair(s) that the engine expects/responds with (there are limitations currently in what can be asserted about the requests but the tests are setup that asserting the request isn't all that important. You'll either get the correct credentials or you wont)
* Fixed SSO response parsing discovered as part of creating the test covering sso in the default chain (response is epoch milliseconds not epoch seconds)
* Refactored constructors to all use consistent names for the same argument types
* Relocated the region/credential chain implementations to be in `aws-config` to get the right exception type(s)

## Notes
* There are still some todos/fixmes. I will follow up with additional smaller PRs to round out the credential provider overhaul. In particular we need to hash out resource management and probably thread through profile override and HttpClientEngine from shared config.

### DIFF
--- a/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CredentialsProviderChain.kt
+++ b/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CredentialsProviderChain.kt
@@ -45,19 +45,18 @@ public open class CredentialsProviderChain(
     }
 
     override fun close() {
-        var closeEx: Throwable? = null
-        providers.forEach {
+        val exceptions = providers.mapNotNull {
             try {
                 (it as? Closeable)?.close()
+                null
             } catch (ex: Exception) {
-                if (closeEx == null) {
-                    closeEx = ex
-                } else {
-                    closeEx!!.addSuppressed(ex)
-                }
+                ex
             }
-
-            if (closeEx != null) throw closeEx!!
+        }
+        if (exceptions.isNotEmpty()) {
+            val ex = exceptions.first()
+            exceptions.drop(1).forEach(ex::addSuppressed)
+            throw ex
         }
     }
 }


kggilmer (approved):
Those test cases are gold.  Nice work.

aws-runtime/aws-config/build.gradle.kts
```diff
@@ -32,10 +30,6 @@ kotlin {
                 implementation("aws.smithy.kotlin:serde-json:$smithyKotlinVersion")
 
 
-                // credential providers
-                implementation("aws.sdk.kotlin.crt:aws-crt-kotlin:$crtKotlinVersion")
-                implementation(project(":aws-runtime:crt-util"))
```
ianbotsf, line 37: **Question**: Can any code be removed from **crt-util** now?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CrtBasedCredentialsProvider.kt
```diff
@@ -1,14 +0,0 @@
...
-
-package aws.sdk.kotlin.runtime.auth.credentials
-
-import aws.sdk.kotlin.crt.auth.credentials.CredentialsProvider as CredentialsProviderCrt
-
-internal interface CrtCredentialsProvider : CredentialsProvider {
-    val crtProvider: CredentialsProviderCrt
-
-    override suspend fun getCredentials(): Credentials = crtProvider.getCredentials().toSdk()
-}
```
ianbotsf, line 14: **Comment**: Seems like a lot of related code can be removed from **aws-crt-kotlin** as well.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/DefaultChainCredentialsProvider.kt
```diff
@@ -5,28 +5,83 @@
...
+    private val provider = CachedCredentialsProvider(chain)
+
+    override suspend fun getCredentials(): Credentials = provider.getCredentials()
+
+    override fun close() {
+        chain.close()
+        if (manageEngine) {
+            httpClientEngine.close()
         }
     }
```
ianbotsf, line 71: **Comment**: Maybe this is part of the open changes to make regarding ownership but I'd expect a `CachedCredentialsProvider` to be `Closeable` and to close its underlying provider when closed.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/DefaultChainCredentialsProvider.kt
```diff
@@ -5,28 +5,83 @@
...
+    private val platformProvider: PlatformProvider = Platform,
+    httpClientEngine: HttpClientEngine? = null
+) : CredentialsProvider, Closeable {
+
+    private val manageEngine = httpClientEngine == null
+    private val httpClientEngine = httpClientEngine ?: CrtHttpEngine()
+
+    public constructor() : this(Platform)
+
+    private val chain = CredentialsProviderChain(
```
ianbotsf, line 45: **Nit**: The secondary constructor seems like it belongs right after the primary constructor definition, not in between private `val`s.

ianbotsf (approved):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/DefaultChainCredentialsProvider.kt
aajtodd: Good point. Cached provider takes ownership so that should just be fixed here probably

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CredentialsProviderChain.kt
```diff
@@ -27,14 +31,20 @@ public open class CredentialsProviderChain(
...
+                chainException.value.addSuppressed(ex)
             }
         }
 
-        throw ClientException("No credentials could be loaded from the chain: $this")
+        throw chainException.value
+    }
+
+    override fun close() {
+        providers.forEach { (it as? Closeable)?.close() }
```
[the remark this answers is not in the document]
aajtodd: Agreed. Will fix

aws-runtime/aws-config/jvm/test/aws/sdk/kotlin/runtime/auth/credentials/DefaultChainCredentialsProviderTest.kt
```diff
@@ -0,0 +1,245 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import kotlinx.serialization.json.*
+import java.io.File
+import java.nio.file.Paths
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+// TODO - refactor to make this work in common
```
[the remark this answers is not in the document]
aajtodd: The idea would be like a kmp version of `loadResource` but it requires build script support. I've seen an example in the wild but yeah it's not straightforward.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CrtBasedCredentialsProvider.kt
aajtodd: We would need to discuss the plan for this. Just because SDK doesn't use them doesn't mean crt-kotlin can't support them (it is after all supposed to be bindings for crt for the language not SDK specific necessarily)

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CredentialsProviderChain.kt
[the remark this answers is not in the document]
aajtodd: What should the behavior be? Should we rethrow the first? Add all as suppressed exceptions?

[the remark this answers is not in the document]
aajtodd: Something like this?

```kotlin
    override fun close() {
        var closeEx: Throwable? = null
        providers.forEach {
            try {
                (it as? Closeable)?.close()
            }catch(ex: Exception) {
                if (closeEx == null) {
                    closeEx = ex
                }else {
                    closeEx!!.addSuppressed(ex)
                }
            }

            if (closeEx != null) throw closeEx!!
        }
    }

```

aws-runtime/aws-config/build.gradle.kts
aajtodd: I don't see anything that should be removed at a glance

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CredentialsProviderChain.kt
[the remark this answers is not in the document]
ianbotsf: Since the throw happens inside the loop, won't that still lead to unclosed providers?

I think we want something like:

```kotlin
val exceptions = providers.mapNotNull {
    try {
        (it as? Closeable)?.close()
        null
    } catch (t: Throwable) {
        t
    }
}
if (exceptions.isNotEmpty()) {
    val t = exceptions.first()
    exceptions.drop(1).forEach(t::addSuppressed)
    throw t
}
```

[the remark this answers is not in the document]
aajtodd: err whoops the throw should be outside the loop. Oddly enough I had something similar to what you are suggesting locally but opted for not allocating a list. Since you suggested it though I'm assuming you'd be in favor of it anyway. Unless anyone has an issue with it I'll just take Ian's suggestion.

PR #496, opened 2022-01-10
fix: work around machine-specific Gradle bug with aws-config variants

ianbotsf:
## Issue \#

(none)

## Description of changes

On my machine, Gradle fails to find the correct variant of **aws-config** for consuming packages. It fails with an error:

```
FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':services:dynamodb:compileJava'.
> Could not resolve all task dependencies for configuration ':services:dynamodb:compileClasspath'.
   > Could not resolve project :aws-runtime:aws-config.
     Required by:
         project :services:dynamodb
      > The consumer was configured to find an API of a library compatible with Java 8, preferably in the form of class files, preferably optimized for standard JVMs, and its dependencies declared externally, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'jvm'. However we cannot choose between the following variants of project :aws-runtime:aws-config:
          - apiElements
          - jvmApiElements
        All of them match the consumer attributes:
          - Variant 'apiElements' capability aws.sdk.kotlin:aws-config:0.9.6-SNAPSHOT declares an API of a library compatible with Java 8, packaged as a jar, and its dependencies declared externally:
              - Unmatched attributes:
                  - Doesn't say anything about its target Java environment (preferred optimized for standard JVMs)
                  - Doesn't say anything about org.jetbrains.kotlin.platform.type (required 'jvm')
          - Variant 'jvmApiElements' capability aws.sdk.kotlin:aws-config:0.9.6-SNAPSHOT declares an API of a library, packaged as a jar, preferably optimized for standard JVMs, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'jvm':
              - Unmatched attributes:
                  - Doesn't say anything about how its dependencies are found (required its dependencies declared externally)
                  - Doesn't say anything about its target Java version (required compatibility with Java 8)
        The following variants were also considered but didn't match the requested attributes:
          - Variant 'metadataApiElements' capability aws.sdk.kotlin:aws-config:0.9.6-SNAPSHOT declares a library, preferably optimized for non-jvm:
              - Incompatible because this component declares a usage of 'kotlin-api' of a component, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'common' and the consumer needed an API of a component, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'jvm'
              - Other compatible attributes:
                  - Doesn't say anything about how its dependencies are found (required its dependencies declared externally)
                  - Doesn't say anything about its target Java version (required compatibility with Java 8)
                  - Doesn't say anything about its elements (required them preferably in the form of class files)
```

Specifying that **apiElements** and **runtimeElements** cannot be consumed resolves the ambiguity.

### DIFF
--- a/aws-runtime/aws-config/build.gradle.kts
+++ b/aws-runtime/aws-config/build.gradle.kts
@@ -173,7 +173,7 @@ codegen.projections.all {
 }
 
 // Necessary to avoid Gradle problems identifying correct variant of aws-config. This stems from the smithy-gradle
-plugin (used by codegen plugin) applying the Java plugin which creates these configurations.
+// plugin (used by codegen plugin) applying the Java plugin which creates these configurations.
 listOf("apiElements", "runtimeElements").forEach {
     configurations.named(it) {
         isCanBeConsumed = false


aws-runtime/aws-config/build.gradle.kts
```diff
@@ -172,3 +171,10 @@ codegen.projections.all {
         kotlin.srcDir(projectedSrcDir)
     }
 }
+
+// Necessary to avoid Gradle problems identifying correct variant of aws-config
```
aajtodd, line 175: can you add a comment indicating where this stems from 

e.g.
`this stems from the smithy-gradle plugin (used by codegen plugin) applying the java plugin which creates these configurations`

aajtodd (approved):

PR #498, opened 2022-01-11
fix: credentials provider ownership

aajtodd:
## Issue \#

N/A

## Description of changes

* Adds a `borrow()` extension that wraps an existing provider with a no-op close implementation
* Close the client config `credentialsProvider` in the service client's close implementation (if it was borrowed it will do nothing, otherwise it will close it)
* ~Implement `Closeable` for shared `AwsClientConfig`~
* (breaking): remove `AwsClientConfig` and associated constructor
* (breaking): make `credentialsProvider` required for `PresignConfig` (otherwise it needs to be `Closeable` because it was instantiating a default chain provider)
* (refactor): inline region resolution into `Client.fromEnvironment()` and update it's signature to use `Client.Config.Builder` instead of `AwsClientLoadOptions`
* (refactor): thread `profileName` through `DefaultCredentialProviderChain` constructor. Make constructor with `httpClientEngine` override public.

### DIFF
--- a/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/PresignerGenerator.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/PresignerGenerator.kt
@@ -317,15 +317,15 @@ class PresignerGenerator : KotlinIntegration {
                 name = "credentialsProvider"
                 documentation = "The AWS credentials provider to use for authenticating requests. If not provided a [aws.sdk.kotlin.runtime.auth.credentials.DefaultChainCredentialsProvider] instance will be used."
                 baseClass = AwsRuntimeTypes.Signing.ServicePresignConfig
-                propertyType = ClientConfigPropertyType.RequiredWithDefault("DefaultChainCredentialsProvider()")
+                // TODO - we could probably relax this and instead create a default chain in createPresignedRequest on-demand in the runtime and close it when done signing
+                propertyType = ClientConfigPropertyType.Required()
             },
             ClientConfigProperty {
                 symbol = AwsRuntimeTypes.Endpoint.AwsEndpointResolver
                 name = "endpointResolver"
                 documentation = "Determines the endpoint (hostname) to make requests to. When not provided a default resolver is configured automatically. This is an advanced client option."
                 baseClass = AwsRuntimeTypes.Signing.ServicePresignConfig
-                // TODO - we could probably relax this and instead create a default chain in createPresignedRequest on-demand in the runtime and close it when done signing
-                propertyType = ClientConfigPropertyType.Required()
+                propertyType = ClientConfigPropertyType.RequiredWithDefault("DefaultEndpointResolver()")
             },
             ClientConfigProperty {
                 symbol = buildSymbol {
--- a/codegen/smithy-aws-kotlin-codegen/src/test/kotlin/aws/sdk/kotlin/codegen/PresignerGeneratorTest.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/test/kotlin/aws/sdk/kotlin/codegen/PresignerGeneratorTest.kt
@@ -225,7 +225,7 @@ class PresignerGeneratorTest {
              * instance is not available.
              */
             class TestPresignConfig private constructor(builder: Builder): ServicePresignConfig {
-                override val credentialsProvider: CredentialsProvider = builder.credentialsProvider ?: DefaultChainCredentialsProvider()
+                override val credentialsProvider: CredentialsProvider = requireNotNull(builder.credentialsProvider) { "credentialsProvider is a required configuration property" }
                 override val endpointResolver: AwsEndpointResolver = builder.endpointResolver ?: DefaultEndpointResolver()
                 override val normalizeUriPath: Boolean = true
                 override val region: String = requireNotNull(builder.region) { "region is a required configuration property" }


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/internal/BorrowedCredentialsProvider.kt
```diff
@@ -0,0 +1,27 @@
...
+private class BorrowedCredentialsProvider(
+    private val borrowed: CredentialsProvider
+) : CredentialsProvider by borrowed, Closeable {
+    override fun close() { }
+}
+
+/**
+ * Wraps another [CredentialsProvider] with a no-op close implementation. This inserts a level of indirection for
+ * use cases when a provider is explicitly given to the SDK, and it's ownership should remain with the caller.
+ * This allows the SDK to treat resources as owned and not have to track ownership state.
```
ianbotsf, line 21: **Nit**: it's → its

aws-runtime/aws-types/common/src/aws/sdk/kotlin/runtime/client/AwsClientConfig.kt
```diff
@@ -29,5 +28,8 @@ public interface AwsClientConfig {
      */
     public val sdkLogMode: SdkLogMode
 
+    // provide default no-op close
+    override fun close() { }
+
```
ianbotsf, line 33: **Comment**: This seems dangerous. Given the implications of forgetting to close something, it seems like there should be no default `close` and implementors should _have_ to make explicit decisions about closing.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/core/AwsHttpProtocolClientGenerator.kt
```diff
@@ -95,6 +95,16 @@ open class AwsHttpProtocolClientGenerator(
...
 
+    override fun renderClose(writer: KotlinWriter) {
+        writer.addImport(RuntimeTypes.IO.Closeable)
+        writer.write("")
+            .openBlock("override fun close() {")
+            .write("client.close()")
+            .write("(config.credentialsProvider as? #T)?.close()", RuntimeTypes.IO.Closeable)
+            .closeBlock("}")
+            .write("")
+    }
```
ianbotsf, line 106: **Comment**: Would be nice to have a new test for this.

ianbotsf (changes requested):
I like the general direction of this, it's nice and clean!

aws-runtime/aws-types/common/src/aws/sdk/kotlin/runtime/client/AwsClientConfig.kt
aajtodd: We could do that but every generated `Config` implements `AwsClientConfig` and so we would have to generate a `close()` method there. 

We could do that and just do `config.close()` instead in the generated client close implementation?

It would be nice to express a way to say that a property represents a closeable resource I guess and have the config generator automatically render it in the close implementation.

kggilmer: > It would be nice to express a way to say that a property represents a closeable resource I guess and have the config generator automatically render it in the close implementation.

+1.  Seems like going this way we'd be able to verify that a given codegen output cleans up resources as expected in CI

kggilmer (approved):

ianbotsf (approved):

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/core/AwsHttpProtocolClientGenerator.kt
```diff
@@ -95,6 +95,16 @@ open class AwsHttpProtocolClientGenerator(
         }
     }
 
+    override fun renderClose(writer: KotlinWriter) {
+        writer.addImport(RuntimeTypes.IO.Closeable)
+        writer.write("")
+            .openBlock("override fun close() {")
+            .write("client.close()")
+            .write("(config.credentialsProvider as? #T)?.close()", RuntimeTypes.IO.Closeable)
```
kggilmer, line 103: question

I'm missing something regarding `Closable` and `CredentialsProvider` types.  The latter doesn't specify a close method, but the `WrappedCredentialsProvider` does combine  the two, and looks here that we check to see if the cp implements closable and then call it if so.  Given this would it not be simpler to have CP simply implement `Closable`?  Put another way, what value are we gaining by having only `BorrowedCredentialsProvider` implement `Closable`?

kggilmer (approved):
Question regarding Closable and CredentialsProvider but nothing to warrant blocking the PR

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/core/AwsHttpProtocolClientGenerator.kt
aajtodd: > what value are we gaining by having only `BorrowedCredentialsProvider` implement `Closable`

It isn't the only one that implements `Closeable`. All of the following do:
* `DefaultChainCredentialsProvider`
* `EcsCredentialsProvider`
* `ImdsCredentialsProvider`
* `ProfileCredentialsProvider`

Environment, StsWebIdentity, StsAssumeRole, and SSO providers do not implement `Closeable`. 

The issue is that the type is just `CredentialsProvider` both here and in `BorrowedCredentialsProvider` so both have to check if the underlying implementation actually needs to be closed or not. I have gone back and forth on this and considered making `CredentialsProvider` implement `Closeable` but ultimately decided against it to be more clear about what types do and don't need to be closed when used directly by a customer. I could be persuaded the other way though if you want to make a case for it.

kggilmer: Personally I'm not sure what's best either way, but it may be more satisfying of "We ensure accessing AWS services is performant, secure, and reliable for developers" if `CredentialProvider` implements `Closable`.  I reckon this may be so because if closability varies, intuitively it seems likely that someone may forget that a particular CP needs to be closed.  Or perhaps a function that takes a non closing CP is passed a requires-closing CP during a refactor but the nuance is dropped in the refactor. If all CPs must be closed (with some of them being NOP), it seems to be less likely that this kind of bug would occur.  This comes at the expense of greater overall complexity of the CP type, however the value may outweigh the cost.  @ianbotsf any thoughts on this?

ianbotsf: I'm in favor of explicitly modelling closability. If it can/must be closed, it should be `Closeable`. Otherwise, I don't think we should unnecessarily attach closing semantics to the `CredentialProvider` interface because _some_ of the implementations are closeable.

aajtodd: Ya I can see both sides here and I hear the concern. I'm going to leave it as is for now and lets see how it shakes out.

PR #512, opened 2022-01-31
refactor(codegen): restructure protocol codegen to use structured parser generators

### LINKED ISSUE #449: Error metadata not set on unmodeled errors
## Describe the bug

When an operation deserializer encounters an unmodeled error it throws a generic service exception (e.g. `S3Exception` rather than `NoSuchBucketException`) it does not populate `sdkErrorMetadata`

## Expected Behavior

`sdkErrorMetadata` should be populated with protocol response, requestId, and other such metadata available from the response.

## Current Behavior

All of the metadata is missing.

## Steps to Reproduce

See #448 which demonstrates the issue and shows that in particular our documentation for debugging will not work:

```kotlin
    } catch (ex: AwsServiceException) {
        val httpResp = ex.sdkErrorMetadata.protocolResponse as? HttpResponse
        // Always null
        if (httpResp != null) {
            println(httpResp.headers)
            println(httpResp.body.readAll()?.decodeToString())
        }
        throw ex
    }
```

## Possible Solution

```kotlin
private suspend fun throwListObjectsError(context: ExecutionContext, response: HttpResponse): kotlin.Nothing {
    ...

    val modeledExceptionDeserializer = when(errorDetails.code) {
        "NoSuchBucket" -> NoSuchBucketDeserializer()
        else -> throw S3Exception(errorDetails.message)   // this needs to also call `setS3ErrorMetadata` before being thrown
    }

    val modeledException = modeledExceptionDeserializer.deserialize(context, wrappedResponse)
    setS3ErrorMetadata(modeledException, wrappedResponse, errorDetails)
    throw modeledException
}
```

## Your Environment

* AWS Kotlin SDK version used: `0.9.3-alpha`
* Operating System and version: MacOS 10.15.7

aajtodd:
## Issue \#

Related to: https://github.com/awslabs/smithy-kotlin/pull/579
closes #449 

## Description of changes

Refactored protocols to be implemented using the new `StructuredDataSerializerGenerator` and `StructuredDataParserGenerator)`

### DIFF
--- a/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3Generator.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3Generator.kt
@@ -61,26 +61,16 @@ class S3Generator : RestXml() {
             namespace = "${ctx.settings.pkg.name}.internal"
         }
 
-        listOf(
-            exceptionBaseSymbol,
-            RuntimeTypes.Http.readAll,
-            RuntimeTypes.Http.StatusCode,
-            AwsRuntimeTypes.Http.withPayload,
-            s3ErrorDetails,
-            setS3ErrorMetadata,
-            parseS3ErrorResponse,
-        ).forEach(writer::addImport)
-
-        writer.write("""val payload = response.body.readAll()""")
-            .write("val wrappedResponse = response.withPayload(payload)")
+        writer.write("val payload = response.body.#T()", RuntimeTypes.Http.readAll)
+            .write("val wrappedResponse = response.#T(payload)", AwsRuntimeTypes.Http.withPayload)
             .write("")
             .write("val errorDetails = try {")
             .indent()
             .call {
                 // customize error matching to handle HeadObject/HeadBucket error responses which have no payload
-                writer.write("if (payload == null && response.status == HttpStatusCode.NotFound) {")
+                writer.write("if (payload == null && response.status == #T.NotFound) {", RuntimeTypes.Http.StatusCode)
                     .indent()
-                    .write("""S3ErrorDetails(code = "NotFound")""")
+                    .write("#T(code = #S)", s3ErrorDetails, "NotFound")
                     .dedent()
                     .write("} else {")
                     .indent()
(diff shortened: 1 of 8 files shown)


codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/core/AwsHttpBindingProtocolGenerator.kt
```diff
@@ -142,7 +156,9 @@ abstract class AwsHttpBindingProtocolGenerator : HttpBindingProtocolGenerator()
             .write("")
 
         if (op.errors.isEmpty()) {
```
ianbotsf, line 158: **Comment**: Now that we're attaching metadata in all cases, I think we could simplify this code by moving the metadata/throwing to a single place:

```kotlin
writer.withBlock("val ex = when (errorDetails.code) {", "}") {
    op.errors.forEach { err ->
        val errSymbol = ctx.symbolProvider.toSymbol(ctx.model.expectShape(err))
        val errDeserializerSymbol = buildSymbol {
            name = "${errSymbol.name}Deserializer"
            namespace = "${ctx.settings.pkg.name}.transform"
        }
        write("#S -> #T.deserialize(context, wrappedResponse)", getErrorCode(ctx, err), errDeserializerSymbol)
    }
    
    write("else -> #T(errorDetails.message)", exceptionBaseSymbol)
}

writer.write("#T(ex, wrappedResponse, errorDetails)", AwsRuntimeTypes.Http.setAseErrorMetadata)
writer.write("throw ex")
```

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3Generator.kt
```diff
@@ -98,7 +98,9 @@ class S3Generator : RestXml() {
             .write("")
 
         if (op.errors.isEmpty()) {
```
ianbotsf, line 100: **Comment**: Duplicated code with **AwsHttpBindingProtocolGenerator.kt**.

ianbotsf (approved):

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3Generator.kt
aajtodd: Other than the improvements you suggested I'm going to leave some duplication here since the S3 error codegen is handling a few S3 specific behaviors

kggilmer (approved):
Loving all the red

PR #515, opened 2022-02-01
refactor: move request ID and error code metadata to smithy-kotlin

ianbotsf:
## Issue \#

Addresses smithy-kotlin#572

## Description of changes

**Companion PR**: smithy-kotlin#581

Moves exception metadata for request ID and error code out of **aws-sdk-kotlin** down into **smithy-kotlin**

### DIFF
--- a/aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/retries/AwsDefaultRetryPolicy.kt
+++ b/aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/retries/AwsDefaultRetryPolicy.kt
@@ -40,7 +40,7 @@ public object AwsDefaultRetryPolicy : StandardRetryPolicy() {
         504 to Timeout,
     )
 
-    override fun evaluateOtherExceptions(ex: Throwable): RetryDirective? = when (ex) {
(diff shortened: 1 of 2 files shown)


aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/retries/AwsDefaultRetryPolicy.kt
```diff
@@ -41,11 +41,11 @@ public object AwsDefaultRetryPolicy : StandardRetryPolicy() {
     )
 
     override fun evaluateOtherExceptions(ex: Throwable): RetryDirective? = when (ex) {
```
aajtodd, line 43: correctness

`StandardRetryPolicy` evaluates service exceptions, will we make it to this `evaluateOtherExceptions` for all of these known error codes? It looks like the base class checks `isRetryable` and the `errorType` and makes a decision. This may not be the correct `RetryError` directive for some of these if those fields are populated.

ianbotsf: Hmm, I think you're right. This looks like a miss in the original retries implementation. Fix incoming...

aajtodd (approved):

PR #535, opened 2022-02-18
fix: improve detection of available read bytes to avoid hang

### LINKED ISSUE #525: GetObject api with default http client hangs while downloading large file
### Describe the bug

When downloading large files(~120mb), the GetObject API with the default http client hangs, it downloads few mb and then doesn't rest of the file.
PS: Same api works as expected when used with Ktor client.

### Expected behavior

GetObject api should download the entire object without any pause.

### Current behavior

GetObject api with default http client pauses download after downloading few bytes while downloading large files. It is able to download small files correctly.

### Steps to Reproduce

1) Upload large file(>100mb) to S3 using AWS Console
2) Download file using GetObject api

        s3.getObject(GetObjectRequest {
                        key = INVALID_KEY
                        bucket = BUCKEY_NAME
                    }) { response -> ... 
        
         }
3) Observe the file progress using the device file manager on AS

### Possible Solution

_No response_

### Context

_No response_

0.12.0-beta

### Platform (JVM/JS/Native)

Native

Android 31

ianbotsf:
## Issue \#

Closes #525 

## Description of changes

This addresses a potential hang that occurs in CRT streaming code when the read continuation is set after a series of writes which consumed the entire window capacity. The fix here is a double-check before setting the continuation so that the code can avoid suspending if possible.

### DIFF
--- a/aws-runtime/http-client-engine-crt/common/src/aws/sdk/kotlin/runtime/http/engine/crt/AbstractBufferedReadChannel.kt
+++ b/aws-runtime/http-client-engine-crt/common/src/aws/sdk/kotlin/runtime/http/engine/crt/AbstractBufferedReadChannel.kt
@@ -78,14 +78,9 @@ internal abstract class AbstractBufferedReadChannel(
         }
 
         return suspendCancellableCoroutine { cont ->
-            if (availableForRead > 0) {
-                cont.resume(true)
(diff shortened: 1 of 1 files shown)


aajtodd (approved):
Couple suggestions but otherwise looks good.

PR #546, opened 2022-03-10
chore: cleanup presign tests

aajtodd:
Speedup presign tests and bucket deletion by order of magnitude + misc cleanup


services/s3/e2eTest/S3TestUtils.kt
```diff
@@ -40,15 +41,7 @@ object S3TestUtils {
...
-            do {
-                val bucketExists = try {
-                    client.headBucket { bucket = testBucket }
-                    true
-                } catch (ex: NotFound) {
-                    delay(300)
-                    false
-                }
-            } while (!bucketExists)
+            client.waitUntilBucketExists { bucket = testBucket }
```
ianbotsf, line 44: **Comment**: Nice!

PR #554, opened 2022-03-18
refactor!: replace default HTTP client engine

aajtodd:
## Issue \#

sibling: https://github.com/awslabs/smithy-kotlin/pull/606

## Description of changes

This PR swaps the default client HTTP engine from the CRT one to an opaque KMP compatible engine. The new default is one backed by Ktor/OkHttp. The existing CRT engine can still be used if desired but requires explicitly setting it.

Also in this PR is a modification of some of the e2e tests. The "big list of naughty" strings has been removed from S3 presign tests. These are not particularly representative of real object keys and are not recommended by S3 [guidelines on object key naming](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html). I've replaced it with a simpler test. We can add to this as needed. The presign tests now execute against the default and CRT engine to ensure we aren't introducing regressions across the HTTP engines we explicitly support.

# Breaking Change
This change is technically not a breaking change in API but may result in a change in behavior or exceptions seen by users. If you wish to continue using the CRT based HTTP engine you can explicitly configure it:

1. Add the dependency to your project e.g. `implementation("aws.smithy.kotlin:http-client-engine-crt:0.7.9")`
2. Explicitly provide the engine when creating a client
```kotlin
val engine = CrtEngine()
val s3 = S3Client {
    ...
    httpClientEngine = engine
}

...

// NOTE: in addition to the service client you are now responsible for the lifetime of the engine as well
s3.close()
engine.close()
```

### DIFF
--- a/tests/e2e-test-util/src/main/kotlin/aws/sdk/kotlin/testing/Utils.kt
+++ b/tests/e2e-test-util/src/main/kotlin/aws/sdk/kotlin/testing/Utils.kt
@@ -12,7 +12,7 @@ import aws.smithy.kotlin.runtime.http.engine.crt.CrtHttpEngine
 /**
  * Printable ASCII characters
  */
-val PRINTABLE_CHARS = ByteArray(127 - 32) { (it + 32).toByte() }.decodeToString()
+val PRINTABLE_CHARS = (32 until 127).map(Int::toChar).joinToString("")
 
 /**
  * Run the [block] with each supported engine


tests/e2e-test-util/src/main/kotlin/aws/sdk/kotlin/testing/Utils.kt
```diff
@@ -0,0 +1,31 @@
...
+package aws.sdk.kotlin.testing
+
+import aws.smithy.kotlin.runtime.http.engine.HttpClientEngine
+import aws.smithy.kotlin.runtime.http.engine.SdkHttpEngine
+import aws.smithy.kotlin.runtime.http.engine.crt.CrtHttpEngine
+
+/**
+ * Printable ASCII characters
+ */
+val PRINTABLE_CHARS = ByteArray(127 - 32) { (it + 32).toByte() }.decodeToString()
```
ianbotsf, line 15: **Comment**: Possibly cleaner:

```kotlin
val PRINTABLE_CHARS = (32..127).map(Int::toChar).joinToString("")
```

ianbotsf (approved):

Issue #557, opened 2022-03-21
Implement flexible checksums


ianbotsf:
### Describe the feature

S3 supports new [flexible checksum algorithm selection](https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/) via model updates which replace the `httpChecksumRequired` trait with `httpChecksum` and additional options. Fully implement support for user-selected checksum algorithms to enable this feature.

### Is your Feature Request related to a problem?

n/a

### Proposed Solution

n/a

### Describe alternative solutions or features you've considered

n/a

### Acknowledge

- [X] I may be able to implement this feature request

### AWS Kotlin SDK version used

0.13.1-beta

### Platform (JVM/JS/Native)

JVM

### Operating System and version

AL2

ianbotsf:
Actually, looks like this request is already captured in [smithy-kotlin#446](https://github.com/awslabs/smithy-kotlin/issues/446). Resolving as duplicate.

PR #558, opened 2022-03-21
fix: temporarily bypass httpchecksum traits until full flexible checksum support is available

ianbotsf:
## Issue \#

(none)

## Description of changes

Without this change, the latest S3 model cannot be included in the SDK since it replaces `httpChecksumRequred` with `httpChecksum` for flexible checksum support. This is a temporary workaround to allow consuming the latest S3 model. Full support for flexible checksums will come in #557.

Companion PR: [smithy-kotlin#608](https://github.com/awslabs/smithy-kotlin/pull/608)


aajtodd (approved):
Should we also drop the `requestValidationModeMember` if present?

ianbotsf:
> Should we also drop the `requestValidationModeMember` if present?

Yes, good catch.

PR #568, opened 2022-04-01
chore: update aws model filename pattern

aajtodd:
## Issue \#

N/A

## Description of changes

Updates AWS model filenames to use `<normalized-sdk-id>.json` instead of the current `{normalized-sdk-id}.{service-version}.json`. 

The normalization rules for using `sdkId` in filename are to replace spaces with `-` and lowercase (before spaces were just stripped and dashes removed). 

This is to align with other systems using this pattern and to make the automated merging of model files easy. Our previous use of this pattern was based on Go and there is no rhyme or reason for it (Go has also updated to this new pattern).

--------------
NOTE: Pay particular attention to package name change. We were deriving package name from `sdkId` in a very roundabout way before using the filename. I've updated it to use the `sdkId` directly from the model (hopefully) maintaining the same package names as before.

### DIFF
--- a/codegen/sdk/build.gradle.kts
+++ b/codegen/sdk/build.gradle.kts
@@ -91,7 +91,7 @@ data class AwsService(
      */
     val description: String? = null,
 
-    )
+)
 
 
 val disabledServices = setOf(


codegen/sdk/build.gradle.kts
```diff
@@ -91,16 +91,16 @@ data class AwsService(
      */
     val description: String? = null,
 
-)
+    )
```
ianbotsf, line 95: **Nit**: Incorrect indentation.

ianbotsf (approved):

Issue #567, opened 2022-03-31
Requests with no body don't sign correctly


lucix-aws:
### Describe the bug

`S3Client.createBucket`, when called with only a bucket name as input, appears to be signing incorrectly. The service will reject the request claiming that the signature is invalid.

### Expected behavior

Bucket should be created.

### Current behavior

`S3Exception` is thrown: `The request signature we calculated does not match the signature you provided. Check your key and signing method.`

### Steps to Reproduce

```
fun main():Unit = runBlocking {
    S3Client.fromEnvironment().use { s3 ->
        s3.createBucket { bucket = "buckettestdgttxcq" }

        s3.listBuckets {}.buckets?.forEach {
            println(it.name)
        }
    }
}
```

### AWS Kotlin SDK version used

0.14.0-beta

### Platform (JVM/JS/Native)

JVM

### Operating System and version

macOS 12.2.1

PR #570, opened 2022-04-04
fix: only set Content-Type when appropriate

lucix-aws:
Protocol tests for Content-Type. Also verifies Content-Length for both cases (body vs. empty).

## Issue \#
#567 

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -12,7 +12,7 @@ sdkVersion=0.14.3-SNAPSHOT
 smithyVersion=1.17.0
 smithyGradleVersion=0.5.3
 # smithy-kotlin codegen and runtime are versioned together
-smithyKotlinVersion=0.8.2
+smithyKotlinVersion=0.8.3-SNAPSHOT
 
 # kotlin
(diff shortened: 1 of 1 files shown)


ianbotsf (approved):

services/s3/model/s3-tests.smithy
```diff
@@ -63,3 +63,38 @@ apply PutObject @httpRequestTests([
...
 ])
+
+apply CreateBucket @httpRequestTests([
+    {
+        id: "CreateBucketNoBody",
+        documentation: "Validates https://github.com/awslabs/aws-sdk-kotlin/issues/567 (empty body)",
+        protocol: "aws.protocols#restXml",
+        method: "PUT",
+        uri: "/mybucket",
+        host: "s3.us-west-2.amazonaws.com",
```
aajtodd, line 74: We may want to be explicit or at least put a comment in here since we can't actually create this exact request in the real world (anything other than `us-east-1` IIRC requires the location constraint which results in a body)

aajtodd (approved):

Issue #581, opened 2022-04-13
Static profile credentials cannot be resolved if a region cannot be detected


ianbotsf:
### Describe the bug

When resolving profile credentials (e.g., in a **~/.aws/credentials** file), the credentials provider throws an exception if the default region resolver chain does successfully detect a region (e.g., in the profile, via env vars, etc.). This happens even if the profile credentials are purely static and don't require a region. Further note that an explicit region passed to a client constructor are unused in region resolution for the purposes of parsing profile credentials.

### Expected behavior

Users with static profile credentials should be able to use those credentials successfully without the region resolver needing to detect the region.

### Current behavior

Users with static profile credentials get an exception when the region resolver cannot detect the region.

### Steps to Reproduce

Create an AWS profile that explicitly specifies `aws_access_key_id` and `aws_secret_access_key` but **not** `region`. Also ensure that no region is available via `AWS_REGION` env var, `aws.region` system property, or via IMDS (e.g., on EC2). Then attempt to create and use an AWS client (e.g., S3):

```kotlin
S3Client
    .fromEnvironment {
        region = "us-west-2"
    }
    .listBuckets { } // throws exception because no credentials can be resolved
```

Workarounds that do succeed include adding `region` to the AWS profile or explicitly providing a region to an explicitly-chosen profile credentials provider:

```kotlin
S3Client
    .fromEnvironment {
        region = "us-west-2"
        credentialsProvider = ProfileCredentialsProvider(region = "us-west-2")
    }
    .listBuckets { }
```

### Possible Solution

Remove the requirement that a region resolves correctly for static credentials which do not require a region.

### AWS Kotlin SDK version used

0.14.2-beta

### Platform (JVM/JS/Native)

JVM

### Operating System and version

AL2

aajtodd:
closed by #582

PR #582, opened 2022-04-13
fix: resolve region only when profile credentials require it

ianbotsf:
## Issue \#

#581 

## Description of changes

This change removes the eager resolution of region in the profile credentials provider, which is unnecessary and can cause exceptions even when the region isn't needed (e.g., for static profile credentials).


aajtodd (approved):
Fix and ship!

PR #593, opened 2022-04-26
refactor!: separate hashing functions into new subproject

ianbotsf:
## Issue \#

Related to [smithy-kotlin#617](https://github.com/awslabs/smithy-kotlin/issues/617)

## Description of changes

Handles the reorganization of hashing functions in **smithy-kotlin**. Because of depending on a breaking change from **smithy-kotlin**, this is a **breaking change**.

**Companion PR**: [smithy-kotlin#632](https://github.com/awslabs/smithy-kotlin/pull/632)

### DIFF
--- a/.changes/30477ab1-a876-4671-9c89-fcfc3c6ff3b8.json
+++ b/.changes/30477ab1-a876-4671-9c89-fcfc3c6ff3b8.json
@@ -1,0 +1,5 @@
+{
+    "id": "30477ab1-a876-4671-9c89-fcfc3c6ff3b8",
+    "type": "documentation",
+    "description": "update API reference doc styling"
+}
--- a/aws-runtime/aws-config/build.gradle.kts
+++ b/aws-runtime/aws-config/build.gradle.kts
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0.
  */
 import aws.sdk.kotlin.gradle.codegen.dsl.smithyKotlinPlugin
+import org.jetbrains.dokka.gradle.DokkaTaskPartial
 
 plugins {
     id("aws.sdk.kotlin.codegen")
@@ -188,3 +189,13 @@ listOf("apiElements", "runtimeElements").forEach {
     }
 }
 
+
+// suppress internal generated clients
+tasks.named<DokkaTaskPartial>("dokkaHtmlPartial") {
+    dokkaSourceSets.configureEach {
+        perPackageOption {
+            matchingRegex.set(""".*\.internal.*""")
+            suppress.set(true)
+        }
+    }
+}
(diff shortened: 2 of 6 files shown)


aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/Prelude.kt
```diff
@@ -36,7 +36,7 @@ public data class Prelude(val totalLen: Int, val headersLength: Int) {
         preludeBuf.writeInt(headersLength)
 
         dest.writeFully(preludeBuf)
-        dest.writeInt(bytes.crc32().toInt())
+        dest.writeUInt(bytes.crc32())
```
aajtodd, line 39: Why did this change to `writeUInt`? This binary format is a defined specification and while I don't think it will matter in practice I do think we should match the spec.

aajtodd (approved):
Minor comment, otherwise looks good.

aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/Prelude.kt
ianbotsf: This functions identically to the previous version, it's just a code simplification. `UInt.toInt()` doesn't alter the underlying bits, it merely changes the interpretation of them, so writing a `UInt` to a binary buffer is the same as writing the converted `Int`. In fact, [`SdkByteBuffer.writeUInt`](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/io/common/src/aws/smithy/kotlin/runtime/io/SdkByteBuffer.kt#L278) delegates to [`Memory.storeUIntAt`](https://github.com/ktorio/ktor/blob/main/ktor-io/common/src/io/ktor/utils/io/bits/MemoryPrimitives.kt#L78) which performs this exact same conversion.

I originally made the new `crc32()` function return an `Int` but then reverted it back to `UInt` as in the prior implementation so I rewrote this line twice. Otherwise I wouldn't have changed it. But when I changed it the second time I opted for the cleaner version.

aajtodd: Ya as I said it won't matter in practice. I think I was more getting at "correctness at a glance". If I was reading this code and comparing it to the spec I'd have to dive into it and be like "why are we writing an unsigned int here instead of signed?" and consider the implications only to come to the same conclusion probably. It's a minor thing overall and not a hill I'm going to die on.

PR #610, opened 2022-05-20
feat: enhance generic codegen to be more KMP-friendly

### LINKED ISSUE #460: Use of library from commonMain in Kotlin Multiplatform Project
### Describe the issue

I can include following for example as `commonMain` dependency in Kotlin Multiplatform project and access/use code from JVM and Android code but if I try to use from `commonMain` source then only certain symbols appear to be resolvable (for example can access `StaticCredentialsProvider` but not `DynamoDbClient`).  Perhaps this is expected right now?

```
api("aws.sdk.kotlin:dynamodb:0.9.4-beta")
```

### Steps to Reproduce

per description

### Current behavior

per description

0.9.4

### Platform (JVM/JS/Native)

JVM/Android/Common

macOS 11.4

ianbotsf:
## Issue \#

Closes #460 

Switches to KMP codegen for service clients.

**Companion PR**: [smithy-kotlin#646](https://github.com/awslabs/smithy-kotlin/pull/646)

## Description of changes

### DIFF
--- a/aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/middleware/AwsRetryMiddleware.kt
+++ b/aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/middleware/AwsRetryMiddleware.kt
@@ -1,0 +1,46 @@
+package aws.sdk.kotlin.runtime.http.middleware
+
+import aws.smithy.kotlin.runtime.http.middleware.Retry
+import aws.smithy.kotlin.runtime.http.operation.*
+import aws.smithy.kotlin.runtime.http.request.header
+import aws.smithy.kotlin.runtime.io.Handler
+import aws.smithy.kotlin.runtime.retries.RetryStrategy
+import aws.smithy.kotlin.runtime.retries.policy.RetryPolicy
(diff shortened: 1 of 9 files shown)


services/s3/e2eTest/S3TestUtils.kt
```diff
@@ -34,6 +34,8 @@ object S3TestUtils {
 
         if (testBucket == null) {
             testBucket = prefix + UUID.randomUUID()
+            println("Creating S3 bucket: $testBucket")
```
aajtodd, line 37: probably be good to add an else branch and print that we are re-using an existing test bucket (which could happen if delete fails or tests are ended abruptly, etc).

aajtodd (approved):
Looks great. Fix and ship.

NOTE: This should be considered a breaking change and we probably want a more detailed description in the changelog about it. The reason being that the jvm artifact is now `-jvm` (e.g. `s3-jvm`) and not the existing artifact (e.g. `s3`) which is now `common`.

If you are _not_ using gradle (or the kotlin plugin) you need to pull in the `-jvm` artifact now. This _should_ be handled automatically for kotlin gradle plugin users.

PR #624, opened 2022-06-03
test: add multipart upload e2e test coverage

aajtodd:
## Issue \#

N/A
## Description of changes

Added a new e2e test for multipart uploads. I often want to test this feature and always have to go looking for my example. Figure we might as well just add it to our test suite.

### DIFF
--- a/services/s3/e2eTest/S3IntegrationTest.kt
+++ b/services/s3/e2eTest/S3IntegrationTest.kt
@@ -215,18 +215,10 @@ class S3BucketOpsIntegrationTest {
 }
 
 // generate sequence of "chunks" where each range defines the inclusive start and end bytes
-private fun File.chunk(partSize: Int): Sequence<LongRange> {
-    val totalSize = length()
-    val firstEndExclusive = minOf(partSize.toLong() + 1, totalSize)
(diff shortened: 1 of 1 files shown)


services/s3/e2eTest/S3IntegrationTest.kt
```diff
@@ -150,6 +155,77 @@ class S3BucketOpsIntegrationTest {
...
+    val firstEndExclusive = minOf(partSize.toLong() + 1, totalSize)
+    val seed = 0 until firstEndExclusive
+    return generateSequence(seed) {
+        if (it.last < totalSize - 1) {
+            it.last + 1 until minOf(it.last + partSize.toLong() + 1, totalSize)
+        } else {
+            null
+        }
+    }
 }
```
ianbotsf, line 229: **Nit**: This implementation actually grabs `partSize + 1` bytes for the first chunk and then `partSize` bytes for every subsequent chunk.

Could be rewritten/simplified to:

```kotlin
fun File.chunk(size: Long) =
    (0 until length() step size).map {
        it until minOf(it + size, length())
    }
```

ianbotsf (approved):

Issue #543, opened 2022-03-09
Event Stream Support


aajtodd:
### Describe the feature

Tracking issue for full support of event streams.

* [Spec](https://awslabs.github.io/smithy/1.0/spec/core/stream-traits.html#event-streams)
* [API Design](https://github.com/awslabs/smithy-kotlin/blob/main/docs/design/event-streams.md)

Completed:
- [x] Codegen support for shape serialization/deserialization to the `vnd.amazon.event-stream` content type
- [x] Runtime support for framing of `vnd.amazon.event-stream` content type
- [x] Integration tests for serialization and deserialization of all message types (except unbound document members) 
- [x] Lift all errors out of the event stream 
- [x] Response streams (e.g. s3 select object) work
- [x] Track initial request signature in the execution context 

Remaining body of work:

- [x] HTTP/2 full duplex stream support of underlying HTTP engine(s)
    - https://github.com/awslabs/smithy-kotlin/pull/694
    - https://github.com/awslabs/aws-sdk-kotlin/pull/687
- [x] Signing support for event stream request messages (details below) 
        - #644
        -  https://github.com/awslabs/smithy-kotlin/pull/670
- [X] Provide explicit coroutine scope for operations to launch background work
        - https://github.com/awslabs/smithy-kotlin/pull/696
        - https://github.com/awslabs/aws-sdk-kotlin/pull/689
- [ ] RPC support (initial request/response comes from the event stream rather than the HTTP request/response)
        - Deferred until later: https://github.com/awslabs/aws-sdk-kotlin/issues/693
- [x] Serialization/deserialization support for "unbound" document members (details below)
        - https://github.com/awslabs/aws-sdk-kotlin/pull/690
- [ ] Testing
  - [ ] Timeouts (add tests as necessary to verify our timeout configuration doesn't adversely affect streaming request/responses)
- [ ] Provide new examples for streaming operations
- [ ] Add service specific tests (integration, e2e, or replay) for all streaming operations 

### Signing

We are currently using the `AwsSignatureType.HTTP_REQUEST_CHUNK`

this results in string to sign of:
```
[INFO] [2022-03-09T01:35:43Z] [000070000cbd5000] [AuthSigning] - (id=0x7fae06a35dd0) Signing successfully built string-to-sign via algorithm SigV4, with contents                                                
AWS4-HMAC-SHA256-PAYLOAD                                                                                                                                                                                         
19731129T213309Z                                                                                                                                                                                                 
[...]

PR #644, opened 2022-06-30
feat(rt): add support for HTTP_REQUEST_EVENT signing

aajtodd:
## Issue \#

#543 
sibling PR: https://github.com/awslabs/smithy-kotlin/pull/670
## Description of changes

* Fix event stream frame signing envelope `:chunk-signature` header. This is expected to be the raw signature bytes, NOT the hex encoding of the signature.
* Fix frame decoding to not throw an exception when channel is closed and no more frames are coming
* Generate the required signing attributes into the `ExecutionContext`

### DIFF
--- a/aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/FrameDecoder.kt
+++ b/aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/FrameDecoder.kt
@@ -63,11 +63,11 @@ private suspend fun readPrelude(chan: SdkByteReadChannel): ByteArray? {
         remaining -= rc
     }
 
-    // partial read -> failure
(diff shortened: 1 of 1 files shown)


lucix-aws (approved):

aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/FrameDecoder.kt
```diff
@@ -52,3 +47,27 @@ public suspend fun decodeFrames(chan: SdkByteReadChannel): Flow<Message> = flow
...
+    var offset = 0
+    while (remaining > 0 && !chan.isClosedForRead) {
+        val rc = chan.readAvailable(dest, offset, remaining)
+        if (rc == -1) break
+        offset += rc
+        remaining -= rc
+    }
+
+    // partial read -> failure
+    if (remaining in 1 until PRELUDE_BYTE_LEN_WITH_CRC) throw EventStreamFramingException("failed to read event stream message prelude from channel: read: $offset bytes, expected $remaining more bytes")
```
ianbotsf, line 67: **Comment**: If this were swapped with the next if condition, it could be simplified to `if (remaining > 0)`.

ianbotsf (approved):

PR #651, opened 2022-07-11
Feat s3 transfermanager

wty-Bryant:
## Description of changes

### DIFF
--- a/hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/handler/Operation.kt
+++ b/hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/handler/Operation.kt
@@ -1,8 +1,9 @@
 package aws.sdk.kotlin.s3.transfermanager.handler
 
 import aws.sdk.kotlin.s3.transfermanager.data.Progress
+import kotlinx.coroutines.Deferred
 
-public interface Operation {
+public interface Operation : Deferred<Unit> {
 // inherits await(), cancel(), cancelAndJoin(), etc. from Deferred from Job
 
     // class reflecting transfer progress and mark whether it is completed


hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -19,31 +21,42 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+            // throw IllegalArgumentException if from path is invalid
+            require(localFile.exists()) { "From path is invalid" }
+
+            if (localFile.isFile()) {
+                uploadFile(localFile, to)
+            } else if (localFile.isDirectory()) {
+                uploadDirectory(localFile, to, progressListener)
+            } else {
+                throw IllegalArgumentException("From path is invalid")
+            }
```
ianbotsf, line 48: **Comment**: It looks like the return values from these methods are unused. I think that's right...we don't need them to create the top-level `Operation`. Given we're not using them, we don't need to create child `Operation` instances in these helper methods nor return anything from them.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/S3TransferManager.kt
```diff
@@ -60,12 +61,20 @@ public interface S3TransferManager {
...
+        progressListener: ProgressListener? = null, // can set a default value to Null
     ): Operation
 
+//    public fun upload(
+//        from: String,
+//        to: S3Uri,
+//        progressListener: ProgressListener? = null, // can set a default value to Null
+//        coroutineScope: CoroutineScope
+//    ): Operation
+
```
ianbotsf, line 77: **Nit**: Can remove now.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/handler/Operation.kt
```diff
@@ -1,9 +1,10 @@
 package aws.sdk.kotlin.s3.transfermanager.handler
 
 import aws.sdk.kotlin.s3.transfermanager.data.Progress
+import kotlinx.coroutines.Deferred
 
 // interface Operation: kotlinx.coroutines.Deferred<Unit> {
```
ianbotsf, line 6: **Nit**: Can remove now.

ianbotsf (changes requested):

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -115,20 +134,21 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
 //                   |_b.jpg	from: Users/direc1/direc2/b.jpg	to: key/direc2/b.jpg
 //               |_direc3	from: Users/direc1/direc3	to: key/direc3
 
-        val subFiles = localFile.listFiles()
-        subFiles.forEach {
-            val subFrom = localFile.toPath().resolve(it.name).toString()
+        val deferred = async {
```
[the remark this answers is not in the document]
LDsBF: Based on last comment, if we don't require helper function to return `Operation`, I agree we can delete such async in `uploadDirectory` method :)

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/S3TransferManager.kt
LDsBF: Sorry I forgot to delete it, I'll keep in mind later

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
LDsBF: I also considered whether to keep them for later child `Operation` merging to parent `Operation` during progress listening and controlling. But as long as top level `Operation` can handle these requirement, I will remove them for now.

ianbotsf: That's an interesting idea. It may turn out that way but I'm not sure yet whether that'll be necessary/helpful. Until we actually need it, it's better to remove it. We can always add it back in later!

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/handler/Operation.kt
ianbotsf: Whoops, that's not quite what I meant. The _commented-out code_ should be removed but the `Operation` interface _definitely needs to implement_ `Deferred<Unit>`. Callers to S3TM's methods will receive an `Operation` back. It's implemented by `DefaultOperation` but they won't know that and they won't be able to access methods from `Deferred` unless those methods also exist on `Operation`.

ianbotsf (changes requested):
Almost there, just a few minor changes and we can merge!

lucix-aws (approved):

Issue #509, opened 2022-01-25
Add Route53 Hosted Zone Trimming


aajtodd:
### Describe the bug

See https://github.com/awslabs/smithy-rs/pull/999 and https://github.com/awslabs/smithy-rs/pull/1472

### Expected behavior

See linked PR

### Current behavior

See linked PR

### Steps to Reproduce

See linked PR

### AWS Kotlin SDK version used

0.11.0-beta

### Platform (JVM/JS/Native)

JVM

### Operating System and version

N/A

aajtodd:
closed by #655

PR #655, opened 2022-07-15
feat: implement automatic prefix trimming for route53 resource ids

lucix-aws:
## Issue \#
#509 

## Description of changes
Trim route53 resource ID prefixes that are generally returned by the service and round-tripped in other requests where the value is URI-bound.

See the added `route53-tests` for input examples.

### DIFF
--- a/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/route53/TrimResourcePrefix.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/route53/TrimResourcePrefix.kt
@@ -62,4 +62,4 @@ private class TrimResourcePrefixMiddleware : ProtocolMiddleware {
 
 private fun MemberShape.shouldTrimResourcePrefix(): Boolean =
     (target.name == "ResourceId" || target.name == "ChangeId") &&
-    hasTrait<HttpLabelTrait>()
+        hasTrait<HttpLabelTrait>()


codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/route53/TrimResourcePrefix.kt
```diff
@@ -0,0 +1,66 @@
...
+                    "val updated = req.subject.copy { #L = req.subject.#L?.replace(prefix, #S) }",
+                    pathMember,
+                    pathMember,
+                    "",
+                )
+                write("next.call(#T(req.context, updated))", RuntimeTypes.Http.Operation.OperationRequest)
+            }
+            withBlock("else {", "}") {
+                write("next.call(req)")
+            }
```
ianbotsf, line 59: **Nit**: There's no need to `if` check here. Attempting to replace a pattern that doesn't match will have no effect:

```kotlin
val prefix = "^/?.*?/"
val updated = req.subject.copy { pathMember = req.subject.pathMember?.replace(prefix, "") }
next.call(OperationRequest(req.context, updated))
```

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/route53/TrimResourcePrefix.kt
```diff
@@ -0,0 +1,66 @@
...
+
+        writer.withBlock("op.execution.initialize.intercept { req, next -> ", "}") {
+            write("val prefix = #S.toRegex()", "^/?.*?/")
+            withBlock("if (req.subject.#L?.contains(prefix) == true) {", "}", pathMember) {
+                write(
+                    "val updated = req.subject.copy { #L = req.subject.#L?.replace(prefix, #S) }",
+                    pathMember,
+                    pathMember,
+                    "",
+                )
```
ianbotsf, line 54: **Style**: I personally find it easier to read codegen with fewer placeholder replacements. Consider using triple quotes instead of using `#S` with string literals:

```kotlin
write("""val prefix = "^/?.*?/".toRegex()""")
withBlock("if (req.subject.#L?.contains(prefix) == true) {", "}", pathMember) {
    write(
        """val updated = req.subject.copy { #L = req.subject.#L?.replace(prefix, "") }""",
        pathMember,
        pathMember,
    )
```

services/route53/model/route53-tests.smithy
```diff
@@ -0,0 +1,81 @@
...
+        documentation: "This test validates that delegation set id is correctly trimmed",
+        method: "GET",
+        protocol: "aws.protocols#restXml",
+        uri: "/2013-04-01/delegationset/DELEGATIONSETID",
+        bodyMediaType: "application/xml",
+        params: {
+            "Id": "/delegationset/DELEGATIONSETID"
+        }
+    },
+])
```
ianbotsf, line 81: **Question**: How/why is `delegationset` being trimmed in this test?

ianbotsf (approved):

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/route53/TrimResourcePrefix.kt
lucix-aws: You avoid an object copy if you do the check, for what it's worth.

services/route53/model/route53-tests.smithy
lucix-aws: * how - the generated regex catches it like anything else
* why - Great question! this gets at the underlying issue that there's no real record of how this customization is meant to work other than how SDKs have implemented it. I learned today that this one in particular is pretty old, it goes back to 2012 in Java v1. Java v2's version basically came from porting over v1. There's apparently no reference for how it's supposed to work outside of how SDKs have written it.

This particular test case is one pulled from rust's codebase.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/route53/TrimResourcePrefix.kt
```diff
@@ -0,0 +1,66 @@
...
+        resolved: List<ProtocolMiddleware>
+    ): List<ProtocolMiddleware> =
+        resolved + TrimResourcePrefixMiddleware()
+}
+
+private class TrimResourcePrefixMiddleware : ProtocolMiddleware {
+    override val name: String
+        get() = "TrimResourcePrefix"
+
+    override val order: Byte = -128
```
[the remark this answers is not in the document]
aajtodd: Important to note that the middleware itself *also* is ordered by phase. The `initialization` phase precedes `serialization` so it wouldn't matter what order they are in for codegen.

PR #656, opened 2022-07-18
Add first version S3 transfer manager download feature, may include some import* issue caused by new incline function sorry :(

wty-Bryant:
## Description of changes

### DIFF
--- a/hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
+++ b/hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
@@ -98,6 +98,7 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
                 bucket = to.bucket
                 key = to.key
             }
+
             val completedParts = mutableListOf<CompletedPart>()
 
             // call uploadPart() iteratively to continue uploading


hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -140,9 +147,112 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+                        bucket = from.bucket
+                        key = from.key
+                    }
+
+                    // if object exist, download that object
+                    validFrom = true
+                    val subTo = localFile.toPath().resolve(from.key.substringAfterLast('/')).toString()
+                    downLoadFile(from, subTo)
+                } catch (_: Exception) {
+                }
```
ianbotsf, line 196: **Comment**: We should generally catch the most specific exception we can. In this case, we specifically want to handle a `NotFound` exception. As written, this will catch (and ignore) other exceptions like authentication problems, filesystem errors, etc.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -140,9 +147,112 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+            }
```
ianbotsf, line 197: **Correctness**: This code will download a file and then continue on to the rest of the method below, including attempting to download a directory. I think we're missing a `return` here or a better structuring of `if` branches.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -140,9 +147,112 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+
+            // check if the current key is a keyPrefix
+            var keyPrefix = from.key
+            if (!keyPrefix.endsWith('/')) {
+                keyPrefix = keyPrefix.plus('/')
+            }
```
ianbotsf, line 203: **Nit**: This doesn't need to be a `var`:

```kotlin
val keyPrefix = if (keyPrefix.endsWith('/')) keyPrefix else keyPrefix + '/'
```

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -140,9 +147,112 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+            val response = s3.listObjectsV2Paginated {
+                bucket = from.bucket
+                prefix = keyPrefix
+            }
+
+            if (response.toList()[0].contents != null) {
+                validFrom = true
+            }
```
ianbotsf, line 211: **Correctness**: Calling `toList()` on a `Flow` suspends until all of the items are collected. In our case, that means repeatedly calling `listObjectsV2` for subsequent pages until no more are found. This could involve dozens or hundreds of calls so we shouldn't do that here.

Instead, what we really want to know is whether there's at least one page with at least one item in the `contents`. We can accomplish that by:

```kotlin
if (response.firstOrNull()?.contents?.isNotEmpty() == true) {
    ...
}
```

The `firstOrNull` call gets the very first page or null if there are no pages. The `isNotEmpty()` call is necessary because the `listObjectsV2` API may return a response with 0 items in it. In that case, `contents` wouldn't be null.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -140,9 +147,112 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+
+            response.collect {
+                downloadDirectory(it, localFile, progressListener)
+            }
+
+            if (!validFrom) {
+                throw IllegalArgumentException("From S3 uri contains invalid bucket/key/keyPrefix")
+            }
```
ianbotsf, line 219: **Comment**: This code works but it could be better organized to take advantage of flows. We don't need to process each page of results individually in `downloadDirectory` (which doesn't download _a directory_ but actually processes _a page_). Instead we can flat map the flow into the individual list of objects and process each of those directly:

```kotlin
response // Flow<ListObjectsV2Response>, a collection of pages
    .transform { it.contents?.forEach { obj -> emit(obj) } } // Flow<Object>, a collection of individual objects
    .collect { obj ->
        val s3Uri = S3Uri(from.bucket, obj.key) // Form a new S3Uri out of the requested bucket and the object key
        val keySuffix =  obj.key.substringAfter(keyPrefix) // The unique part of the key after the prefix
        val subPath = Paths.get(to, keySuffix).toString() // Concat the original `to` with the key to get a destination file
        downloadFile(s3Uri, subPath)
    }
```

(Comments added for explanation. They're not necessary to duplicate in the actual PR.)

Note that this code assumes that `downloadFile` is correctly creating destination directories. See comment below.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -140,9 +147,112 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+     */
+    context(CoroutineScope)
+    override fun download(from: S3Uri, to: String, progressListener: ProgressListener?): Operation {
+        val deferred = async {
+            var validFrom = false
+            val localFile = File(to)
+            // create the target directory if to path doesn't exist
+            if (!localFile.exists() || !localFile.isDirectory()) {
+                withContext(Dispatchers.IO) {
+                    Files.createDirectories(Paths.get(to))
```
ianbotsf, line 179: **Comment**: Given that this needs to happen for every downloaded file, it probably belongs in the `downloadFile` method instead.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -140,9 +147,112 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+    private fun downLoadFile(from: S3Uri, to: String) {
+        async {
+            val request = GetObjectRequest {
+                bucket = from.bucket
+                key = from.key
+            }
+            s3.getObject(request) { resp ->
+                val rc = resp.body?.writeToFile(Paths.get(to))
+                rc
+            }
```
ianbotsf, line 235: **Correctness**: The `downloadFile` method itself should be creating subdirectories as necessary before downloading the file:

```kotlin
s3.getObject(request) { resp ->
    val toPath = Paths.get(to)
    Files.createDirectories(toPath.parent)
    resp.body?.writeToFile(toPath)
}
```

ianbotsf, line 235: **Nit**: We never use the value `rc` so there's no reason to capture it as a `val`.

ianbotsf (changes requested):

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -140,9 +147,112 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+        }
+    }
+
+    context(CoroutineScope)
+    private fun downloadDirectory(response: ListObjectsV2Response, localFile: File, progressListener: ProgressListener?) {
+//     to recursively execute single object download or sub-directory download at local next level
+
+        response.contents?.forEach {
+            val subFrom = response.prefix?.let { it1 -> it.key?.substringAfter(it1) }
+            if (subFrom != null) {
```
lucix-aws, line 245: Let's return early (or filter these items we want to skip out first) instead of wrapping the whole loop body.

eg.
```kotlin
val subFrom = response.prefix?.let { it1 -> it.key?.substringAfter(it1) }
if (subFrom == null) {
    continue
}

// rest of loop body
```

hlls/s3-transfermanager/e2eTest/S3TransferManagerTestUtils.kt
```diff
@@ -4,7 +4,7 @@
  */
 package aws.sdk.kotlin.e2etest
 
-import aws.sdk.kotlin.services.s3.S3Client
+import aws.sdk.kotlin.services.s3.*
```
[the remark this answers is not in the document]
LDsBF: I try to add their import one by one, but compiler automatically merge them into single * import when I added too many separate import lol. I will separate their import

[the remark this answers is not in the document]
ianbotsf: That isn't the compiler, it's IntelliJ. You can configure the import settings under **File** → **Settings** → **Editor** → **Code Style** → **Kotlin** → **Imports**.

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
```diff
@@ -91,4 +89,13 @@ class S3TransferManagerIntegrationTest {
...
         }
     }
+
+    @Test
+    fun testDownload() = runTest {
+        val s3Uri = S3Uri(testBucket, "folder1")
+        s3TransferManager.upload(testDirectory.toString(), s3Uri)
+
+        val operation = s3TransferManager.download(s3Uri, testDirectory.toString())
+        assertNotNull(operation, "The transfer manager doesn't tackle directory download")
```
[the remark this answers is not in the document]
LDsBF: I just added a `operation.await()` after `assertNotNull` to wait for async download to complete in case of any coroutine's exception or error. Not sure whether it can solve.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -140,9 +147,112 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+                    s3.headObject {
+                        bucket = from.bucket
+                        key = from.key
+                    }
+
+                    // if object exist, download that object
+                    validFrom = true
+                    val subTo = localFile.toPath().resolve(from.key.substringAfterLast('/')).toString()
+                    downLoadFile(from, subTo)
+                } catch (_: Exception) {
```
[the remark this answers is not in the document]
LDsBF: This `try {} catch () {}` is used to catch `NotFound` exception from `s3.headObject`, which throw that exceotion when no object with given bucket/key is found. Since we need to continue following s3 key prefix directory searching, `IllegalArgumentException` will be thrown later if the key refers to nothing.

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
```diff
@@ -96,6 +97,7 @@ class S3TransferManagerIntegrationTest {
         s3TransferManager.upload(testDirectory.toString(), s3Uri)
 
         val operation = s3TransferManager.download(s3Uri, testDirectory.toString())
-        assertNotNull(operation, "The transfer manager doesn't tackle directory download")
+        assertNotNull(operation, "The transfer manager didn't start directory download")
+        operation.await()
```
ianbotsf, line 101: **Question**: Does this work consistently? The `upload` operation is asynchronous too so it's possible for the download to begin or even complete before the upload is completed. We probably should `await` the completion of the upload before we begin downloading.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -44,12 +49,10 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
-                uploadFile(localFile, to)
-            } else if (localFile.isDirectory()) {
-                uploadDirectory(localFile, to, progressListener)
-            } else {
-                throw IllegalArgumentException("From path is invalid")
+            when {
+                localFile.isFile() -> uploadFile(localFile, to)
+                localFile.isDirectory() -> uploadDirectory(localFile, to, progressListener)
+                else -> throw IllegalArgumentException("From path is invalid")
             }
```
ianbotsf, line 56: **Comment**: Nice!

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -190,65 +194,79 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
 
                     // if object exist, download that object
                     validFrom = true
-                    val subTo = localFile.toPath().resolve(from.key.substringAfterLast('/')).toString()
-                    downLoadFile(from, subTo)
-                } catch (_: Exception) {
+                    val subTo = Paths.get(to).resolve(from.key.substringAfterLast('/')).toString()
+                    downloadFile(from, subTo)
+                } catch (_: NotFound) {
                 }
```
ianbotsf, line 200: **Correctness**: We should move this `catch` to just surround the `headObject` call. We want to silently ignore failures there but not in any of our subsequent code.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -190,65 +194,79 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
             }
 
-            // check if the current key is a keyPrefix
-            var keyPrefix = from.key
-            if (!keyPrefix.endsWith('/')) {
-                keyPrefix = keyPrefix.plus('/')
+            if (validFrom) { // if from refers to specific object, just download it and return
+                return@async
             }
```
ianbotsf, line 205: **Comment**: The only place that we set `validFrom` before this `if` block is when we have a file to download. We should just `return` immediately after `downloadFile` and not bother with `validFrom`.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
ianbotsf: Did you have a chance to look into this suggestion? It may simplify the code but I'm not sure if it covers every case or if there's bugs in it.

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
LDsBF: agree

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -185,25 +183,13 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
-                } catch (_: NotFound) {
+                    return@async
                 }
             }
-
-            if (validFrom) { // if from refers to specific object, just download it and return
-                return@async
-            }
-
+            println("Continue!")
```
ianbotsf, line 192: **Nit**: Let's not commit `println`s in business logic.

ianbotsf (commented):
I like how much cleaner this is now that `objectExists` is a separate function! Great job!

Did you see [my comment about flow transformation yet](https://github.com/awslabs/aws-sdk-kotlin/pull/656#discussion_r923655681)? Have you had a chance to play with that?

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -212,22 +198,36 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
-            if (!validFrom) {
-                throw IllegalArgumentException("From S3 uri contains invalid key/keyPrefix")
-            }
         }
 
         return DefaultOperation(deferred)
     }
 
+    context(CoroutineScope)
+    private suspend fun objectExists(s3Uri: S3Uri): Boolean {
```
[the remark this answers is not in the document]
LDsBF: No, I delete it and everything still works

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -140,9 +150,103 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+            }
+            response // Flow<ListObjectsV2Response>, a collection of pages
+                .transform { it.contents?.forEach { obj -> emit(obj) } }
+                .collect { obj ->
+                    val s3Uri = obj.key?.let { S3Uri(from.bucket, it) }
+                    val keySuffix = obj.key?.substringAfter(keyPrefix)
+                    val subTo = Paths.get(to, keySuffix).toString()
+                    if (s3Uri != null) {
+                        downloadFile(s3Uri, subTo)
+                    }
```
ianbotsf, line 209: **Comment**: I believe that `key`'s nullability is a limitation of our codegen and not an actual possibility in responses from the service (at least, I don't understand what an object with a null key would mean semantically). I suggest you non-null assert the key inside this block and then remove the null checking elsewhere.

ianbotsf (approved):

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
```diff
@@ -75,20 +75,30 @@ class S3TransferManagerIntegrationTest {
...
         }
     }
+
+    @Test
+    fun testDownload() = runTest {
+        val s3Uri = S3Uri(testBucket, "folder1")
+        var operation = s3TransferManager.upload(testDirectory.toString(), s3Uri)
+        operation.await()
+
+        operation = s3TransferManager.download(s3Uri, testDirectory.toString())
```
lucix-aws, line 100: I'm still not a huge fan of these tests because they're not really verifying the end-to-end behavior here. Sure, you're calling upload and download, and I assume the methods execute, but we should verify that the contents we upload are really there in s3, and the contents we download are on the filesystem how we expect.

lucix-aws (approved):

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
LDsBF: I will add `headObject()` amd `FIle(downloadedFile)`to test them in next PR

PR #661, opened 2022-07-25
Add more logic of multipart download and e2e tests

wty-Bryant:
## Description of changes

### DIFF
--- a/hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
+++ b/hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
@@ -5,7 +5,10 @@ import aws.sdk.kotlin.s3.transfermanager.data.S3Uri
 import aws.sdk.kotlin.s3.transfermanager.headObjectOrNull
 import aws.sdk.kotlin.services.s3.S3Client
 import aws.sdk.kotlin.services.s3.listObjectsV2
+import aws.sdk.kotlin.services.s3.model.GetObjectRequest
 import aws.sdk.kotlin.services.s3.paginators.listObjectsV2Paginated
+import aws.smithy.kotlin.runtime.content.asByteStream
+import aws.smithy.kotlin.runtime.content.toByteArray
 import aws.smithy.kotlin.runtime.testing.RandomTempFile
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.coroutineScope
@@ -51,18 +54,25 @@ class S3TransferManagerIntegrationTest {
 
     @BeforeEach
     private fun createResources(): Unit = runBlocking {
+        testBucket = S3TransferManagerTestUtils.getTestBucket(s3TransferManager.config.s3)
+    }
+
+    @AfterEach
+    private fun cleanup() = runBlocking {
+        S3TransferManagerTestUtils.deleteBucketAndAllContents(s3TransferManager.config.s3, testBucket)
+    }
+
+    private fun createUploadDirectory() {
+        //        test directory structure like this:
+        //        testUploadDirectory/
+        //            file1.txt
+        //            testUploadDirectory1/
+        //                file2.png
+        //                file3.jpeg
+        //            testUploadDirectory2/
         val home: String = System.getProperty("user.home")
         val dir = Paths.get(home, "Downloads")
-        testBucket = S3TransferManagerTestUtils.getTestBucket(s3TransferManager.config.s3)
         testUploadDirectory = Files.createTempDirectory(dir, "testUploadDirectory")
-        testDownloadDirectory = Files.createTempDirectory(dir, "testDownloadDirectory")
-//        test directory structure like this:
-//        testUploadDirectory/
-//            file1.txt
-//            testUploadDirectory1/
-//                file2.png
-//                file3.jpeg
-//            testUploadDirectory2/
         File.createTempFile("file1", ".txt", testUploadDirectory.toFile())
         val testUploadDirectory1 = Files.createTempDirectory(testUploadDirectory, "testUploadDirectory1")
         File.createTempFile("file2", ".png", testUploadDirectory1.toFile())
@@ -70,31 +80,31 @@ class S3TransferManagerIntegrationTest {
         Files.createTempDirectory(testUploadDirectory, "testUploadDirectory2")
     }
 
-    @AfterEach
-    private fun cleanup() = runBlocking {
-        deleteFiles(testUploadDirectory.toFile())
-        deleteFiles(testDownloadDirectory.toFile())
-        S3TransferManagerTestUtils.deleteBucketAndAllContents(s3TransferManager.config.s3, testBucket)
+    private fun createDownloadDirectory() {
+        val home: String = System.getProperty("user.home")
+        val dir = Paths.get(home, "Downloads")
+        testDownloadDirectory = Files.createTempDirectory(dir, "testDownloadDirectory")
     }
 
-    private fun deleteFiles(file: File) {
-        if (file.isDirectory) {
-            val subFiles = file.listFiles()
-            subFiles.forEach {
-                deleteFiles(it)
+    private fun File.deleteRecursive() {
+        if (isDirectory) {
+            listFiles().forEach {
+                it.deleteRecursive()
             }
         }
-        file.delete()
+        delete()
     }
 
     @Test
     fun testUpload() = runTest {
+        createUploadDirectory()
         val keyPrefix = "folder1"
         val toUri = S3Uri(testBucket, keyPrefix)
         val operation = s3TransferManager.upload(testUploadDirectory.toString(), toUri)
         assertNotNull(operation, "The transfer manager didn't start directory upload")
         operation.await()
         assertTrue(checkUpload(testUploadDirectory.toFile(), toUri))
+        testUploadDirectory.toFile().deleteRecursive()
     }
 
     @Test
@@ -106,12 +116,15 @@ class S3TransferManagerIntegrationTest {
         assertNotNull(operation, "The transfer manager didn't start parts upload")
         operation.await()
         assertTrue(checkUpload(testLargeFile, toUri))
-        testLargeFile.delete()
+        testLargeFile.deleteRecursive()
     }
 
     private suspend fun checkUpload(localFile: File, to: S3Uri): Boolean {
         when {
             localFile.isFile -> {
+                if (localFile.length() > s3TransferManager.config.chunkSize) {
+                    return chunksCompare(localFile, to)
+                }
                 return s3TransferManager.config.s3.headObjectOrNull(to) != null
             }
 
@@ -150,14 +163,18 @@ class S3TransferManagerIntegrationTest {
 
     @Test
     fun testDownload() = runTest {
+        createUploadDirectory()
         val s3Uri = S3Uri(testBucket, "folder1")
         val uploadOperation = s3TransferManager.upload(testUploadDirectory.toString(), s3Uri)
         uploadOperation.await()
+        testUploadDirectory.toFile().deleteRecursive()
 
+        createDownloadDirectory()
         val downloadOperation = s3TransferManager.download(s3Uri, testDownloadDirectory.toString())
         assertNotNull(downloadOperation, "The transfer manager didn't start directory download")
         downloadOperation.await()
         assertTrue(checkDownload(s3Uri, testDownloadDirectory.toFile()))
+        testDownloadDirectory.toFile().deleteRecursive()
     }
 
     @Test
@@ -167,14 +184,22 @@ class S3TransferManagerIntegrationTest {
         val s3Uri = S3Uri(testBucket, largeFileKey)
         val uploadOperation = s3TransferManager.upload(testLargeFile.path, s3Uri)
         uploadOperation.await()
+        testLargeFile.deleteRecursive()
+
+        createDownloadDirectory()
         val downloadOperation = s3TransferManager.download(s3Uri, testDownloadDirectory.toString())
         downloadOperation.await()
         val downloadFile = Paths.get(testDownloadDirectory.toString(), largeFileKey).toFile()
         assertTrue(checkDownload(s3Uri, downloadFile))
+        testDownloadDirectory.toFile().deleteRecursive()
     }
 
     private suspend fun checkDownload(from: S3Uri, localFile: File): Boolean {
-        if (!from.key.endsWith('/') && s3TransferManager.config.s3.headObjectOrNull(from) != null) {
+        val headObjectResponse = s3TransferManager.config.s3.headObjectOrNull(from)
+        if (!from.key.endsWith('/') && headObjectResponse != null) {
+            if (headObjectResponse.contentLength > s3TransferManager.config.chunkSize) {
+                return chunksCompare(localFile, from)
+            }
             return localFile.isFile()
         }
 
@@ -188,9 +213,10 @@ class S3TransferManagerIntegrationTest {
             .transform { it.contents?.forEach { obj -> emit(obj) } }
             .collect { obj ->
                 val key = obj.key!!
+                val subFrom = S3Uri(from.bucket, key)
                 val keySuffix = key.substringAfter(keyPrefix)
                 val subFile = Paths.get(localFile.toString(), keySuffix).toFile()
-                if (!subFile.isFile()) {
+                if (!checkDownload(subFrom, subFile)) {
                     checkResult = false
                     return@collect
                 }
@@ -201,10 +227,6 @@ class S3TransferManagerIntegrationTest {
 
     @Test
     fun testDownloadInvalidFromBucket() = runTest {
-        val s3Uri = S3Uri(testBucket, "folder1")
-        val operation = s3TransferManager.upload(testUploadDirectory.toString(), s3Uri)
-        operation.await()
-
         assertFailsWith<IllegalArgumentException>("The download is completed without throwing from bucket error") {
             coroutineScope {
                 s3TransferManager.download(S3Uri("s3://${testBucket}${Random.nextLong(Long.MAX_VALUE)}/folder1"), "/Users/wty/Desktop/folder1/haha").await()
@@ -214,14 +236,42 @@ class S3TransferManagerIntegrationTest {
 
     @Test
     fun testDownloadInvalidFromKey() = runTest {
-        val s3Uri = S3Uri(testBucket, "folder1")
-        val operation = s3TransferManager.upload(testUploadDirectory.toString(), s3Uri)
-        operation.await()
-
         assertFailsWith<IllegalArgumentException>("The download is completed without throwing from key error") {
             coroutineScope {
                 s3TransferManager.download(S3Uri(testBucket, "${Random.nextLong(Long.MAX_VALUE)}/${Random.nextInt(Int.MAX_VALUE)}"), "/Users/wty/Desktop/folder1/haha").await()
             }
         }
+    }
+
+    private suspend fun chunksCompare(localFile: File, s3Uri: S3Uri): Boolean {
+        val fileSize = localFile.length()
+        if (fileSize != s3TransferManager.config.s3.headObjectOrNull(s3Uri)?.contentLength) {
+            return false
+        }
+        val chunkRanges = (0 until fileSize step s3TransferManager.config.chunkSize).map {
+            it until minOf(it + s3TransferManager.config.chunkSize, fileSize)
+        }
+        var isEqual = true
+
+        chunkRanges.forEach {
+            val contentRange = "bytes=${it.first}-${it.last}"
+            val request = GetObjectRequest {
+                bucket = s3Uri.bucket
+                key = s3Uri.key
+                range = contentRange
+            }
+            s3TransferManager.config.s3.getObject(request) { resp ->
+                // compare s3 object chunk and local file's corresponding chunk after converting both to ByteArray
+                if (!resp.body?.toByteArray().contentEquals(localFile.asByteStream(it).toByteArray())) {
+                    isEqual = false
+                    return@getObject
+                }
+            }
+            if (!isEqual) {
+                return isEqual
+            }
+        }
+
+        return isEqual
     }
 }
--- a/hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
+++ b/hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
@@ -20,14 +20,15 @@ import aws.sdk.kotlin.services.s3.uploadPart
 import aws.smithy.kotlin.runtime.content.ByteStream
 import aws.smithy.kotlin.runtime.content.asByteStream
 import aws.smithy.kotlin.runtime.content.fromFile
-import aws.smithy.kotlin.runtime.content.toByteArray
+import aws.smithy.kotlin.runtime.io.SdkByteReadChannel
+import aws.smithy.kotlin.runtime.io.copyTo
+import aws.smithy.kotlin.runtime.io.writeChannel
+import aws.smithy.kotlin.runtime.util.InternalApi
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.async
 import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.flow.transform
 import java.io.File
-import java.io.RandomAccessFile
-import java.nio.ByteBuffer
 import java.nio.file.Files
 import java.nio.file.Paths
 import kotlin.io.path.createFile
@@ -53,7 +54,7 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
             val localFile = File(from)
             // throw IllegalArgumentException if from path is invalid
             require(localFile.exists()) { "From path is invalid" }
-
+            println("Checking file or directory....")
             when {
                 localFile.isFile() -> uploadFile(localFile, to)
                 localFile.isDirectory() -> uploadDirectory(localFile, to, progressListener)
@@ -72,6 +73,7 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
         // determine upload with single request or split parts request according to file size
         val fileSize = localFile.length()
         if (fileSize <= config.chunkSize) { // for file smaller than config chunk size
+            println("Downloading whole small file...")
             uploadWholeFile(localFile, to)
         } else { // for large file over config chunk size
             uploadFileParts(localFile, to)
@@ -81,6 +83,7 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
     context(CoroutineScope)
     private fun uploadWholeFile(localFile: File, to: S3Uri) {
         async<Unit> {
+            println("Start downloading from S3Client!!!")
             s3.putObject {
                 bucket = to.bucket
                 key = to.key
@@ -188,12 +191,15 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
                 throw IllegalArgumentException("The bucket does not exist or has no access to it")
             }
 
-            if (!from.key.endsWith('/') && s3.headObjectOrNull(from) != null) {
+            if (!from.key.endsWith('/')) {
                 val response = s3.headObjectOrNull(from)
-                val subTo = Paths.get(to).resolve(from.key.substringAfterLast('/')).toString()
-                downloadFile(response!!.contentLength, from, subTo)
-                return@async
+                if (response != null) {
+                    val subTo = Paths.get(to).resolve(from.key.substringAfterLast('/')).toString()
+                    downloadFile(response.contentLength, from, subTo)
+                    return@async
+                }
             }
+
             // check if the current key is a keyPrefix
             val keyPrefix = if (from.key.endsWith('/')) from.key else from.key.plus('/')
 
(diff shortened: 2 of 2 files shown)


hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -199,31 +209,32 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
-                    downloadFile(s3Uri, subTo)
+                    downloadFile(obj.size, s3Uri, subTo)
                 }
         }
 
         return DefaultOperation(deferred)
     }
 
-    private suspend fun objectExists(s3Uri: S3Uri): Boolean =
+    private suspend fun headObject(s3Uri: S3Uri): HeadObjectResponse? =
```
lucix-aws, line 227: This probably makes more sense to exist as an extension method on `S3Client`, which basically overloads `headObject` to take an `S3Uri`

ie.
```kotlin
// "headObjectOrNull" because it wraps the not-found exception logic to return null
// when the object doesn't exist - this is the "idiomatic" standard for
// "get-or-default"-type checks in kotlin
private suspend fun S3Client.headObjectOrNull(uri: S3Uri): HeadObjectResponse? {
    // as you have it now
}

File: hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt [hunk 2]
```diff
@@ -179,14 +184,19 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
-            } catch (e: Exception) {
+            } catch (e: S3Exception) {
                 throw IllegalArgumentException("The bucket does not exist or has no access to it")
             }
 
-            if (!from.key.endsWith('/') && objectExists(from)) {
-                val subTo = Paths.get(to).resolve(from.key.substringAfterLast('/')).toString()
-                downloadFile(from, subTo)
-                return@async
+            if (!from.key.endsWith('/')) {
```
lucix-aws, line 191: Not sure this needed to change? I see that you changed your function to check for whether an object exists, but even so I'm pretty sure you can just do
```kotlin
if (!from.key.endsWith('/') && headObject(from) != null) {
    // ...
}
```

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -179,14 +184,19 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+                val response = headObject(from)
+                when {
+                    response != null -> {
```
lucix-aws, line 194: Pretty sure this will go away based on the comment right above this one, but a `when` with a single outcome is basically just an `if` and should be written that way.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -232,7 +243,15 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
      * to refers to specific path ending with the file name
      */
     context(CoroutineScope)
-    private fun downloadFile(from: S3Uri, to: String) {
+    private fun downloadFile(fileSize: Long, from: S3Uri, to: String) {
+        when {
```
lucix-aws, line 247: This might go away, but for what it's worth, a `when` with one outcome and an `else` is basically just an `if`/`else` and should be written that way.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -247,6 +266,56 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+        }
+    }
+
+    private suspend fun downloadFilePart(objectContent: ByteStream, localFile: File, position: Long): Long {
+        val randomAccessFile = RandomAccessFile(localFile, "rw")
+        val channel = randomAccessFile.channel
+        channel.position(position)
+        val filePosition: Long
+
+        try {
```
lucix-aws, line 306: Do we need this `try`/`finally' for anything? The latter's block is empty.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
lucix-aws, line 247: This check may not be necessary.

Theoretically, if `fileSize <= config.chunkSize`, `downloadFileParts` should still be usable - it would just only map out to a single `downloadFilePart` call internally.

lucix-aws (changes requested):

hlls/s3-transfermanager/jvm/test/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManagerTest.kt
```diff
@@ -34,63 +44,102 @@ class DefaultS3TransferManagerTest {
...
 //    @Test
 //    fun testUpload() = runTest {
-//        val defaultPutObjectResponse = PutObjectResponse {}
 //        var testFile = RandomTempFile(6000000)
-////        coEvery {
-////            s3Client.putObject {
-////                body = ByteStream.fromFile(testFile)
-////            }
-////        } returns defaultPutObjectResponse
+//        coEvery {
```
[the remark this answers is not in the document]
LDsBF: I tried to fix such unit test last week but still had some error while mocking the inline request and response, I will discuss with Ian and try to configure this week, sorry :(

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
```diff
@@ -72,33 +78,74 @@ class S3TransferManagerIntegrationTest {
...
 
-        var operation = s3TransferManager.upload(testDirectory.toString(), toUri)
+        var operation = s3TransferManager.upload(testUploadDirectory.toString(), toUri)
         assertNotNull(operation, "The transfer manager didn't start directory upload")
         operation.await()
+        val listObjectsResponse = s3TransferManager.config.s3.listObjectsV2 {
+            bucket = testBucket
+            prefix = keyPrefix
+        }
+        assert(listObjectsResponse.contents!!.isNotEmpty())
```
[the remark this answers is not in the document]
ianbotsf: Agreed. I'll also mention that kotlin-test has more expressive assertions that we should use instead of the language-level `assert`. We should use something more expressive like:

```kotlin
val contents = assertNotNull(listObjectsResponse.contents) // This handles the `!!` part for you
assertFalse(contents.isEmpty) // This captures nicely the semantics we want to convey
```

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
LDsBF: If we use such `&&` here, we need to get `HeadObjectResponse` later by calling the `headObject()` again. I just think it's better to decrease reading from S3Client.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
LDsBF: I will change that, sometimes confused about how to choose `if` and `when` lol. Now I know to use `if ` for just single condition judge :)

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
lucix-aws: In both versions of the code, you'd only make the `headObject` call if the first part of the expression was `true` (`!from.key.endsWith('/')`) because of [short-circuiting](https://en.wikipedia.org/wiki/Short-circuit_evaluation). Most (if not all?) languages implement this kind of behavior, including Kotlin.

lucix-aws: Ok most wikipedia pages on programming/technical topics are pretty horrendous to read and that one is no exception, so to clarify:

For any given boolean expression, say `X && Y && Z`, languages will stop evaluating the expression at the first point that they can logically determine the result.

So in the above- if X was true, and Y was false, the truthiness of Z would not be evaluated, since the overall expression is only true if ALL of its operands are true. Obviously this is one example and this kind of shortcut can apply to infinitely large/complex expressions and with the or / `||` operator as well, I chose an example that matches closest to what you've done in your code.

So given that short-circuiting is a thing, the following two snippets are programmatically the exact same, in the sense that the function in the expression is only evaluated if `x > 5`.

```kotlin
if (x > 5 && extremelyExpensiveBooleanFunctionThatTakesFiveMinutesToCompute()) {
  // ...
}
```
```kotlin
if (x > 5) {
  if (extremelyExpensiveBooleanFunctionThatTakesFiveMinutesToCompute()) {
    // ...
  }
}
```

LDsBF: Yep I know that from Java, but I think we need `response.contentLength` in this block, so I just wonder how to get that `headObjectResponse` later without calling `headObject()` again. If I write something like `response = headObject()` within upper `if`, will Kotlin directly judge the `=` operation to be `true`?

lucix-aws: I think you're getting at a different concern here, which is how you can avoid calling `headObject` more than you need to. It's a valid concern. All I'm pointing out here is that in this particular change, moving the `headObject` call inside this one particular if block wasn't changing anything.

Given that you need the results of `headObject` elsewhere, you probably just want to call it and store the value further up in your method.

LDsBF: Make sense, I will just merge these two `if` and call `headObject` again if I need its return value.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
LDsBF: I did try to move it, but seems like `try` must be followed by at least one from `catch`/`finally` to deal with possible exception. I referred to this block from Java v1's S3 transfer manager, within which they wrote 
`finally {
            IOUtils.closeQuietly(objectContent, LOG);
            IOUtils.closeQuietly(randomAccessFile, LOG);
            IOUtils.closeQuietly(channel, LOG);
 }`
I'm not quite sure if we need to also close ``randomAccessFIle`/`channel` after downloading :(

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
LDsBF: Absolutely, since `S3Client` only has `getObject()` for single/multipart download instead of `uploadPart()` for upload operation. I will merge both cases to single `downloadFilePart`, thx!

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
```diff
@@ -38,29 +41,32 @@ class S3TransferManagerIntegrationTest {
...
             Thread {
                 file1.delete()
                 file2.delete()
                 file3.delete()
-                testDirectory1.toFile().delete()
-                testDirectory2.toFile().delete()
-                testDirectory.toFile().delete()
+                testUploadDirectory1.toFile().delete()
+                testUploadDirectory2.toFile().delete()
+                testUploadDirectory.toFile().delete()
```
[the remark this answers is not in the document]
LDsBF: I should consider such design earlier, my laziness... Many thanks!

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
```diff
@@ -38,67 +45,183 @@ class S3TransferManagerIntegrationTest {
...
+    @Test
+    fun testUploadLargeFile() = runTest {
         val testLargeFile = RandomTempFile(10000000)
-        operation = s3TransferManager.upload(testLargeFile.path, toUri)
+        val largeFileKey = "largefile"
+        val toUri = S3Uri(testBucket, largeFileKey)
+        val operation = s3TransferManager.upload(testLargeFile.path, toUri)
         assertNotNull(operation, "The transfer manager didn't start parts upload")
         operation.await()
+        assertTrue(checkUpload(testLargeFile, toUri))
```
lucix-aws, line 108: I was thinking about this some more, and since your upload/download code is explicitly reconstructing the file out of smaller parts (ie. its logic IS responsible for the final contents of the file in a non-trivial way), we should probably include verifying that the contents of the file are correct as well for the "large file" tests only.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -199,55 +205,95 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+                    if (objectContent != null) {
+                        position = downloadFilePart(objectContent, file, position)
+                    }
+                }
             }
         }
     }
 
+    private suspend fun downloadFilePart(objectContent: ByteStream, localFile: File, position: Long): Long {
+        val randomAccessFile = RandomAccessFile(localFile, "rw")
```
lucix-aws, line 265: Missed this on the first pass.

We shouldn't need this `RandomAccessFile` mechanism. You should just be able to get an `SdkByteWriteChannel` from the file object you create outside this method, and call multiple writes to it with each of your downloaded chunks. That would also allow you to get rid of having to pass and return the `position` value in order to continuously create and seek on the channel.

My guess is that making the switch to write channel will pretty much get rid of the need for this helper method as well:

```kotlin
// ... get SdkByteWriteChannel for target file
val objectContent = resp.body
if (objectContent != null) {
  // get SdkByteReadChannel from ByteStream, write it to write channel
}

File: hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt [hunk 11]
```diff
@@ -199,55 +205,95 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+            val file = File(to)
+            chunkRanges.forEach {
+                // contentRange format: "bytes=0-7999999"
+                val contentRange = "bytes=${it[0]}-${it[1]}"
+                val request = GetObjectRequest {
+                    bucket = from.bucket
+                    key = from.key
+                    range = contentRange
+                }
+                s3.getObject(request) { resp ->
```
lucix-aws, line 254: Missed this on the first pass.

I don't think we're getting the full benefit of doing chunked downloads unless we're doing them in parallel. As it's written AFAIK the only thing we're really avoiding is potentially timing out on big files.

I think doing that parallelization is almost certainly a `# FUTURE` though, I don't really want to blow up scope on you here since that change would be a bit more involved, and since I'm pretty sure we didn't do parallel chunked uploading.

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
```diff
@@ -38,67 +45,183 @@ class S3TransferManagerIntegrationTest {
...
+                    val subKey = Paths.get(to.key, it.name).toString()
+                    val subTo = S3Uri(to.bucket, subKey)
+                    if (!checkUpload(it, subTo)) {
+                        return false
+                    }
+                }
+                return true
+            }
+
+            else -> return true // there might be DS data that doesn't affect test
```
lucix-aws, line 138: Is this ever actually hit? Not sure I understand the comment.

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
```diff
@@ -38,67 +45,183 @@ class S3TransferManagerIntegrationTest {
...
         assertNotNull(operation, "The transfer manager didn't start parts upload")
         operation.await()
+        assertTrue(checkUpload(testLargeFile, toUri))
+        testLargeFile.delete()
+    }
+
+    private suspend fun checkUpload(localFile: File, to: S3Uri): Boolean {
+        when {
+            localFile.isFile -> {
+                return s3TransferManager.config.s3.headObjectOrNull(to) != null
```
lucix-aws, line 115: We should check that the names match as well.

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
```diff
@@ -38,67 +45,183 @@ class S3TransferManagerIntegrationTest {
...
+        }
+        var checkResult = true
+        response // Flow<ListObjectsV2Response>, a collection of pages
+            .transform { it.contents?.forEach { obj -> emit(obj) } }
+            .collect { obj ->
+                val key = obj.key!!
+                val keySuffix = key.substringAfter(keyPrefix)
+                val subFile = Paths.get(localFile.toString(), keySuffix).toFile()
+                if (!subFile.isFile()) {
+                    checkResult = false
```
lucix-aws, line 194: Couldn't we just `return false` out of the overall method right here?

lucix-aws (changes requested):

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
lucix-aws: Also, very minor, we don't really need the temporary `objectContent`. You can just reference `resp.body` directly.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
ianbotsf: The main reason to do chunked downloads is independent of parallelism—it's to do with providing timely status updates via the progress listeners. If we download a massive file in a single chunk, the listener would only get one status update. This way, we provide updates every _n_ bytes which allows for things like filling a progress bar, providing time estimates, etc.

Parallelism is a nice-to-have (assuming it actually improves performance, that may be a lot of file locking in the OS).

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
```diff
@@ -38,67 +45,183 @@ class S3TransferManagerIntegrationTest {
...
 
-    @AfterAll
+    @AfterEach
     private fun cleanup() = runBlocking {
+        deleteFiles(testUploadDirectory.toFile())
+        deleteFiles(testDownloadDirectory.toFile())
         S3TransferManagerTestUtils.deleteBucketAndAllContents(s3TransferManager.config.s3, testBucket)
     }
 
+    private fun deleteFiles(file: File) {
```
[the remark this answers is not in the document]
LDsBF: Perhaps I should consider whether it's better to append some helper function as class extension function in later design, good concept :)

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
```diff
@@ -38,67 +45,183 @@ class S3TransferManagerIntegrationTest {
...
-            Thread {
-                file1.delete()
-                file2.delete()
-                file3.delete()
-                testDirectory1.toFile().delete()
-                testDirectory2.toFile().delete()
-                testDirectory.toFile().delete()
-            }
-        )
+        testUploadDirectory = Files.createTempDirectory(dir, "testUploadDirectory")
```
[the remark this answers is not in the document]
LDsBF: I think I understand what you want, but in `testDownload()` case, `testUploadDirectory` is just used to create local temp files to upload for later downloading test. I will try to remove `@BeforeEach` / `@AfterEach` annotation and try to call them only when needed.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -179,13 +184,14 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
                 s3.headBucket {
                     bucket = from.bucket
                 }
-            } catch (e: Exception) {
+            } catch (e: S3Exception) {
                 throw IllegalArgumentException("The bucket does not exist or has no access to it")
             }
 
-            if (!from.key.endsWith('/') && objectExists(from)) {
+            if (!from.key.endsWith('/') && s3.headObjectOrNull(from) != null) {
```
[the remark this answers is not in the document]
LDsBF: Don't worry, I have reverted it to extended function's chain `if` code

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
LDsBF: So for large file, we should check local and S3 files' each chunk's ByteStream and validate if they are the same?

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
LDsBF: I used to discuss with Ian about that for `upload()`, some file is neither `isFile` nor `isDirectory`. Since we don't care about their upload, in this check function, I just set it to return true.

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
LDsBF: In `upload()` and `download()` API, file/object is uploaded/downloaded with the same name mutually (under given local directory name and key prefix directory). Therefore, at the end of `testUpload()`, when we call `checkUpload(testUploadDirectory.toFile(), toUri)`, `toUri` just refers to our upload target key prefix directory when we call `s3TransferManager.upload(testUploadDirectory.toString(), toUri)`, so all `subTo` 's `subKey` is concatenated from that key prefix and local file/directory name, and the result of heading them automatically validate the uploaded objects' name matches local file name, I think :)

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
LDsBF: I tried but compile error shows `'return' is not allowed here`, seems like collect doesn't allow immediate return like java :(

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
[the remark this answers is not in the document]
LDsBF: For download tests, `testUploadDirectory` should exist so that S3 objects can be uploaded first and then to be downloaded. I'm not sure whether we should prepare a `testUploadDirectory` / `testDownloadDirectory` for each test case separately to decrease our code amount or specify each case's test directory creation inside their own block but with more code.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -199,55 +205,95 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
     }
 
+    private suspend fun downloadFilePart(objectContent: ByteStream, localFile: File, position: Long): Long {
+        val randomAccessFile = RandomAccessFile(localFile, "rw")
+        val channel = randomAccessFile.channel
+        channel.position(position)
+        val filePosition: Long
+
+        try {
+            val byteArray = objectContent.toByteArray()
```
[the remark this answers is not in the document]
LDsBF: But when I checked `SdkByteWriteChannel` I found both `writeFully` and `writeAvailable` accept `src: ByteArray` as parameter so I'm confused whether to convert `ByteStream` to `ByteArray`. Additionally, for `SdkByteReadChannel`, it can be obtained from `ByteStram` by implementing `OneShotStream` or `ReplayableStream`, but for their example code, I just see one `FileContent` class but it just constructs using `File` insteand of `ByteStream`. So is there any more explaination about that guess :)

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
lucix-aws: Oh you're not wrong. `headObject`'s key will have the filename. I'm silly.

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
lucix-aws: That's right.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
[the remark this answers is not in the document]
lucix-aws: Yeah I see the confusion on `writeFully`/`writeAvailable`. The method to write a read channel to a write channel is actually exposed as `SdkByteReadChannel.copyTo`. FWIW that seems like an API miss for us and we should probably add an extension on write channel that takes a read channel.

`ByteStream` like you said will have an underlying read channel when it's either implemented by `OneShotStream` or `ReplayableStream`. You'd have to write a utility to pull out the channel or throw if it's implemented by `Buffer`. Again this seems like something we should really be directly supporting such that customers can get at these channels and use them however they need to.

I'm working on a snippet that I'll post in a second here.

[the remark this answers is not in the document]
lucix-aws: Here's a sample of one write channel for writing two parts of an 8-byte file.

```kotlin
fun ByteStream.toReadChannel(): SdkByteReadChannel =
    when (this) {
        is ByteStream.OneShotStream -> readFrom()
        is ByteStream.ReplayableStream -> newReader()
        is ByteStream.Buffer -> TODO()
    }

// pretty sure the hll project is set up to not require this
@OptIn(InternalApi::class)
suspend fun writeABunchOfStreams() {
    val f = File("./tmp")
    val writeChannel = f.writeChannel()

    S3Client.fromEnvironment().apply {
        getObject(GetObjectRequest {
            bucket = "buckettestdgttxcq"
            key = "test.txt"
            range = "bytes=0-3"
        }) {
            it.body?.toReadChannel()?.copyTo(writeChannel, close = false)
        }
        getObject(GetObjectRequest {
            bucket = "buckettestdgttxcq"
            key = "test.txt"
            range = "bytes=4-7"
        }) {
            it.body?.toReadChannel()?.copyTo(writeChannel, close = false)
        }
    }

    // you'll need to ensure the write channel is closed, either on the final write call or manually after all of them
}
```

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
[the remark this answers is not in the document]
lucix-aws: Yeah, ideally the "precondition" for the download test is that the test data to be downloaded exists on S3 and your local file environment is blank.

The reverse goes for upload - the upload directory should be there, and the bucket should be blank.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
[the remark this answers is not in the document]
LDsBF: Seems like `@OptIn(InternalApi::class)` can not be removed for `hlls`, but other design just works. Thanks for instruction :)

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -199,55 +211,80 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+            chunkRanges.forEach {
+                // contentRange format: "bytes=0-7999999"
+                val contentRange = "bytes=${it[0]}-${it[1]}"
+                val request = GetObjectRequest {
+                    bucket = from.bucket
+                    key = from.key
+                    range = contentRange
+                }
+                s3.getObject(request) { resp ->
+                    resp.body?.toReadChannel()?.copyTo(writeChannel, close = false)
```
lucix-aws, line 261: love it

lucix-aws (approved):

ianbotsf (approved):
I'm approving this PR for now with the comments below. Please follow up on these cleanups in your next PR!

PR #664, opened 2022-07-29
Feat s3 transfermanager

wty-Bryant:
## Description of changes

### DIFF
--- a/hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
+++ b/hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
@@ -407,7 +407,7 @@ public fun partition(fileSize: Long, chunkSize: Long): List<LongRange> =
     }
 
 @InternalSdkApi
-public fun String.ensureEndsWith(c: Char) = if (endsWith(c)) this else plus(c)
+public fun String.ensureEndsWith(c: Char): String = if (endsWith(c)) this else plus(c)
 
 private fun ByteStream.toReadChannel(): SdkByteReadChannel = when (this) {
     is ByteStream.OneShotStream -> readFrom()


hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
```diff
@@ -188,19 +209,22 @@ class S3TransferManagerIntegrationTest {
...
-        if (!from.key.endsWith('/') && headObjectResponse != null) {
-            if (headObjectResponse.contentLength > s3TransferManager.config.chunkSize) {
-                return chunksCompare(localFile, from)
+        if (!from.key.endsWith('/')) {
+            val headObjectResponse = s3TransferManager.config.s3.headObjectOrNull(from)
+            if (headObjectResponse != null) {
+                if (headObjectResponse.contentLength > s3TransferManager.config.chunkSize) {
+                    return chunksCompare(localFile, from)
+                }
+                return localFile.isFile()
```
ianbotsf, line 226: **Question**: Why do we do a `chunksCompare` if the file is big but only check `isFile()` if the file is small? Shouldn't we be checking the content of small files too? (And can we reuse `chunksCompare` for that as-is?)

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -265,8 +264,117 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+            }
+        }
+    }
+
+    context(CoroutineScope)
+    private fun copyObjectParts(fileSize: Long, from: S3Uri, to: S3Uri) {
+        async {
+            val chunkRanges = (0 until fileSize step config.chunkSize).map {
+                arrayOf(it, minOf(it + config.chunkSize - 1, fileSize - 1))
+            }
```
ianbotsf, line 343: **Question**: Why are we using arrays of start/end indices rather than [ranges](https://kotlinlang.org/docs/ranges.html)?

Ranges allow a convenient syntax (`until`) that doesn't require using `- 1` to skip the last index:

```kotlin
it until minOf(it + config.chunkSize, fileSize)
```

And ranges have properties (`start` and `endInclusive`) that are more meaningful than array indices:

```kotlin
val contentRange = "bytes=${chunkRange.start}-${chunkRange.endInclusive}"
```

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -283,6 +391,17 @@ public suspend fun S3Client.headObjectOrNull(s3Uri: S3Uri): HeadObjectResponse?
...
+    return try { // first check if bucket exists
+        headBucket {
+            bucket = bucketName
+        }
+        true
+    } catch (e: S3Exception) {
+        false
+    }
+}
+
```
ianbotsf, line 404: **Question**: Why does this need to be `public`? It seems like it should be an implementation detail (and thus `private`).

ianbotsf (changes requested):

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -265,8 +264,117 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+                key = to.key
+            }
+        }
+    }
+
+    context(CoroutineScope)
+    private fun copyObjectParts(fileSize: Long, from: S3Uri, to: S3Uri) {
+        async {
+            val chunkRanges = (0 until fileSize step config.chunkSize).map {
+                arrayOf(it, minOf(it + config.chunkSize - 1, fileSize - 1))
```
lucix-aws, line 342: I should have called this out earlier, but we have a more idiomatic way of representing "ranges" in kotlin by virtue of [`ClosedRange`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-closed-range/) - specifically [`LongRange`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-long-range/) for this use case.

You can construct them manually, eg.
```kotlin
val chunkRanges = (0 until fileSize step config.chunkSize).map {
    LongRange(it, minOf(it + config.chunkSize - 1, fileSize - 1))
}
```

but there's also syntax support for constructing them literally eg.
```kotlin
val chunkRanges = (0 until fileSize step config.chunkSize).map {
    it..minOf(it + config.chunkSize - 1, fileSize - 1)
}
```

It can be used in code like
```kotlin
val uploadPartCopyResponse = s3.uploadPartCopy {
    copySourceRange = "bytes=${chunkRange.start}-${chunkRange.endInclusive}"
    // ...
}
```

FWIW you might notice that you've technically been using literal ranges in your code already -
```kotlin
(0 until fileSize step config.chunkSize)
```
is literal syntax for constructing an open-end range with a step.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -265,8 +264,117 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+            s3.copyObject {
+                copySource = "${from.bucket}/${from.key}"
+                bucket = to.bucket
+                key = to.key
+            }
+        }
+    }
+
+    context(CoroutineScope)
+    private fun copyObjectParts(fileSize: Long, from: S3Uri, to: S3Uri) {
```
lucix-aws, line 339: This method in general is pretty much identical to your `uploadObject` equivalents - we may be able to share logic if we have one method that takes an optional `copySource` (that seems to be the main difference).

lucix-aws (changes requested):

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
lucix-aws: Commented the same below 😂

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
lucix-aws: EDIT: Ian's example used `until` in constructing the ranges such that you could avoid the `x - 1`. That's way better than what I have here and you should do that instead.

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
LDsBF: Luc put forward the large file chunk check in last PR. I guess that's because there are more small files than large files. perhaps I can also check small file's chunk.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
LDsBF: Just thought `copySourceRange` only requres each chunk's start and end index. I have modified that.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -265,8 +264,117 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+
+            if (!from.key.endsWith('/')) {
+                val response = s3.headObjectOrNull(from)
+                if (response != null) {
+                    copyObject(response.contentLength, from, to)
+                    return@async
+                }
+            }
+
+            val keyPrefix = if (from.key.endsWith('/')) from.key else from.key.plus('/')
```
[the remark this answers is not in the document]
LDsBF: I have never imagined that we could even add extension to basic data structure lol, I will add that

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
```diff
@@ -48,12 +48,14 @@ class S3TransferManagerIntegrationTest {
...
 
+    private lateinit var testBucket1: String
+
     private lateinit var testUploadDirectory: Path
 
     private lateinit var testDownloadDirectory: Path
 
     @BeforeEach
-    private fun createResources(): Unit = runBlocking {
+    private fun createResources() = runBlocking {
```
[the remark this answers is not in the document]
LDsBF: `: Unit` means this method returns void, so I think it might be cleaner to remove that

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -265,8 +264,117 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+        }
+
+        return DefaultOperation(deferred)
+    }
+
+    context(CoroutineScope)
+    private fun copyObject(fileSize: Long, from: S3Uri, to: S3Uri) {
+        // if file size <= chunkSize, call copyWholeObject
+        // otherwise, call copyObjectParts
+        if (fileSize <= config.chunkSize) {
```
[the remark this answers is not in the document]
LDsBF: Seems not, I check `S3Client` API and found that copy operation is more similar to upload operation while dealing with different size's files. For large file, a `createMultipartUpload()` and a `completeMultipartUpload()` are also needed. Design can be found in `S3Client` code as well as java version's doc https://docs.aws.amazon.com/AmazonS3/latest/userguide/CopyingObjectsMPUapi.html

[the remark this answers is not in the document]
LDsBF: That might mean for small files, we can just call `copyObject` without going through such 3 steps.

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -265,8 +264,117 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+            }
+
+            val createMultipartCopyResponse = s3.createMultipartUpload {
+                bucket = to.bucket
+                key = to.key
+            }
+            val completedParts = mutableListOf<CompletedPart>()
+
+            chunkRanges.forEachIndexed { index, chunkRange ->
+                // contentRange format: "bytes=0-7999999"
```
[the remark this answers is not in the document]
LDsBF: Ok I will just delete this comment

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
LDsBF: But I'm afraid that will make merged API not so clear and longer, since `uploadPart` and `uploadPartCopy` are different API in`S3Client` and takes different parameter like `body/copySource` and `copySourceRange` (equivalent parameter of which is not defined in `uploadPart`). I prefer to separate them now. What's your opinion towards that reason :) ?

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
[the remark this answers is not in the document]
ianbotsf: Return types are tricky in Kotlin. Since you're using `=` syntax for the method body, the compiler will infer the return type for you. `runBlocking` is generic and returns the result of its inner block. The last statement in the inner block is `Files.createTempDirectory` which returns a `Path`. Thus, this whole `createResources` method now has a return type of `Path`. (You should be able to confirm this by showing the method signature in your IDE.)

This method probably doesn't want to return `Path`—`Unit` seems more natural. So I recommend you keep the `: Unit` explicit return type.

(Note that this return type resolution _does not_ happen when using the `{ }` syntax for method bodies. Then, the return type is `Unit` unless explicitly specified.)

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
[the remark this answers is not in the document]
ianbotsf: +1. Avoiding `createMultipartUpload` and `completeMultipartUpload` calls are reason enough to warrant the special case.

hlls/s3-transfermanager/e2eTest/S3TransferManagerIntegrationTest.kt
lucix-aws: My argument was that for small files, we're just doing a straight copy, so the onus of getting that content correct is really on the service.

For multiparts, we're reconstructing the file ourselves, so we're then responsible for its correctness and therefore we should test for it (IMO).

hlls/s3-transfermanager/jvm/src/aws/sdk/kotlin/s3/transfermanager/DefaultS3TransferManager.kt
```diff
@@ -199,55 +207,202 @@ internal class DefaultS3TransferManager(override val config: S3TransferManager.C
...
+    } catch (e: S3Exception) {
+        null
+    }
+
+public fun partition(fileSize: Long, chunkSize: Long): List<LongRange> =
+    (0 until fileSize step chunkSize).map {
+        it until minOf(it + chunkSize, fileSize)
+    }
+
+public fun String.makeEndWithSingleSlash(): String {
```
lucix-aws, line 400: We swung a bit too hard on this one - the character we "ensure" the string ends with should be a parameter to this extension.

Also could improve the naming. Thinking something like `ensureEndsWith`?

lucix-aws (approved):

ianbotsf (approved):

PR #699, opened 2022-09-19
chore: upgrade smithy-kotlin version

ianbotsf:
## Issue \#

(none)

## Description of changes

Update **smithy-kotlin** version to just-released **0.12.6**.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -12,7 +12,7 @@ sdkVersion=0.17.6-SNAPSHOT
 smithyVersion=1.23.0
 smithyGradleVersion=0.6.0
 # smithy-kotlin codegen and runtime are versioned together
-smithyKotlinVersion=0.12.6-SNAPSHOT
+smithyKotlinVersion=0.12.6
 
 # kotlin
 kotlinVersion=1.7.10


gradle.properties
```diff
@@ -39,4 +39,4 @@ slf4jVersion=1.7.36
 
 # dokka config (values specified at build-time as needed)
 smithyKotlinPackageListUrl=
-smithyKotlinDocBaseUrl=https://docs.aws.amazon.com/sdk-for-kotlin/latest/reference/smithy-kotlin
\ No newline at end of file
+smithyKotlinDocBaseUrl=https://docs.aws.amazon.com/sdk-for-kotlin/latest/reference/smithy-kotlin
```
aajtodd, line 42: FYI this will need to change but it looks like the [logic consuming ](https://github.com/awslabs/aws-sdk-kotlin/blob/230751e55e50afe97c263bcf2698202b959a18b0/build.gradle.kts#L60) this might be off anyway (would require setting the package list and base URL if I'm not mistaken). We can fix this in next release it will just mean the docs won't link correctly to smithy-kotlin.

aajtodd (approved):

Issue #701, opened 2022-09-21
implement adaptive retry strategy


lucix-aws:
### Describe the feature

Implement the "adaptive" retry mode (`RetryStrategy`) that has been standardized across AWS SDKs.

Ensure env and profile values propagate through for this setting.

### Is your Feature Request related to a problem?

n/a

### Acknowledge

- [ ] I may be able to implement this feature request

### AWS Kotlin SDK version used

n/a

### Platform (JVM/JS/Native)

n/a

### Operating System and version

n/a

Issue #486, opened 2022-01-03
Support retry config


aajtodd:
### Describe the feature

Add support for retry configuration from environment and profiles

Environment variables:
* `AWS_MAX_ATTEMPTS`
* `AWS_RETRY_MODE`

Profile keys:
* `max_attempts`
* `retry_mode`

These need added to the shared configuration `AwsClientConfig` and threaded through the runtime and generated clients.

### Is your Feature Request related to a problem?

Feature parity

### Acknowledge

- [ ] I may be able to implement this feature request

### AWS Kotlin SDK version used

0.9.5-beta

### Platform (JVM/JS/Native)

N/A

### Operating System and version

N/A

lauzadis:
Closed by #702

PR #702, opened 2022-09-21
feat: enable retry configuration

lauzadis:
Enables configuration of the retry strategy (specifically, max attempts and retry mode) through environment variables, system properties, and AWS profile configuration.

## Issue \#

#486 

## Description of changes

This change is required to achieve feature parity with existing SDKs. It solves the problem of lacking user configuration for max attempts and retry mode. 

Currently, we only support `legacy` and `standard` retry modes (with `legacy` behaving equivalently to `standard`). Support for `adaptive` retry mode will be added in #701.


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
```diff
@@ -0,0 +1,49 @@
...
+    val profile = asyncLazy { loadActiveAwsProfile(platformProvider) }
+
+    // resolve max attempts
+    val maxAttempts: Int = (
+        AwsSdkSetting.AwsMaxAttempts.resolve(platformProvider)
+            ?: profile.get().maxRetryAttempts
+            ?: (StandardRetryStrategyOptions.Default.maxAttempts).toString()
+        ).toInt()
+
+    if (maxAttempts < 1) {
```
lucix-aws, line 36: Prefer `require(maxAttempts < 1)` { "max attempts was $maxAttempts but should be at least 1" }

Note that the type of exception thrown by `require` is an `IllegalArgumentException`.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Attempt to resolve the retry strategy used to make requests by fetching the max attempts and retry mode. Currently,
+ * we only support the legacy and standard retry modes.
+ */
+@InternalSdkApi
+public suspend fun resolveRetryStrategy(platformProvider: PlatformProvider = Platform): RetryStrategy {
+    val profile = asyncLazy { loadActiveAwsProfile(platformProvider) }
+
+    // resolve max attempts
+    val maxAttempts: Int = (
```
lucix-aws, line 30: You should be able to omit the type declaration here.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
```diff
@@ -0,0 +1,49 @@
...
+        AwsSdkSetting.AwsMaxAttempts.resolve(platformProvider)
+            ?: profile.get().maxRetryAttempts
+            ?: (StandardRetryStrategyOptions.Default.maxAttempts).toString()
+        ).toInt()
```
lucix-aws, line 34: **Question**: Do we want to handle a type conversion failure here? Or just let it bubble?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    if (maxAttempts < 1) {
+        throw IllegalArgumentException("max attempts was $maxAttempts but should be at least 1")
+    }
+
+    // resolve retry mode
+    val retryMode = AwsSdkSetting.AwsRetryMode.resolve(platformProvider) ?: profile.get().retryMode ?: "standard"
+    if (retryMode == "adaptive") {
```
lucix-aws, line 42: We can make this more explicit with a `when` construct e.g.

(If you haven't come across it yet, `when` is kotlin's version of `switch`, and can be used as an expression as well as a freestanding statement like in other languages)

```kotlin
return when (retryMode) {
    "standard", "legacy" -> StandardRetryStrategy(/* ... */)
    "adaptive" -> throw NotImplementedError(/* ... */)
    else -> throw IllegalArgumentException(/* ... */)
}
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
```diff
@@ -0,0 +1,49 @@
...
+ * Attempt to resolve the retry strategy used to make requests by fetching the max attempts and retry mode. Currently,
+ * we only support the legacy and standard retry modes.
+ */
+@InternalSdkApi
+public suspend fun resolveRetryStrategy(platformProvider: PlatformProvider = Platform): RetryStrategy {
+    val profile = asyncLazy { loadActiveAwsProfile(platformProvider) }
+
+    // resolve max attempts
+    val maxAttempts: Int = (
+        AwsSdkSetting.AwsMaxAttempts.resolve(platformProvider)
```
lucix-aws, line 31: This expression is awkward. You should be able to simplify it by invoking `toInt()` directly on the retrieved `maxRetryAttempts` from profile, instead of string-ifying and then re-intifying the whole expression.

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategyTest.kt
```diff
@@ -0,0 +1,216 @@
...
+
+    @Test
+    fun itThrowsOnUnsupportedRetryModes() = runTest {
+        val retryMode = "unsupported-retry-mode"
+
+        val platform = TestPlatformProvider(
+            env = mapOf(AwsSdkSetting.AwsRetryMode.environmentVariable to retryMode),
+        )
+
+        assertThrows(UnsupportedOperationException::class.java) {
```
lucix-aws, line 84: You can use `assertFailsWith<T: Exception>(...)` here (and in your other failure checks) rather than having to get at the class reference.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
```diff
@@ -0,0 +1,49 @@
...
+        throw IllegalArgumentException("max attempts was $maxAttempts but should be at least 1")
+    }
+
+    // resolve retry mode
+    val retryMode = AwsSdkSetting.AwsRetryMode.resolve(platformProvider) ?: profile.get().retryMode ?: "standard"
+    if (retryMode == "adaptive") {
+        throw NotImplementedError("Retry mode $retryMode is not implemented yet. https://github.com/awslabs/aws-sdk-kotlin/issues/701")
+    } else if (retryMode != "legacy" && retryMode != "standard") {
+        throw UnsupportedOperationException("Retry mode $retryMode is not supported, should be one of: \"legacy\", \"standard\", \"adaptive\"")
+    }
```
ianbotsf, line 46: **Nit**: Support matching of non-lowercase values like `Standard` or `STANDARD`.

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategyTest.kt
```diff
@@ -0,0 +1,216 @@
...
+import aws.smithy.kotlin.runtime.retries.StandardRetryStrategyOptions
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Assertions.assertThrows
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class ResolveRetryStrategyTest {
```
lucix-aws, line 19: This is a comprehensive test suite, well done!

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
ianbotsf, line 46: **Correctness**: We don't support a "legacy" retry mode and we never will. If a user tries to configure a retry mode of "legacy" they should get an error message.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
```diff
@@ -53,3 +53,16 @@ public val AwsProfile.roleArn: String?
...
+ * aws.maxAttempts = 1 will have 0 retries.
+ */
+public val AwsProfile.maxRetryAttempts: String?
+    get() = this["max_attempts"]
+
+/**
+ * Which RetryMode to use for the default RetryPolicy, when one is not specified at the client level.
+ */
+public val AwsProfile.retryMode: String?
+    get() = this["retry_mode"]
```
ianbotsf, line 68: **Comment**: It feels odd to me to treat these values as strings when we know they shouldn't be. I see we haven't defined any non-string values from profile yet but I believe these could be the first.

I'd suggest playing with parsing a non-null string into an `Int`/`RetryMode` at this level, including throwing an error when the values are nonsensical.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
```diff
@@ -0,0 +1,49 @@
...
+    if (maxAttempts < 1) {
+        throw IllegalArgumentException("max attempts was $maxAttempts but should be at least 1")
+    }
+
+    // resolve retry mode
+    val retryMode = AwsSdkSetting.AwsRetryMode.resolve(platformProvider) ?: profile.get().retryMode ?: "standard"
+    if (retryMode == "adaptive") {
+        throw NotImplementedError("Retry mode $retryMode is not implemented yet. https://github.com/awslabs/aws-sdk-kotlin/issues/701")
+    } else if (retryMode != "legacy" && retryMode != "standard") {
+        throw UnsupportedOperationException("Retry mode $retryMode is not supported, should be one of: \"legacy\", \"standard\", \"adaptive\"")
```
lucix-aws, line 45: Forgot this on the first pass, but let's use `IllegalArgumentException` here.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
```diff
@@ -53,3 +53,16 @@ public val AwsProfile.roleArn: String?
  */
 public val AwsProfile.sourceProfile: String?
     get() = this["source_profile"]
+
+/**
+ * The maximum number of request attempts to perform. This is one more than the number of retries, so
+ * aws.maxAttempts = 1 will have 0 retries.
+ */
+public val AwsProfile.maxRetryAttempts: String?
```
lucix-aws, line 61: Consider having this extension handle the type cast. This is the first time we've referenced a non-string field here, AFAIK, so I'm curious what the others think.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AwsSdkSetting.kt
```diff
@@ -130,6 +130,17 @@ public sealed class AwsSdkSetting<T>(
...
+    /**
+     * The maximum number of request attempts to perform. This is one more than the number of retries, so
+     * aws.maxAttempts = 1 will have 0 retries.
+     */
+    public object AwsMaxAttempts : AwsSdkSetting<String>("AWS_MAX_ATTEMPTS", "aws.maxAttempts")
+
+    /**
+     * Which RetryMode to use for the default RetryPolicy, when one is not specified at the client level.
+     */
+    public object AwsRetryMode : AwsSdkSetting<String>("AWS_RETRY_MODE", "aws.retryMode")
```
ianbotsf, line 143: **Comment**: Same as in **AwsProfile.kt**—let's treat these as `Int`/`RetryMode` settings rather than strings.

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategyTest.kt
```diff
@@ -0,0 +1,216 @@
...
+    @Test
+    fun itThrowsOnInvalidMaxAttemptsValues() = runTest {
+        val invalidMaxAttemptsValues = listOf(-91, -5, 0)
+
+        for (invalidMaxAttempts in invalidMaxAttemptsValues) {
+            val platform = TestPlatformProvider(
+                env = mapOf(AwsSdkSetting.AwsMaxAttempts.environmentVariable to invalidMaxAttempts.toString()),
+            )
+
+            assertThrows(IllegalArgumentException::class.java) {
```
ianbotsf, line 68: **Style**: There's a slightly nicer syntax for this directly in **kotlin-test**:

```kotlin
assertThrows<IllegalArgumentException> { ... }
```

ianbotsf (changes requested):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
aajtodd: Well what if they share profiles between SDKs or switch from one SDK to another, etc? It seems reasonable to me to interpret `legacy` as `standard` for our SDK but I'm not opposed to throwing an error either 🤷

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategyTest.kt
```diff
@@ -0,0 +1,216 @@
...
+    @Test
+    fun itThrowsOnUnsupportedRetryModes() = runTest {
+        val retryMode = "unsupported-retry-mode"
+
+        val platform = TestPlatformProvider(
+            env = mapOf(AwsSdkSetting.AwsRetryMode.environmentVariable to retryMode),
+        )
+
+        assertThrows(UnsupportedOperationException::class.java) {
+            runBlocking {
```
aajtodd, line 85: I'm not sure this is necessary `runTest` provides a scope already and `assertThrows` should be inline IIRC. 

`runBlocking` shouldn't even work in `common`, it's not available on every platform, not sure why it works here

aajtodd (commented):
Nice work

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
lauzadis: +1 to what Aaron said, I thought we agreed this morning to treat `legacy` exactly like `standard`, but I'm also open to just throwing an error

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategyTest.kt
lauzadis: I was having trouble because `resolveRetryStrategy` is a suspending function, and I followed the solution here: https://github.com/junit-team/junit5/issues/1851

I'll see if the recommended `assertFailsWith<T: Exception>(...)` fixes this

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
ianbotsf: Ooh, that's an interesting possibility I did not consider. OK I can live with that, especially given the verbiage in the [generic SDK docs on retry behavior](https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html). I withdraw my concern here.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
lauzadis: I will handle type conversion failures at the AwsProfile / AwsSdkSetting level (throw an error when reading the values, instead of when trying to use them)

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
```diff
@@ -53,3 +55,20 @@ public val AwsProfile.roleArn: String?
...
 public val AwsProfile.sourceProfile: String?
     get() = this["source_profile"]
+
+/**
+ * The maximum number of request attempts to perform. This is one more than the number of retries, so
+ * aws.maxAttempts = 1 will have 0 retries.
+ */
+public val AwsProfile.maxAttempts: Int?
+    get() = try { this["max_attempts"]?.toInt() } catch (e: NumberFormatException) {
+        throw IllegalArgumentException("Failed to parse maxAttempts " + this["max_attempts"] + " as an integer", e)
```
lucix-aws, line 65: I see you used the string interpolation feature elsewhere, you can do it with more complex statements if you wrap the expression in `{}`s:

```kotlin
throw IllegalArgumentException("Failed to parse maxAttempts ${this["max_attempts"]} as an int", e)
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Attempt to resolve the retry strategy used to make requests by fetching the max attempts and retry mode. Currently,
+ * we only support the legacy and standard retry modes.
+ */
+@InternalSdkApi
+public suspend fun resolveRetryStrategy(platformProvider: PlatformProvider = Platform): RetryStrategy {
+    val profile = asyncLazy { loadActiveAwsProfile(platformProvider) }
+
+    // resolve max attempts
```
lucix-aws, line 29: **Nit**: I'm generally not a fan of shortform-type comments like this where it describes the behavior that's otherwise evident based on the code itself.

lucix-aws (approved):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
aajtodd: We might consider `ConfigurationException` here

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AwsSdkSetting.kt
```diff
@@ -159,6 +160,11 @@ public inline fun <reified T> AwsSdkSetting<T>.resolve(platform: PlatformEnviron
             Int::class -> strValue.toInt()
             Long::class -> strValue.toLong()
             Boolean::class -> strValue.toBoolean()
+            RetryMode::class -> {
+                try { RetryMode.valueOf(strValue.uppercase()) } catch (e: IllegalArgumentException) {
+                    throw IllegalArgumentException("Retry mode $strValue is not supported, should be one of: \"legacy\", \"standard\", \"adaptive\"", e)
+                }
```
ianbotsf, line 166: **Style**: This works but relies on uppercasing, exception catching, and duplicating the values from the enum in a string. Consider:

```kotlin
RetryMode
    .values()
    .firstOrNull { it.name.equals(strValue, ignoreCase = true) }
    ?: throw IllegalArgumentException("Retry mode $strValue is not supported, should be one of: ${RetryMode.values().joinToString(", ")}"
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/RetryMode.kt
```diff
@@ -0,0 +1,7 @@
+package aws.sdk.kotlin.runtime.config.retries
+
+public enum class RetryMode {
```
aajtodd, line 3: fix: missing documentation

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AwsSdkSetting.kt
```diff
@@ -159,6 +160,11 @@ public inline fun <reified T> AwsSdkSetting<T>.resolve(platform: PlatformEnviron
...
+            }
```
ianbotsf, line 167: **Comment**: I assumed there'd be an easy generic way to parse _any_ enum class but after trying several different implementations it seems it's not straightforward (even with reified generics). Alas, it seems we're stuck with separate branches for any future enum class.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
```diff
@@ -58,11 +60,15 @@ public val AwsProfile.sourceProfile: String?
  * The maximum number of request attempts to perform. This is one more than the number of retries, so
  * aws.maxAttempts = 1 will have 0 retries.
  */
-public val AwsProfile.maxRetryAttempts: String?
-    get() = this["max_attempts"]
+public val AwsProfile.maxAttempts: Int?
+    get() = try { this["max_attempts"]?.toInt() } catch (e: NumberFormatException) {
+        throw IllegalArgumentException("Failed to parse maxAttempts " + this["max_attempts"] + " as an integer", e)
```
ianbotsf, line 65: **Style**: Consider using null coalescing instead of exception catching:

```kotlin
get() = this["max_attempts"]?.run { toIntOrNull() ?: throw IllegalArgumentException(...) }
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/RetryMode.kt
```diff
@@ -0,0 +1,7 @@
...
+    LEGACY,
+    STANDARD,
+    ADAPTIVE;
+}
```
ianbotsf, line 7: **Nit**: Please add KDocs.

ianbotsf (approved):
Only minor issues left. Fix at least the **Nit** (consider the **Style**s optional) and ship.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AwsSdkSetting.kt
```diff
@@ -148,6 +160,11 @@ public inline fun <reified T> AwsSdkSetting<T>.resolve(platform: PlatformEnviron
             Int::class -> strValue.toInt()
             Long::class -> strValue.toLong()
             Boolean::class -> strValue.toBoolean()
+            RetryMode::class -> {
```
aajtodd, line 163: Currently all the other AwsSdkSettings are simple types and match what they are allowed to be in config/environment and later converted as necessary. I suppose this is fine though.

aajtodd (approved):
Fix and ship!

PR #708, opened 2022-09-27
Initial tracing framework via ExecutionContext

ianbotsf:
## Issue \#

[smithy-kotlin#677](https://github.com/awslabs/smithy-kotlin/issues/677)

## Description of changes

Incorporate necessary changes for implementing tracing via `ExecutionContext`.

**Companion PR**: [smithy-kotlin#707](https://github.com/awslabs/smithy-kotlin/pull/707)

### DIFF
--- a/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/imds/ImdsClient.kt
+++ b/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/imds/ImdsClient.kt
@@ -31,6 +31,7 @@ import aws.smithy.kotlin.runtime.retries.delay.StandardRetryTokenBucketOptions
 import aws.smithy.kotlin.runtime.time.Clock
 import aws.smithy.kotlin.runtime.tracing.NoOpTraceSpan
 import aws.smithy.kotlin.runtime.tracing.TraceSpan
+import aws.smithy.kotlin.runtime.tracing.withRootSpan
 import aws.smithy.kotlin.runtime.util.Platform
 import aws.smithy.kotlin.runtime.util.PlatformProvider
 import kotlin.time.Duration
@@ -136,7 +137,6 @@ public class ImdsClient private constructor(builder: Builder) : InstanceMetadata
             context {
                 operationName = path
                 service = SERVICE
-                this.traceSpan = traceSpan
 
                 // artifact of re-using ServiceEndpointResolver middleware
                 set(SdkClientOption.LogMode, sdkLogMode)
@@ -151,7 +151,9 @@ public class ImdsClient private constructor(builder: Builder) : InstanceMetadata
             next.call(req)
         }
 
-        return op.roundTrip(httpClient, Unit)
+        return op.context.withRootSpan(traceSpan) {
+            op.roundTrip(httpClient, Unit)
+        }
     }
 
     override fun close() {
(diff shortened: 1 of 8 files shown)


aajtodd (commented):
Overall looks good, see other PR feedback

aajtodd (approved):
Couple small suggestions but nothing blocking.

I think threading it through as far as you did is very illustrative so thanks for taking the time to do so. I still am not sure I love that we _have_ to do that but it may be the right thing ultimately. I'd like to maybe hear from the rest of the team. It may just be a shift in mindset that we will expect to  see optional `traceSpan` parameters throughout the runtime.

Taking a partial log output from the `smithy-kotlin` PR I can see what you mean about the nesting:

```
Attempt-1/Signing/DefaultChain/Cached/Chain/Profile @ aws.sdk.kotlin.runtime.auth.credentials.ProfileCredentialsProvider
```

I'm wondering since the event/logging has the source component if this is adding anything or if it should be more like `Signing/ResolveCredentials/ProfileProvider`? I don't think there is anything wrong with what you have but we need to probably document (perhaps in our contributing guide) guidelines for when it is appropriate to add new child spans, how to name them, etc because it's not clear to me right now (you had a good verbal summary that may have answered this that I have since forgotten).

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/imds/ImdsClient.kt
```diff
@@ -48,8 +50,11 @@ private const val SERVICE = "imds"
 public interface InstanceMetadataProvider : Closeable {
     /**
      * Gets the specified instance metadata value by the given path.
+     * @param path The URI path to fetch
+     * @param traceSpan The span to which tracing events should be omitted. Defaults to [NoOpTraceSpan] if none is
```
lauzadis, line 54: nit(typo): should it be "the span to which tracing events should be **emitted**"?

lauzadis (approved):
Looks good! I agree with Aaron that threading these traceSpans through the existing code will require a change in mindset with us expecting these parameters to appear all over the codebase. I don't know if it's very developer-friendly but that might change once it's threaded through everything.

For naming the child spans, I think it would be good to have a clear naming scheme. It would be helpful to have more clarity (i.e the child span named "Cached" should probably be called the full "CachedCredentialsProvider") so that users can search their logs easier.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/imds/ImdsClient.kt
ianbotsf: > For naming the child spans, I think it would be good to have a clear naming scheme. It would be helpful to have more clarity (i.e the child span named "Cached" should probably be called the full "CachedCredentialsProvider") so that users can search their logs easier.

Spans themselves are contextualized by ancestors. I specifically wanted to _avoid_ a span hierarchy which serialized as `S3/Op-ID/Attempt-1/Signing/DefaultCredentialsProviderChain/CachedCredentialsProvider/ProfileCredentialsProvider` because that's overly verbose. Spans should identify a purpose (e.g., `Signing/Credentials/Cached/Profile`) whereas component names should identify a code location (e.g., `ProfileCredentialsProvider`).

That said, I fully agree with you and @aajtodd that we need a taxonomy for trace spans, almost certainly flatter and simpler than the strawman I've used in this iteration of the PRs.

ianbotsf:
Deprecating this PR in favor of using coroutine context.

Issue #718, opened 2022-10-05
Obey the `expireCredentialsAfter` in CachedCredentialsProvider


lauzadis:
### Describe the issue

[CachedCredentialsProvider has a field called `expireCredentialsAfter`](https://github.com/awslabs/aws-sdk-kotlin/blob/main/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CachedCredentialsProvider.kt#L49) which allows the user to specify a maximum amount of time to wait before refreshing credentials. It defaults to 15 minutes.

Currently, we do not use this field when we receive credentials with a non-null expiration time. So, if the default `expireCredentialsAfter` are used and we receive credentials with an expiration 30 minutes from now, we will only refresh them after those 30 minutes, but they should be refreshed after only 15 minutes.  

### Steps to Reproduce

- Create a CachedCredentialsProvider with `expireCredentialsAfter` time of 0 seconds
- Mock receiving some credentials with an expiration time sometime in the near future (i.e 5 mins, 10 mins, 30 mins)
- Try to getCredentials again. They will not be refreshed, you will be served the cached value

### Current behavior

N/A

### AWS Kotlin SDK version used

N/A

### Platform (JVM/JS/Native)

N/A

### Operating System and version

N/A

lauzadis:
Closed by #719

Issue #707, opened 2022-09-26
Implement static stability support for credentials providers


ianbotsf:
### Describe the feature

TBD

### Is your Feature Request related to a problem?

n/a

### Acknowledge

- [X] I may be able to implement this feature request

### AWS Kotlin SDK version used

0.17.6-beta

### Platform (JVM/JS/Native)

Any

### Operating System and version

Any

lauzadis:
Closed by #719

PR #719, opened 2022-10-05
feat(rt): support static stability for IMDS credentials

lauzadis:
Adds support for static stability for IMDS credentials. 
Uses the `expireCredentialsAfter` time in CachedCredentialsProvider for credentials which have an expiration time set.

## Issue \#
#707
#718

### \#707:

This change is required to comply with AWS SDK standards. Specifically, this will support the use of expired IMDS credentials. This is useful because in the case of service degradation in IMDS which prevents credentials refresh, AWS will allow us to keep using the expired credentials.

When we use expired credentials, the CachedCredentialsProvider will continuously try to refresh them. To solve this, a new field called `nextRefresh` was added to the Credentials data class. When present, this will override any expiration times, and the providers will only try to refresh the credentials if the `nextRefresh` time has passed.

### \#718:
When receiving credentials with an expiration time, set the cached credentials' expiration time to the minimum of `expireCredentialsAfter` and the received expiration time. This will ensure we are refreshing credentials at the rate the user specifies.

### DIFF
--- a/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
+++ b/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
@@ -146,8 +146,7 @@ public class ImdsCredentialsProvider(
     private suspend fun useCachedCredentials(ex: Exception): Credentials? = when {
         ex is SdkIOException || ex is EC2MetadataError && ex.statusCode == HttpStatusCode.InternalServerError.value -> {
             mu.withLock {
-                previousCredentials?.run { nextRefresh = clock.now() + DEFAULT_CREDENTIALS_REFRESH_SECONDS.seconds }
-                return previousCredentials
+                previousCredentials?.apply { nextRefresh = clock.now() + DEFAULT_CREDENTIALS_REFRESH_SECONDS.seconds }
             }
         }
         else -> null


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
```diff
@@ -56,23 +63,61 @@ public class ImdsCredentialsProvider(
             throw CredentialsNotLoadedException("AWS EC2 metadata is explicitly disabled; credentials not loaded")
         }
 
+        // if we have previously served IMDS credentials and it's not time for a refresh, just return the previous credentials
+        if (previousCredentials != null &&
```
lauzadis, line 67: Is there a nicer way to do this null check? Previously I had:
```kotlin
previousCredentials?.run {
    this.nextRefresh?.run {
        if (clock.now() < this) {
            return previousCredentials as Credentials
        }
    }
}
```
and was torn between the two

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
```diff
@@ -56,23 +63,61 @@ public class ImdsCredentialsProvider(
...
+
         val profileName = try {
             profile.get()
         } catch (ex: Exception) {
-            throw CredentialsProviderException("failed to load instance profile", ex)
+            when {
+                ex is IOException ||
+                    ex is EC2MetadataError && ex.statusCode == HttpStatusCode.InternalServerError.value -> {
+                    previousCredentials = previousCredentials?.copy(nextRefresh = clock.now() + DEFAULT_CREDENTIALS_REFRESH_SECONDS.seconds)
+                    return previousCredentials ?: throw CredentialsProviderException("failed to load instance profile", ex)
```
lauzadis, line 81: is there a nice way to remove this duplicated `throw ...`? (i.e jump from the elvis operator to the else block)

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
```diff
@@ -56,23 +63,61 @@ public class ImdsCredentialsProvider(
             throw CredentialsNotLoadedException("AWS EC2 metadata is explicitly disabled; credentials not loaded")
         }
 
+        // if we have previously served IMDS credentials and it's not time for a refresh, just return the previous credentials
+        if (previousCredentials != null &&
+            previousCredentials!!.nextRefresh != null &&
+            clock.now() < previousCredentials!!.nextRefresh!!
+        ) {
+            return previousCredentials as Credentials
+        }
```
ianbotsf, line 72: **Question**: Is this the correct layer for this behavior? This means we'll effectively cache the credentials at this layer and _not_ try fetching new ones until `nextRefresh`. That sounds correct for a caching layer but not what I'd necessarily expect from the direct IMDS provider.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
ianbotsf: I believe you could use the safe call operator and `takeIf` to clean this up slightly:

```kotlin
previousCredentials?.nextRefresh?.takeIf { Clock.now() < it }?.run {
    return previousCredentials as Credentials
}
```

ianbotsf: Why is the `as Credentials` cast necessary?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
```diff
@@ -16,13 +16,18 @@ import aws.smithy.kotlin.runtime.http.HttpStatusCode
 import aws.smithy.kotlin.runtime.io.Closeable
 import aws.smithy.kotlin.runtime.logging.Logger
 import aws.smithy.kotlin.runtime.serde.json.JsonDeserializer
+import aws.smithy.kotlin.runtime.time.Clock
 import aws.smithy.kotlin.runtime.util.Platform
 import aws.smithy.kotlin.runtime.util.PlatformEnvironProvider
 import aws.smithy.kotlin.runtime.util.asyncLazy
+import java.io.IOException
```
aajtodd, line 23: fix: JVM only (not sure why this even compiles)

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
aajtodd: Yeah this seems to duplicate the cache behavior does it not?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
```diff
@@ -56,23 +63,61 @@ public class ImdsCredentialsProvider(
...
+                    resp.accessKeyId,
+                    resp.secretAccessKey,
+                    resp.sessionToken,
+                    resp.expiration,
+                    PROVIDER_NAME,
+                )
+
+                if (creds.expiration!! < clock.now()) {
+                    // let the user know we will be using expired credentials
+                    logger.info { STATIC_STABILITY_LOG_MESSAGE.format(DEFAULT_CREDENTIALS_REFRESH_SECONDS / 60) }
```
aajtodd, line 114: fix: It looks like [String.format](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/format.html) is JVM only.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
```diff
@@ -56,23 +63,61 @@ public class ImdsCredentialsProvider(
...
+            is JsonCredentialsResponse.SessionCredentials -> {
+                var creds = Credentials(
+                    resp.accessKeyId,
+                    resp.secretAccessKey,
+                    resp.sessionToken,
+                    resp.expiration,
+                    PROVIDER_NAME,
+                )
+
+                if (creds.expiration!! < clock.now()) {
```
aajtodd, line 112: Question: Is this condition sufficient or should it include that `previousCredentials` wasn't `null` as well? I suppose it would be odd that IMDS would be returning expired creds so maybe this is ok?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
```diff
@@ -56,23 +63,61 @@ public class ImdsCredentialsProvider(
...
         return when (val resp = deserializeJsonCredentials(deserializer)) {
-            is JsonCredentialsResponse.SessionCredentials -> Credentials(
-                resp.accessKeyId,
-                resp.secretAccessKey,
-                resp.sessionToken,
-                resp.expiration,
-                PROVIDER_NAME,
-            )
+            is JsonCredentialsResponse.SessionCredentials -> {
+                var creds = Credentials(
```
aajtodd, line 104: Suggestion: I think we can restructure this to remove the mutable `creds` variable and the copy, something like:


```kotlin
val nextRefresh = if (resp.expiration != null && resp.expiration < clock.now()) {
     logger.warn(...)
     clock.now() + DEFAULT_CREDENTIALS_REFRESH_SECONDS.seconds
}else {
    null
}

return Credentials(..., nextRefresh = nextRefresh)
      .also { previousCredentials = it }

```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
aajtodd: Also probably log at `warn` level?

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProviderTest.kt
```diff
@@ -9,21 +9,34 @@ import aws.sdk.kotlin.runtime.config.AwsSdkSetting
...
 import aws.smithy.kotlin.runtime.http.response.HttpResponse
 import aws.smithy.kotlin.runtime.httptest.buildTestConnection
 import aws.smithy.kotlin.runtime.time.Instant
 import aws.smithy.kotlin.runtime.time.ManualClock
 import io.kotest.matchers.string.shouldContain
+import io.mockk.coVerify
+import io.mockk.spyk
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.test.runTest
+import java.net.SocketTimeoutException
```
aajtodd, line 32: fix: JVM only

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProviderTest.kt
```diff
@@ -200,4 +217,324 @@ class ImdsCredentialsProviderTest {
...
+        }
+
+        // make sure ImdsClient only gets called once
+        coVerify(exactly = 1) {
+            client.get(any())
+        }
+    }
+
+    @Test
+    fun testDontRefreshUntilNextRefreshTimeHasPassed() = runTest {
```
aajtodd, line 328: fix/correctness: This seems like it should be a test on the `CachedCredentialsProvider` rather than this provider (related to Ian's [comment](https://github.com/awslabs/aws-sdk-kotlin/pull/719/files#r989279250) I suppose)

aajtodd (changes requested):
Looks great overall. Couple questions and callouts (in addition to Ian's callout on thread safety)

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CachedCredentialsProvider.kt
```diff
@@ -58,9 +58,10 @@ public class CachedCredentialsProvider(
         logger.trace { "refreshing credentials cache" }
         val providerCreds = source.getCredentials()
         if (providerCreds.expiration != null) {
-            ExpiringValue(providerCreds, providerCreds.expiration!!)
+            val expirationTime = providerCreds.nextRefresh ?: minOf(providerCreds.expiration!!, (clock.now() + expireCredentialsAfter))
```
aajtodd, line 61: missed this in review: Missing test for this

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
```diff
@@ -56,23 +63,61 @@ public class ImdsCredentialsProvider(
...
+        }
+
         val profileName = try {
             profile.get()
         } catch (ex: Exception) {
-            throw CredentialsProviderException("failed to load instance profile", ex)
+            when {
+                ex is IOException ||
+                    ex is EC2MetadataError && ex.statusCode == HttpStatusCode.InternalServerError.value -> {
+                    previousCredentials = previousCredentials?.copy(nextRefresh = clock.now() + DEFAULT_CREDENTIALS_REFRESH_SECONDS.seconds)
```
aajtodd, line 80: Another miss in my initial review.

Question: Should this be the `DEFAULT_CREDENTIALS_REFRESH_SECONDS`? I see our guidance is random between 5-10 minutes which would attempt to refresh a little more aggressively.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
lauzadis: I think removing this would violate the guidelines for waiting 5-10 minutes between requests when the user has expired credentials and is using a direct IMDS provider, right?

This was added for the case when a user specifies a custom CredentialsProvider (i.e one consisting only of ImdsCredentialsProvider). If we remove this, then we will reach out to IMDS to get credentials [for every HTTP request](https://github.com/awslabs/smithy-kotlin/blob/cfa0fd3a30b4c50b75485786f043d4e2ad803f55/runtime/auth/aws-signing-default/common/src/aws/smithy/kotlin/runtime/auth/awssigning/DefaultAwsSigner.kt#L26), even if the returned credentials are expired. 

If we get expired credentials, we are supposed to wait 5-10 minutes before reaching out to IMDS again. This internal caching logic only runs when we have expired credentials (`nextRefresh != null`). If everything is operating normally, nextRefresh will remain null and there will be no caching.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
lauzadis: It is possible to receive expired credentials from IMDS when it's in static stability mode: 
> Test 1: SDK can use IMDS credential provider if first IMDS call returns expired credentials.

So you are proposing the conditional to be:
```kotlin
if (creds.expiration!! < clock.now() && previousCredentials != null)
```

I'm wondering what you think we should do if the incoming credentials are expired but previousCredentials is null? This can happen if we start an SDK client during an outage -- we will have no previousCredentials but the incoming credentials will be expired

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
aajtodd: Hmm I'm a little conflicted here. All of our providers work this way and would refresh on every HTTP request without a caching provider wrapping it. We even note this in some of the provider [documentation](https://github.com/awslabs/aws-sdk-kotlin/blob/230751e55e50afe97c263bcf2698202b959a18b0/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt#L35-L36).

On one hand I would sort of like to keep that clean separation, on the other I can understand your point. 🤔 
@ianbotsf thoughts?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
aajtodd: I think either is fine I was mostly just asking if we think we should be more aggressive here or not

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
aajtodd: > This can happen if we start an SDK client during an outage

Hmm good point. That seems sufficient then

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
lauzadis: without `as Credentials`, I get this error: `Smart cast to 'Credentials' is impossible, because 'previousCredentials' is a mutable property that could have been changed by this time`

Another option that works is adding the non-null assertion `!!`
```kotlin
mu.withLock {
    previousCredentials?.nextRefresh?.takeIf { clock.now() < it }?.run {
        return previousCredentials!!
    }
}
```

I will change it to that.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
ianbotsf: I think the logic of using cached credentials from within the `ImdsCredentialsProvider` is correct _if_ we're in stability mode (because that's an implementation detail of how to use IMDS). If we're not in stability mode, I'd expect the provider to fetch fresh credentials every time `getCredentials` is called (just like every other provider does) and relegate _regular_ caching to the `CachedCredentialsProvider`.

How complicated would that be to implement?

lauzadis: There is no way to tell static stability is enabled with 100% certainty, so we can't really create any logic around that:
> IMDS doesn’t provide a formal flag, so the SDKs must always assume Static Stability is enabled.

One solution is to remove the `nextRefresh` logic from CachedCredentialsProvider entirely and leave the expired-credential caching solely in ImdsCredentialsProvider. In this scenario, when the default CachedCredentialsProvider is used and we are in static stability mode, the IMDS credentials will _always_ appear as expired. To fetch credentials, we walk the chain of CredentialsProviders, and when we reach ImdsCredentialsProvider, we will branch into this caching logic and return the credentials.

One downside is that we will have to check the entire chain, which could be costly. But the added overhead could be fixed [with this TODO](https://github.com/awslabs/aws-sdk-kotlin/blob/230751e55e50afe97c263bcf2698202b959a18b0/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CredentialsProviderChain.kt#L13) where we would cache the credential provider.

ianbotsf: Will we know we're in stability mode the first time we call IMDS and get either expired credentials or some type of service error?

lauzadis: @ianbotsf We could but it would not be 100% certain. For example, there may have been a service error for some other reason (i.e there's an availability issue, but static stability is not enabled yet)

@aajtodd I am leaning towards the idea I proposed. One more possible downside is that we are adding a whole new field to Credentials (`nextRefresh`) just to be used in ImdsCredentialsProvider. Would it make sense to keep this as a member in ImdsCredentialsProvider instead of in the Credentials data class?

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProviderTest.kt
```diff
@@ -194,10 +211,321 @@ class ImdsCredentialsProviderTest {
...
+        )
+
+        val provider = ImdsCredentialsProvider(
+            profileOverride = "imds-test-role",
+            client = lazyOf(client),
+            clock = testClock,
+        )
+
+        // call getCredentials 3 times
+        for (i in 1..3) {
```
aajtodd, line 316: nit: `repeat(3) { ... }`

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
```diff
@@ -56,23 +69,70 @@ public class ImdsCredentialsProvider(
...
+                }
+            }
+        }
+
         val profileName = try {
             profile.get()
         } catch (ex: Exception) {
-            throw CredentialsProviderException("failed to load instance profile", ex)
+            when {
+                ex is SdkIOException ||
```
aajtodd, line 85: Lines 85-92 are nearly same as 99-107, we could probably consolidate

e.g.

```kotlin
val profileName = try {
    ... 
} catch(ex :Exception) {
    return useCachedCredentials(ex) ?: throw CredentialsProviderException("failed to load instance profile", ex)
}


val payload = try {
    ... 
} catch(ex :Exception) {
    return useCachedCredentials(ex) ?: throw CredentialsProviderException("failed to load credentials", ex)
}



private fun useCachedCredentials(ex: Exception): Credentials? = when(ex) {
    is SdkIOException || is EC2MetadataError && ex.statusCode == HttpStatusCode.InternalServerError.value -> {
        mu.withLock {
            previousCredentials?.run { nextRefresh = clock.now() + ... }
            previousCredentials        
         }
     }
     else -> null
}
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
```diff
@@ -43,8 +50,14 @@ public class ImdsCredentialsProvider(
     private val profileOverride: String? = null,
     private val client: Lazy<InstanceMetadataProvider> = lazy { ImdsClient() },
     private val platformProvider: PlatformEnvironProvider = Platform,
+    private var previousCredentials: Credentials? = null,
+    private val clock: Clock = Clock.System,
 ) : CredentialsProvider, Closeable {
     private val logger = Logger.getLogger<ImdsCredentialsProvider>()
+    private val mu = Mutex()
```
aajtodd, line 57: nit: You might clarify what this mutex protects:

e.g.

```kotlin
// protects previousCredentials + nextRefresh
private val mu = Mutex()
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
```diff
@@ -43,8 +50,14 @@ public class ImdsCredentialsProvider(
     private val profileOverride: String? = null,
     private val client: Lazy<InstanceMetadataProvider> = lazy { ImdsClient() },
     private val platformProvider: PlatformEnvironProvider = Platform,
+    private var previousCredentials: Credentials? = null,
```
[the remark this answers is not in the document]
lauzadis: It was nice for testing purposes, but not necessary. I will refactor it and update the test cases

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
```diff
@@ -16,14 +16,21 @@ import aws.smithy.kotlin.runtime.http.HttpStatusCode
...
 import aws.smithy.kotlin.runtime.util.asyncLazy
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlin.time.Duration.Companion.seconds
 
 private const val CREDENTIALS_BASE_PATH: String = "/latest/meta-data/iam/security-credentials"
 private const val CODE_ASSUME_ROLE_UNAUTHORIZED_ACCESS: String = "AssumeRoleUnauthorizedAccess"
 private const val PROVIDER_NAME = "IMDSv2"
 
+public expect class SdkIOException : Exception // FIXME move this to the proper place when we do the larger KMP Exception refactor
```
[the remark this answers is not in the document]
lauzadis: When I switch this to `internal` or `private`, I get the following error in the JVM definition of SDKIOException:
```
Actual typealias 'SdkIOException' has no corresponding expected declaration The following declaration is incompatible because visibility is different:
internal final expect class SdkIOException : Exception /* = Exception */ defined in aws.sdk.kotlin.runtime.auth.credentials in file ImdsCredentialsProvider.kt
```

I played around with it for a while and this `expect` seems to be required to be public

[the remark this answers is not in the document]
aajtodd: Expect and actual visibility has to match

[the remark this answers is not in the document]
lauzadis: They were both matching in this case -- I tried all combinations and it does not compile when both are set to `internal`. This seems to be a common issue: https://youtrack.jetbrains.com/issue/KT-37316

It can be fixed by adding `@Suppress("ACTUAL_WITHOUT_EXPECT")` decorator to the JVM implementation. I will go that route unless anybody is against it?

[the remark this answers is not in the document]
aajtodd: Ahh I see, because the type aliased actual differs in visibility. The suppression suggested seems fine then

aajtodd (approved):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt
```diff
@@ -98,4 +142,14 @@ public class ImdsCredentialsProvider(
             throw ex
         }
     }
+
+    private suspend fun useCachedCredentials(ex: Exception): Credentials? = when {
+        ex is SdkIOException || ex is EC2MetadataError && ex.statusCode == HttpStatusCode.InternalServerError.value -> {
+            mu.withLock {
+                previousCredentials?.run { nextRefresh = clock.now() + DEFAULT_CREDENTIALS_REFRESH_SECONDS.seconds }
+                return previousCredentials
```
ianbotsf, line 150: **Nit**: I believe this can be simplified to:

```
previousCredentials?.apply { nextRefresh = clock.now() + DEFAULT_CREDENTIALS_REFRESH_SECONDS.seconds }
```

ianbotsf (approved):
Looks pretty good to me!

PR #721, opened 2022-10-06
chore: migrate codegen diff tooling to S3 CDN

aajtodd:
## Issue \#
N/A

## Description of changes
Migrates our codegen diff tooling to generate HTML diffs and upload them to S3. Previously we would generate codegen output and check it in and compare against the `__generated-main` branch (which is updated when merges to main happen). This new way will allow properly diffing against the target base ref for any arbitrary branch. 


aajtodd:
Example of output can be found here: 
https://github.com/awslabs/aws-sdk-kotlin/pull/722

ianbotsf (approved):

lauzadis (approved):

PR #729, opened 2022-10-14
feat(rt): add ProcessCredentialsProvider

lauzadis:
Adds a ProcessCredentialsProvider which will be used to support [process credentials](https://docs.aws.amazon.com/sdkref/latest/guide/feature-process-credentials.html) resolution

## Issue \#
N/A

## Description of changes
This change is required to achieve feature parity with existing SDKs. It allows more customization on where the SDK gets users' credentials from.

Process credentials may be returned without an expiration field. In this case, we are expected to treat the credentials as non-expiring, and no credential refreshes should be attempted. To support this behavior, the concept of `NonExpiringCredentials` was added to the JSON deserializer used for credentials. If we receive non-expiring credentials, we set the expiration to `Instant.MAX_VALUE` to ensure they "never" expire. 

### DIFF
--- a/.changes/07860a56-6515-4545-8c04-4bbabaca7d36.json
+++ b/.changes/07860a56-6515-4545-8c04-4bbabaca7d36.json
@@ -1,0 +1,5 @@
+{
+    "id": "07860a56-6515-4545-8c04-4bbabaca7d36",
+    "type": "misc",
+    "description": "Refactor the way service client configuration is generated"
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [0.19.5-beta] - 01/19/2023
+
+### Miscellaneous
+* Sync AWS service models.
+
 ## [0.19.4-beta] - 01/13/2023
 
 ### Miscellaneous
--- a/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
+++ b/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
@@ -1,0 +1,44 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package aws.sdk.kotlin.runtime.config
+
+import aws.sdk.kotlin.runtime.client.AwsSdkClientConfig
+import aws.sdk.kotlin.runtime.config.retries.resolveRetryStrategy
+import aws.sdk.kotlin.runtime.region.resolveRegion
+import aws.smithy.kotlin.runtime.client.SdkClient
+import aws.smithy.kotlin.runtime.client.SdkClientConfig
+import aws.smithy.kotlin.runtime.client.SdkClientFactory
+
+/**
+ * Abstract base class all AWS client companion objects inherit from
+ *
+ * @param TConfig the type of the service client configuration
+ * @param TConfigBuilder the type of builder that creates [TConfig] instances
+ * @param TClient the type of service client
+ * @param TClientBuilder the type of builder that creates [TClient] instances
+ */
+public abstract class AbstractAwsSdkClientFactory<
+    TConfig,
+    TConfigBuilder,
+    TClient : SdkClient,
+    TClientBuilder : SdkClient.Builder<TConfig, TConfigBuilder, TClient>,
+    > : SdkClientFactory<TConfig, TConfigBuilder, TClient, TClientBuilder>
+    where TConfig : SdkClientConfig,
+          TConfig : AwsSdkClientConfig,
+          TConfigBuilder : SdkClientConfig.Builder<TConfig>,
+          TConfigBuilder : AwsSdkClientConfig.Builder {
+    /**
+     * Construct a [TClient] by resolving the configuration from the current environment.
+     */
+    public suspend fun fromEnvironment(block: (TConfigBuilder.() -> Unit)? = null): TClient {
+        val builder = builder()
+        if (block != null) builder.config.apply(block)
+
+        builder.config.region = builder.config.region ?: resolveRegion()
+        builder.config.retryStrategy = builder.config.retryStrategy ?: resolveRetryStrategy()
+        return builder.build()
+    }
+}
--- a/aws-runtime/aws-config/jvm/test/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactoryTest.kt
+++ b/aws-runtime/aws-config/jvm/test/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactoryTest.kt
@@ -1,0 +1,102 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package aws.sdk.kotlin.runtime.config
+
+import aws.sdk.kotlin.runtime.client.AwsSdkClientConfig
+import aws.smithy.kotlin.runtime.client.AbstractSdkClientBuilder
+import aws.smithy.kotlin.runtime.client.SdkClient
+import aws.smithy.kotlin.runtime.client.SdkClientConfig
+import aws.smithy.kotlin.runtime.client.SdkLogMode
+import aws.smithy.kotlin.runtime.retries.Outcome
+import aws.smithy.kotlin.runtime.retries.RetryOptions
+import aws.smithy.kotlin.runtime.retries.RetryStrategy
+import aws.smithy.kotlin.runtime.retries.StandardRetryStrategy
+import aws.smithy.kotlin.runtime.retries.policy.RetryPolicy
+import io.kotest.extensions.system.withSystemProperties
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertIs
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class AbstractAwsSdkClientFactoryTest {
+    @Test
+    fun testFromEnvironmentFavorsExplicitConfig() = runTest {
+        val explicitRegion = "explicit-region"
+        val explicitRetryStrategy = StandardRetryStrategy()
+        val client = TestClient.fromEnvironment {
+            region = explicitRegion
+            retryStrategy = explicitRetryStrategy
+        }
+
+        assertEquals(explicitRegion, client.config.region)
+        assertEquals(explicitRetryStrategy, client.config.retryStrategy)
+    }
+
+    @Test
+    fun testFromEnvironmentResolvesDefaults() = runTest {
+        withSystemProperties(
+            mapOf(
+                AwsSdkSetting.AwsRegion.jvmProperty to "resolved-region",
+                AwsSdkSetting.AwsRetryMode.jvmProperty to "standard",
+            ),
+        ) {
+            val client = TestClient.fromEnvironment()
+            assertEquals("resolved-region", client.config.region)
+            assertIs<StandardRetryStrategy>(client.config.retryStrategy)
+        }
+    }
+}
+
+private interface TestClient : SdkClient {
+    override val serviceName: String
+        get() = "test"
+
+    override val config: Config
+
+    // refactored: mostly in an abstract base now
+    companion object : AbstractAwsSdkClientFactory<Config, Config.Builder, TestClient, Builder>() {
+        override fun builder(): Builder = Builder()
+    }
+
+    class Builder internal constructor() : AbstractSdkClientBuilder<Config, Config.Builder, TestClient>() {
+        override val config: Config.Builder = Config.Builder()
+
+        override fun newClient(config: Config): TestClient {
+            return DefaultTestClient(config)
+        }
+    }
+
+    class Config private constructor(builder: Config.Builder) : SdkClientConfig, AwsSdkClientConfig {
+        override val retryStrategy: RetryStrategy = builder.retryStrategy ?: TestRetryStrategy()
+        override val region: String = builder.region ?: error("region is required")
+
+        // new: inherits builder equivalents for Config base classes
+        class Builder : AwsSdkClientConfig.Builder, SdkClientConfig.Builder<Config> {
+            override var region: String? = null
+            override var retryStrategy: RetryStrategy? = null
+            override var sdkLogMode: SdkLogMode = SdkLogMode.Default
+            override fun build(): Config = Config(this)
+        }
+    }
+}
+
+private class DefaultTestClient(
+    override val config: TestClient.Config,
+) : TestClient {
+    override fun close() { }
+}
+
+// some non standard retry strategy used as a marker
+private class TestRetryStrategy : RetryStrategy {
+    override val options: RetryOptions
+        get() = error("not needed for test")
+
+    override suspend fun <R> retry(policy: RetryPolicy<R>, block: suspend () -> R): Outcome<R> {
+        error("not needed for test")
+    }
+}
--- a/aws-runtime/aws-core/common/src/aws/sdk/kotlin/runtime/client/AwsSdkClientConfig.kt
+++ b/aws-runtime/aws-core/common/src/aws/sdk/kotlin/runtime/client/AwsSdkClientConfig.kt
@@ -1,0 +1,30 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package aws.sdk.kotlin.runtime.client
+
+import aws.smithy.kotlin.runtime.client.SdkClientConfig
+
+/**
+ * Base interface all generated AWS SDK Kotlin clients implement
+ */
+public interface AwsSdkClientConfig : SdkClientConfig {
+
+    /**
+     * The AWS region (e.g. `us-west-2`) to make requests to. See about AWS
+     * [global infrastructure](https://aws.amazon.com/about-aws/global-infrastructure/regions_az/) for more
+     * information
+     */
+    public val region: String
+
+    public interface Builder {
+        /**
+         * The AWS region (e.g. `us-west-2`) to make requests to. See about AWS
+         * [global infrastructure](https://aws.amazon.com/about-aws/global-infrastructure/regions_az/) for more
+         * information
+         */
+        public var region: String?
+    }
+}
--- a/codegen/sdk/aws-models/billingconductor.json
+++ b/codegen/sdk/aws-models/billingconductor.json
@@ -122,15 +122,6 @@
                                                     "ref": "Endpoint"
                                                 }
                                             ]
-                                        },
-                                        {
-                                            "fn": "parseURL",
-                                            "argv": [
-                                                {
-                                                    "ref": "Endpoint"
-                                                }
-                                            ],
-                                            "assign": "url"
                                         }
                                     ],
                                     "type": "tree",
@@ -266,15 +257,7 @@
                                                             "conditions": [],
                                                             "endpoint": {
                                                                 "url": "https://billingconductor-fips.{Region}.api.aws",
-                                                                "properties": {
-                                                                    "authSchemes": [
-                                                                        {
-                                                                            "name": "sigv4",
-                                                                            "signingRegion": "us-east-1",
-                                                                            "signingName": "billingconductor"
-                                                                        }
-                                                                    ]
-                                                                },
+                                                                "properties": {},
                                                                 "headers": {}
                                                             },
                                                             "type": "endpoint"
@@ -326,15 +309,7 @@
                                                             "conditions": [],
                                                             "endpoint": {
                                                                 "url": "https://billingconductor-fips.{Region}.amazonaws.com",
-                                                                "properties": {
-                                                                    "authSchemes": [
-                                                                        {
-                                                                            "name": "sigv4",
-                                                                            "signingRegion": "us-east-1",
-                                                                            "signingName": "billingconductor"
-                                                                        }
-                                                                    ]
-                                                                },
+                                                                "properties": {},
                                                                 "headers": {}
                                                             },
                                                             "type": "endpoint"
@@ -386,15 +361,7 @@
                                                             "conditions": [],
                                                             "endpoint": {
                                                                 "url": "https://billingconductor.{Region}.api.aws",
-                                                                "properties": {
-                                                                    "authSchemes": [
-                                                                        {
-                                                                            "name": "sigv4",
-                                                                            "signingRegion": "us-east-1",
-                                                                            "signingName": "billingconductor"
-                                                                        }
-                                                                    ]
-                                                                },
+                                                                "properties": {},
                                                                 "headers": {}
                                                             },
                                                             "type": "endpoint"
@@ -487,12 +454,18 @@
                                             "rules": [
                                                 {
                                                     "conditions": [],
-                                                    "endpoint": {
-                                                        "url": "https://billingconductor-fips.{Region}.{PartitionResult#dualStackDnsSuffix}",
-                                                        "properties": {},
-                                                        "headers": {}
-                                                    },
-                                                    "type": "endpoint"
+                                                    "type": "tree",
+                                                    "rules": [
+                                                        {
+                                                            "conditions": [],
+                                                            "endpoint": {
+                                                                "url": "https://billingconductor-fips.{Region}.{PartitionResult#dualStackDnsSuffix}",
+                                                                "properties": {},
+                                                                "headers": {}
+                                                            },
+                                                            "type": "endpoint"
+                                                        }
+                                                    ]
                                                 }
                                             ]
                                         },
@@ -597,12 +570,18 @@
                                             "rules": [
                                                 {
                                                     "conditions": [],
-                                                    "endpoint": {
-                                                        "url": "https://billingconductor.{Region}.{PartitionResult#dualStackDnsSuffix}",
-                                                        "properties": {},
-                                                        "headers": {}
-                                                    },
-                                                    "type": "endpoint"
+                                                    "type": "tree",
+                                                    "rules": [
+                                                        {
+                                                            "conditions": [],
+                                                            "endpoint": {
+                                                                "url": "https://billingconductor.{Region}.{PartitionResult#dualStackDnsSuffix}",
+                                                                "properties": {},
+                                                                "headers": {}
+                                                            },
+                                                            "type": "endpoint"
+                                                        }
+                                                    ]
                                                 }
                                             ]
                                         },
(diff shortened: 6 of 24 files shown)


aajtodd (changes requested):
Looks pretty good overall! Couple fixes and suggestions

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProcessCredentialsProvider.kt
```diff
@@ -0,0 +1,70 @@
...
+                resp.secretAccessKey,
+                resp.sessionToken,
+                resp.expiration,
+                PROVIDER_NAME,
+            )
+            is JsonCredentialsResponse.NonExpiringCredentials -> Credentials(
+                resp.accessKeyId,
+                resp.secretAccessKey,
+                resp.sessionToken,
+                Instant.MAX_VALUE,
```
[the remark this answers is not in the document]
lauzadis: Yeah, I was going to use this initially, but Java's `Instant` class does not like taking Long.MAX_VALUE as input.

```
java.time.DateTimeException: Instant exceeds minimum or maximum instant
```

Also the epoch-seconds-date of Long.MAX_VALUE is sometime in 2262, but Instant.MAX_VALUE is [defined to be `1000000000-01-01T00:00Z`](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/time/Instant.html#MAX) which is much further in the future.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/JsonCredentialsDeserializer.kt
```diff
@@ -128,3 +136,58 @@ internal suspend fun deserializeJsonCredentials(deserializer: Deserializer): Jso
         else -> JsonCredentialsResponse.Error(code, message)
     }
 }
+
+/**
+ * Deserialize credentials coming from process credentials. Used by [ProcessCredentialsProvider].
+ * The difference between this and [deserializeJsonCredentials] is that process credentials _must_ provide a version field,
+ * the session token field is called `SessionToken` instead of `Token`, and the expiration field is optional.
+ */
+internal fun deserializeJsonProcessCredentials(deserializer: Deserializer): JsonCredentialsResponse {
```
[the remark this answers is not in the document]
lauzadis: I agree that the duplication is not great. Initially I followed your suggestion to only validate `Version` as 1 if it's present, but the specification says that the `Version` field is **required** for process credentials, so I wanted to enforce that and didn't see a way to keep it backwards-compatible with the current credentials providers.

Another issue I ran into when re-using the existing `deserializeJsonCredentials` is parsing the session token. The current credentials providers parse it from the `Token` key, but credentials providers provide it using the `SessionToken` key.

So the `Version` and `SessionToken` vs. `Token` issues made me lean towards creating an entirely new method instead. I guess it's a question of special-casing in the current deserializer function or making an entirely new one for this process credentials format.

lauzadis:
Question: should I add `@InternalSdkAPI` decorators to the new functions such as `executeCommand`? It's not really intended for SDK users and could be subject to breaking changes in the future (i.e if we want to return stdout **and** stderr instead of just one of them, or generalize it even more with a larger refactor)

aajtodd:
> Question: should I add `@InternalSdkAPI` decorators to the new functions such as `executeCommand`? It's not really intended for SDK users and could be subject to breaking changes in the future (i.e if we want to return stdout **and** stderr instead of just one of them, or generalize it even more with a larger refactor)


It's marked `internal` already so it isn't consumable by anything other than from this project/compilation unit.

aws-runtime/aws-config/jvm/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommand.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+internal actual suspend fun executeCommand(command: String, isWindows: Boolean): Pair<Int, String> {
+    val cmd = ArrayList<String>()
+
+    // add the platform's shell
+    if (isWindows) {
+        cmd.add("cmd.exe")
+        cmd.add("/C")
+    } else {
+        cmd.add("sh")
```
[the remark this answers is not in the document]
lauzadis: Passing the command to a shell rather than starting it ourselves gives a little bit more configurability for users. 

For example, there are some things that a direct `.exec` call does not support, but sending command through the shell does support:
- variable substitution
- squiggle (~) / bang (!) / globbing (*) support 
- see more here: https://stackoverflow.com/a/68250664

The specification would be completely satisfied by executing the user-supplied command directly, but this just adds more configurability. This is also how the other SDKs do it, so I think we should support features like variable substitution for compatibility with scripts used today.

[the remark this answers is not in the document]
ianbotsf: This also alleviates us from the responsibility of parsing a single string into the command-and-args-array expected by `ProcessBuilder`. Correctly handling quoted and nested arguments is tricky.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/JsonCredentialsDeserializer.kt
```diff
@@ -29,6 +30,13 @@ internal sealed class JsonCredentialsResponse {
         val expiration: Instant,
     ) : JsonCredentialsResponse()
 
+    // Credentials that don't expire
+    data class NonExpiringCredentials(
```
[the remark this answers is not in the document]
lauzadis: Yeah, that would work. It would require some extra checks in the other providers (which today expect `expiration` to be non-null).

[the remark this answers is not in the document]
aajtodd: They don't really expect it though, they just pass it through to `Credentials` constructor which takes `Instant?` don't they?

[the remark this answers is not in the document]
lauzadis: ImdsCredentialsProvider expects it to be non-null [when calculating the nextRefresh time](https://github.com/awslabs/aws-sdk-kotlin/blob/main/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ImdsCredentialsProvider.kt#L99). I can add a non-null check here and it will behave the same

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/JsonCredentialsDeserializer.kt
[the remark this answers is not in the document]
lauzadis: Yeah I would like to get the rest of the team's opinion on this. The only one concern I have with your suggestion (and what I initially implemented) is that we will not be **requiring** the `Version` field to be set, which is required by the specification:

> The Version key must be set to 1 (as an integer type).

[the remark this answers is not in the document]
aajtodd: > I'd much prefer we were able to codegen these structs and deserializers just like model shapes

That would pretty much guarantee that they couldn't be shared (it also wouldn't allow custom logic like validating `version` or dealing with different versions).

> we also do this in [S3ErrorMetadata](https://github.com/awslabs/aws-sdk-kotlin/blob/230751e55e50afe97c263bcf2698202b959a18b0/services/s3/common/src/aws/sdk/kotlin/services/s3/internal/S3ErrorMetadata.kt#L45-L78)

Maybe I'm missing something but I'm not sure what the issue is with hand written deserializers? Even libraries like `kotlinx.serialization`, `Jackson`, and others allow you to write serde code by hand when the default annotations/bindings don't yield what you want. This seems entirely within the expected use to me.

I can live with the duplication it just irks me because the whole idea of having a sealed type was to deal with differences in JSON responses. It would seem the expected class of JSON responses is more dissimilar than we originally expected :(

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProcessCredentialsProvider.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        return when (val resp = deserializeJsonProcessCredentials(deserializer)) {
+            is JsonCredentialsResponse.SessionCredentials -> Credentials(
+                resp.accessKeyId,
+                resp.secretAccessKey,
+                resp.sessionToken,
+                resp.expiration ?: Instant.MAX_VALUE,
+                PROVIDER_NAME,
+            )
+            else -> {
```
aajtodd, line 57: style: unnecessary block `{}`

aajtodd (approved):
Looks great. Nice work. Have you tested this out with real CLI process on different OS (e.g. linux + windows)?

lauzadis:
> Looks great. Nice work. Have you tested this out with real CLI process on different OS (e.g. linux + windows)?

Yeah, I just finished manual testing on all platforms (mac, linux, windows) and it works everywhere. I tested for command successes and failures (where stderr message is expected) for both absolute command paths and resolving the command from the system's PATH.

lauzadis:
Changes since last review:
- Add a configurable max bytes read from stdout
- Add a configurable command timeout
- Handle spaces in command by wrapping the entire string in quotes
- Add security warning to the KDocs
- Prevent calling getCredentials() after close() in CachedCredentialsProvider

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CachedCredentialsProvider.kt
```diff
@@ -53,8 +53,9 @@ public class CachedCredentialsProvider(
 ) : CredentialsProvider, Closeable {
 
     private val cachedCredentials = CachedValue<Credentials>(null, bufferTime = refreshBufferWindow, clock)
+    private var isClosed: Boolean = false
 
-    override suspend fun getCredentials(): Credentials = cachedCredentials.getOrLoad {
+    override suspend fun getCredentials(): Credentials = if (!isClosed) cachedCredentials.getOrLoad {
```
ianbotsf, line 58: **Correctness**: Access/mutation of `isClosed` should be thread-safe.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CachedCredentialsProvider.kt
```diff
@@ -65,9 +66,11 @@ public class CachedCredentialsProvider(
...
             ExpiringValue(creds, expiration)
         }
-    }
+    } else throw IllegalStateException("Credentials provider is closed")
 
     override fun close() {
+        isClosed = true
+        cachedCredentials.close()
         (source as? Closeable)?.close()
     }
```
ianbotsf, line 75: **Suggestion**: Since we now track closed status, consider only invoking this logic if `isClosed == false`.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProcessCredentialsProvider.kt
```diff
@@ -4,37 +4,48 @@ import aws.smithy.kotlin.runtime.auth.awscredentials.Credentials
...
+): Pair<Int, String>
 
 private const val PROVIDER_NAME = "Process"
 
 /**
  * [CredentialsProvider] that invokes a command to retrieve [Credentials] from stdout.
  * If the exit code is non-zero, log and throw an exception including the stderr from the command.
  *
+ * WARNING: This [CredentialsProvider] sources credentials from an external process.
+ * This can potentially be dangerous, so proceed with caution. Other credential providers should be preferred if at all possible.
```
ianbotsf, line 27: **Nit**: "at all" is redundant.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProcessCredentialsProvider.kt
```diff
@@ -4,37 +4,48 @@ import aws.smithy.kotlin.runtime.auth.awscredentials.Credentials
...
+ * If using this option, you should make sure that the config file and any process or script files are as locked down as possible using
+ * security best practices for your operating system.
```
ianbotsf, line 29: **Style**: This feels both verbose and accusatory (via the word "you"). Perhaps it would be clearer, friendlier, and still equally cautionary to say something like:

> If using this option, please secure the config file and process/script files using the best practices for your operating system.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProcessCredentialsProvider.kt
```diff
@@ -4,37 +4,48 @@ import aws.smithy.kotlin.runtime.auth.awscredentials.Credentials
...
+    private val maxOutputLengthBytes: Long = 64 * 1024,
+    private val timeoutMillis: Long = 60_000,
 ) : CredentialsProvider {
     override suspend fun getCredentials(): Credentials {
         val logger = Logger.getLogger<ProcessCredentialsProvider>()
 
         val (exitCode, output) = try {
-            executeCommand(credentialProcess, platformProvider.osInfo().family == OsFamily.Windows)
+            executeCommand("\"$credentialProcess\"", platformProvider, maxOutputLengthBytes, timeoutMillis)
         } catch (ex: Exception) {
```
ianbotsf, line 49: **Question**: [Public documentation for this feature](https://docs.aws.amazon.com/sdkref/latest/guide/feature-process-credentials.html#feature-process-credentials-detail-path) already shows examples of putting the process name in quotes. Will this logic lead to two pairs of quotes around process names? Does that work correctly?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProcessCredentialsProvider.kt
```diff
@@ -52,19 +63,10 @@ public class ProcessCredentialsProvider(
...
-                resp.secretAccessKey,
-                resp.sessionToken,
-                Instant.MAX_VALUE,
+                resp.expiration ?: Instant.MAX_VALUE,
                 PROVIDER_NAME,
             )
-            is JsonCredentialsResponse.Error -> {
-                throw CredentialsProviderException("Error parsing credentials from process: code=${resp.code}; ${resp.message}")
-            }
+            else -> throw CredentialsProviderException("Credentials response was not of expected format")
```
ianbotsf, line 69: **Question**: The extensible credentials providers spec says we should make the error message available to the user. Why did we remove the error code and message from the thrown exception?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/CachedValue.kt
```diff
@@ -59,4 +60,6 @@ internal class CachedValue<T> (
         val refreshed = initializer().also { value = it }
         return refreshed.value
     }
+
+    override fun close() { value = null }
```
ianbotsf, line 64: **Correctness**: Wrap the `value` mutation in the mutex lock.

ianbotsf, line 64: **Comment**: A bit outside the scope of this PR but the name of this class makes even less sense now that it's closeable. This class isn't really a "cached value"...it's a "caching value provider" or, more commonly, a "cache".

aws-runtime/aws-config/jvm/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommandJVM.kt
```diff
@@ -0,0 +1,67 @@
...
+    return withContext(Dispatchers.IO) {
+        val process = ProcessBuilder().command(cmd).start()
+
+        val reader = process.inputStream.bufferedReader()
+
+        val output = StringBuilder()
+        val start = clock.now().epochMilliseconds
+
+        while (process.isAlive && clock.now().epochMilliseconds - start < timeoutMillis) {
+            reader.readLine()?.let { output.append(it) }
```
ianbotsf, line 50: **Correctness**: This reads lines from the process output but writes inline to the `StringBuilder` thus munching all the newlines.

aws-runtime/aws-config/jvm/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommandJVM.kt
```diff
@@ -0,0 +1,67 @@
...
+
+            if (output.length > maxOutputLengthBytes) {
+                throw RuntimeException("Process output exceeded limit of $maxOutputLengthBytes bytes")
+            }
+        }
```
ianbotsf, line 55: **Question**: This loop seems like it could be fairly CPU-intensive. How does this perform on processes that take a long time to return (e.g., due to network calls)?

ianbotsf (changes requested):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CachedCredentialsProvider.kt
aajtodd: See [HttpClientEngineBase](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/engine/HttpClientEngine.kt#L45) for an example of using an atomic for this purpose.

**style**: This would also be clearer written as a normal function body at this point IMO:

```kotlin
suspend fun getCredentials(): Credentials {
    check(!isClosed) { ... }

    return cachedCredentials.getOrLoad { ... }
}
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CachedCredentialsProvider.kt
aajtodd: **fix**: I'll take Ian's suggestion a bit further and say that we want to make close idempotent. If we are already closed just return.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/CachedValue.kt
aajtodd: **correctness**: I don't think this type should implement `Closeable`. It's not a resource and could theoretically be used again after "closing" it.  I'd probably just add a new function such as `evict()` that sets the expired value to `null`. Then we should just evict the cached value in the `close()` implementation of `CachedCredentialsProvider`

aws-runtime/aws-config/jvm/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommandJVM.kt
aajtodd: Also how does this respect timeouts? At any rate it seems like it would just be easier to wrap the reader loop in a `withTimeout` block, e.g.:

```kotlin

val process = ...
val reader = ...

withTimeout(timeout) {

    while(process.isAlive) {
        reader.readLine()?.let { ... } // or whatever the appropriate read logic is

        // check output length, etc
    }
}

```

aajtodd (changes requested):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProcessCredentialsProvider.kt
lauzadis: Sure, I can update the phrasing.

FWIW, this warning statement was taken directly from the [public documentation](https://docs.aws.amazon.com/sdkref/latest/guide/feature-process-credentials.html) and only slightly modified to reference process/script files in addition to the config file.

aws-runtime/aws-config/jvm/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommandJVM.kt
lauzadis: It respects timeouts because `reader.readline()` does not block, it either returns a line of text or `null` if no text is available. So the timeout condition in the `while` loop is constantly being evaluated.

I think `withTimeout` is a clearer option though 👍 

I will also refactor to avoid consuming the newline characters

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/CachedValue.kt
lauzadis: Wrapping this mutation with the mutex makes the function suspending, which means it can't be called from a `Closeable` because the overridden `close()` function can't be suspending.

I do think a mutex is necessary here to ensure we are nullifying the value properly. So, should I remove the implementation of `Closeable` in `CachedCredentialsProvider` and make `close()` a suspending function?

lauzadis: Actually that idea will not work, because [the credentials chain only closes `Closeable`s](https://github.com/awslabs/aws-sdk-kotlin/blob/main/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CredentialsProviderChain.kt#L55)

ianbotsf: I take it this concern is allayed if we go remove `Closeable` and go with `evict` instead?

lauzadis: I removed `Closeable` and renamed the function, but the race condition still exists right?

Without a mutex, if one thread calls `evict()` while another tries to `get()` the value, it's not clear what will happen.
If I make `evict()` use a mutex, it cannot be called from `CachedCredentialsProvider`'s `close()`.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProcessCredentialsProvider.kt
lauzadis: This is a different error message that gets thrown during parsing, it was not modified since the last review.

The error message provided to the user is not removed, it's here: https://github.com/awslabs/aws-sdk-kotlin/pull/729/files#diff-3c4a3e2a9ebfd55820d9f8123b9ac08bebda66c723e49b7826302d73532f9047R54

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProcessCredentialsProvider.kt
lauzadis: It does not work properly, I will need to handle this in a better way

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/CachedValue.kt
```diff
@@ -29,34 +33,58 @@ internal class CachedValue<T> (
...
+
     /**
      * Get the value if it has not expired yet. Returns null if the value has expired
      */
-    suspend fun get(): T? = mu.withLock {
-        if (!isExpiredUnlocked()) return value!!.value else null
+    fun get(): T? {
+        check(!closed.value) { "value is closed" }
+
+        if (ref == null || isExpired(ref!!)) {
```
aajtodd, line 61: **correctness**: This is doing 3 atomic loads of a `ref`. Typically with atomics you'll want to load and store it into a local variable.

e.g.

```kotlin
check(!closed) { ... }
val curr = ref ?: return null
return isExpired(curr)
```

This ensures the value isn't updated from a different thread/coroutine in middle of a single function call.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/CachedValue.kt
```diff
@@ -29,34 +33,58 @@ internal class CachedValue<T> (
...
     /**
-     * Check if the value is expired or not as compared to the time [now]
+     * Check if the value is expired or not as compared to the time now
      */
-    suspend fun isExpired(): Boolean = mu.withLock { isExpiredUnlocked() }
-
-    private fun isExpiredUnlocked(): Boolean {
-        val curr = value ?: return true
-        return clock.now() >= (curr.expiresAt - bufferTime)
+    fun isExpired(): Boolean {
```
aajtodd, line 48: **style**: this can be turned into a property now 


```kotlin
val isExpired: Boolean
    get() { ... }
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/CachedValue.kt
aajtodd: Moreover since this doesn't modify the value should this throw an exception or just return `null`? 🤔 

```kotlin
fun get(): T? {
    val curr = ref
    if (curr == null || !isClosed || isExpired(curr)) return null
    return curr.takeIf { !isClosed && it != null && !isExpired(it) }
}
```


Summon @ianbotsf.

ianbotsf: I would argue that it calling `get` on a closed instance should throw an exception, not return `null`.

```kotlin
fun get(): T? {
    check(!closed.value) { "value is closed" }
    return ref?.takeIf { !isExpired(it) }
}
```

ianbotsf (approved):
Looks good to me!

aws-runtime/aws-config/build.gradle.kts
```diff
@@ -43,6 +43,9 @@ kotlin {
 
                 // additional dependencies required by generated sso provider
                 implementation(project(":aws-runtime:protocols:aws-json-protocols"))
+
+                // atomics
+                implementation("org.jetbrains.kotlinx:atomicfu:$atomicFuVersion")
```
aajtodd, line 48: fix:

This should apply the atomicfu gradle plugin as well (see [example](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/io/build.gradle.kts#L6-L25) from smithy-kotlin)

```kotlin
// right after the copyright (buildscript blocks must come before anything else except imports)
buildscript {
    val atomicFuVersion: String by project
    repositories {
        mavenCentral()
    }

    dependencies {
        classpath("org.jetbrains.kotlinx:atomicfu-gradle-plugin:$atomicFuVersion")
    }
}

...
val coroutinesVersion: String by project
val atomicFuVersion: String by project

apply(plugin = "kotlinx-atomicfu")

...

```

aws-runtime/aws-config/jvm/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommandJVM.kt
```diff
@@ -0,0 +1,72 @@
...
+        val buffer = CharArray(1024)
+
+        withTimeout(timeoutMillis) {
+            while (true) {
+                val rc = reader.read(buffer)
+                if (rc == -1) break
+
+                output.append(buffer)
+                if (output.length > maxOutputLengthBytes) {
+                    throw RuntimeException("Process output exceeded limit of $maxOutputLengthBytes bytes")
```
aajtodd, line 56: **question**: Should this be `ClientException` or `CredentialsProviderException`?

aajtodd (approved):
Fix and ship!

aws-runtime/aws-config/jvm/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommandJVM.kt
lauzadis: I changed it to `CredentialsProviderException` 👍

Issue #711, opened 2022-09-29
Plumb explicit HTTP engine and region through credential provider chain


aajtodd:
### Describe the feature

If a user sets a custom `httpClientEngine` on their service client config we don't currently plumb this through to the credential provider chain.

```kotlin
interface S3Client {
     ...

    public class Config private constructor(builder: Builder): HttpClientConfig, SdkClientConfig {
        public val credentialsProvider: CredentialsProvider = builder.credentialsProvider?.borrow() ?: DefaultChainCredentialsProvider()
        ...
        override val httpClientEngine: HttpClientEngine? = builder.httpClientEngine
     }
      ...
```

This appears to also be true for `region`

### Is your Feature Request related to a problem?

N/A

### Proposed Solution

N/A

### Describe alternative solutions or features you've considered

N/A

### Acknowledge

- [X] I may be able to implement this feature request

### AWS Kotlin SDK version used

N/A

### Platform (JVM/JS/Native)

N/A

### Operating System and version

N/A

lauzadis:
Closed by #730 and [awslabs/smithy-kotlin/721](https://github.com/awslabs/smithy-kotlin/pull/721)

PR #730, opened 2022-10-19
fix(rt): plumb client configuration properties

lauzadis:
Plumb client configuration properties through to other client configuration properties 

## Issue \#
#711 

## Description of changes
When a user specifies an HTTP engine or region in the client configuration, these are not passed through to the CredentialsProvider. This PR passes those configurations to the DefaultCredentialsChainProvider, and also passes `region` from the DefaultCredentialsChainProvider to the ProfileCredentialsProvider.

Note: if a user specifies a custom CredentialsProvider, they will still need to specify the engine both there and in the general client configuration.

---

### DIFF
--- a/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/AwsServiceConfigIntegration.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/AwsServiceConfigIntegration.kt
@@ -25,7 +25,7 @@ class AwsServiceConfigIntegration : KotlinIntegration {
                     AWS region to make requests to
             """.trimIndent()
             propertyType = ClientConfigPropertyType.Required()
-            order = -1
+            order = -100
         }
 
         val CredentialsProviderProp: ClientConfigProperty = ClientConfigProperty {


lucix-aws (approved):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/DefaultChainCredentialsProvider.kt
```diff
@@ -41,14 +41,15 @@ public class DefaultChainCredentialsProvider constructor(
     private val profileName: String? = null,
     private val platformProvider: PlatformProvider = Platform,
     httpClientEngine: HttpClientEngine? = null,
+    region: String? = null,
```
aajtodd, line 44: fix: Needs documentation

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/AwsServiceConfigIntegration.kt
```diff
@@ -25,6 +25,7 @@ class AwsServiceConfigIntegration : KotlinIntegration {
                     AWS region to make requests to
             """.trimIndent()
             propertyType = ClientConfigPropertyType.Required()
+            order = -1
```
aajtodd, line 28: Same comment as `smithy-kotlin`, leave yourself some room between the default to allow for insertion of properties in-between these if necessary.

aajtodd (approved):
Fix and ship.

The codegen diff preview action appears to have a bug in it related to getting the correct version of `smithy-kotlin` to use. Ignore for now, I'll create an issue for it.

PR #748, opened 2022-11-10
refactor(rt): migrate to SdkBuffer

aajtodd:
## Issue \#

N/A
## Description of changes

Tracks upstream changes from https://github.com/awslabs/smithy-kotlin/pull/730

### DIFF
--- a/aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/CrcUtil.kt
+++ b/aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/CrcUtil.kt
@@ -1,0 +1,34 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package aws.sdk.kotlin.runtime.protocol.eventstream
+
+import aws.smithy.kotlin.runtime.hashing.Crc32
+import aws.smithy.kotlin.runtime.io.SdkSink
+import aws.smithy.kotlin.runtime.io.SdkSource
+import aws.smithy.kotlin.runtime.io.internal.SdkSinkObserver
+import aws.smithy.kotlin.runtime.io.internal.SdkSourceObserver
+
+internal class CrcSource(source: SdkSource) : SdkSourceObserver(source) {
+    private val _crc = Crc32()
+
+    val crc: UInt
+        get() = _crc.digestValue()
+
+    override fun observe(data: ByteArray, offset: Int, length: Int) {
+        _crc.update(data, offset, length)
+    }
+}
+
+internal class CrcSink(sink: SdkSink) : SdkSinkObserver(sink) {
+    private val _crc = Crc32()
+
+    val crc: UInt
+        get() = _crc.digestValue()
+
+    override fun observe(data: ByteArray, offset: Int, length: Int) {
+        _crc.update(data, offset, length)
+    }
+}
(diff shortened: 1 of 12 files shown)


aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/Message.kt
```diff
@@ -44,42 +43,43 @@ public data class Message(val headers: List<Header>, val payload: ByteArray) {
 
     public companion object {
         /**
-         * Read a message from [buffer]
+         * Read a message from [source]
          */
-        public fun decode(buffer: Buffer): Message {
-            val totalLen = buffer.readUInt()
+        public fun decode(source: SdkBufferedSource): Message {
+            val totalLen = source.peek().use { it.readInt().toUInt() }
```
ianbotsf, line 49: **Comment**: Reading/writing unsigned variants seems like a good use case for extension methods up in **smithy-kotlin**.

aajtodd: Yeah okio doesn't have unsigned variants for reading/writing. Since this is just reinterpreting the underlying bytes it's not super important to explicitly support unsigned reading/writing but we could move the conversion there if you prefer.

aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/CrcUtil.kt
```diff
@@ -0,0 +1,34 @@
...
+internal class CrcSink(sink: SdkSink) : SdkSinkObserver(sink) {
+    private val _crc = Crc32()
+
+    val crc: UInt
+        get() = _crc.digestValue()
+
+    override fun observe(data: ByteArray, offset: Int, length: Int) {
+        _crc.update(data, offset, length)
+    }
+}
```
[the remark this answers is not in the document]
aajtodd: I agree but I didn't want to upset the runtime relationships with this refactor. okio supports a [HashingSource](https://square.github.io/okio/3.x/okio/okio/okio/-hashing-source/index.html) and [HashingSink](https://square.github.io/okio/3.x/okio/okio/okio/-hashing-sink/index.html) directly. Ideally we would have something analogous (and also something like this CRC source/sink) in our `io` module OR as extensions in our `hashing` module. 

As it stands right now though `hashing` doesn't depend on `io` and `io` doesn't depend on `hashing`. Runtime core does depend on `io` though. We need to untangle all these when we refactor the runtime. I think right now I'm keen on just leaving these here and moving them in a future refactor. Thoughts? How do you think these packages should be structured?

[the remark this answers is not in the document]
ianbotsf: I think I'd want these to live in **io** (and thus **io** would depend on **hashing**) but I agree that it could be left to a broader future refactor.

ianbotsf (approved):

aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/CrcUtil.kt
```diff
@@ -0,0 +1,34 @@
...
+
+package aws.sdk.kotlin.runtime.protocol.eventstream
+
+import aws.smithy.kotlin.runtime.hashing.Crc32
+import aws.smithy.kotlin.runtime.io.SdkSink
+import aws.smithy.kotlin.runtime.io.SdkSource
+import aws.smithy.kotlin.runtime.io.internal.SdkSinkObserver
+import aws.smithy.kotlin.runtime.io.internal.SdkSourceObserver
+
+internal class CrcSource(source: SdkSource) : SdkSourceObserver(source) {
```
lauzadis, line 14: naming: Crc32Source?  considering we have CRC32C now

lauzadis (approved):

aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/CrcUtil.kt
aajtodd: We can _hash_ out naming when we move these and make them re-usable.

lucix-aws (approved):

PR #799, opened 2022-12-29
feat: implement flexible checksums customization

lauzadis:
This PR adds support for the `httpChecksum` trait, otherwise known as flexible checksums.
Users can specify a checksum algorithm to be used in requests and also opt-in to checksum validation in responses.

## Issue \#

https://github.com/awslabs/smithy-kotlin/issues/446

## Description of changes

This change is required to achieve feature parity with other SDKs which already support the `httpChecksum` trait for sending checksums and validating received checksums.

### DIFF
--- a/.changes/07860a56-6515-4545-8c04-4bbabaca7d36.json
+++ b/.changes/07860a56-6515-4545-8c04-4bbabaca7d36.json
@@ -1,5 +1,0 @@
-{
-    "id": "07860a56-6515-4545-8c04-4bbabaca7d36",
-    "type": "misc",
-    "description": "Refactor the way service client configuration is generated"
-}
--- a/.changes/28f9e2e9-1bf0-4bea-be63-847e6ab1af5e.json
+++ b/.changes/28f9e2e9-1bf0-4bea-be63-847e6ab1af5e.json
@@ -1,5 +1,0 @@
-{
-  "id": "28f9e2e9-1bf0-4bea-be63-847e6ab1af5e",
-  "type": "feature",
-  "description": "add ProcessCredentialsProvider which invokes a user-specified command to retrieve credentials"
-}
--- a/.changes/8b4e9549-05e6-445d-b116-9ce34513e8e6.json
+++ b/.changes/8b4e9549-05e6-445d-b116-9ce34513e8e6.json
@@ -1,5 +1,0 @@
-{
-    "id": "8b4e9549-05e6-445d-b116-9ce34513e8e6",
-    "type": "feature",
-    "description": "Allow config override for one or more operations with an existing service client."
-}
--- a/.changes/9ba26a2a-1de0-489f-8d8d-e7002c9c9fa6.json
+++ b/.changes/9ba26a2a-1de0-489f-8d8d-e7002c9c9fa6.json
@@ -1,5 +1,0 @@
-{
-    "id": "9ba26a2a-1de0-489f-8d8d-e7002c9c9fa6",
-    "type": "misc",
-    "description": "Update to latest AWS service models."
-}
--- a/.changes/9c5244b5-bf00-4928-bc1d-20e50f8e6c53.json
+++ b/.changes/9c5244b5-bf00-4928-bc1d-20e50f8e6c53.json
@@ -1,5 +1,0 @@
-{
-    "id": "9c5244b5-bf00-4928-bc1d-20e50f8e6c53",
-    "type": "misc",
-    "description": "**Breaking** Remove `Closeable` supertype from `HttpClientEngine` interface. See [this discussion post](https://github.com/awslabs/aws-sdk-kotlin/discussions/818) for more information."
-}
--- a/.changes/9e689c56-19d5-4e61-b91e-2aade1c47bba.json
+++ b/.changes/9e689c56-19d5-4e61-b91e-2aade1c47bba.json
@@ -1,5 +1,0 @@
-{
-    "id": "9e689c56-19d5-4e61-b91e-2aade1c47bba",
-    "type": "misc",
-    "description": "Upgrade Kotlin version to 1.8.0"
-}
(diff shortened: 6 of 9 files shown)


codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsResponse.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    override fun customizeMiddleware(ctx: ProtocolGenerator.GenerationContext, resolved: List<ProtocolMiddleware>) =
+        resolved + flexibleChecksumsResponseMiddleware
+
+    private val flexibleChecksumsResponseMiddleware = object : ProtocolMiddleware {
+        override val name: String = "FlexibleChecksumsResponse"
+
+        override fun isEnabledFor(ctx: ProtocolGenerator.GenerationContext, op: OperationShape): Boolean {
+            val httpChecksumTrait = op.getTrait<HttpChecksumTrait>()
+            val input = op.input.getOrNull()?.let { ctx.model.expectShape<StructureShape>(it) }
```
aajtodd, line 34: **nit**: We used to write code this way and it's technically fine but all operations have a defined input and output shape FWIW (because we transform the model to make it so).

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsResponse.kt
```diff
@@ -0,0 +1,55 @@
...
+
+            return (httpChecksumTrait != null) &&
+                (httpChecksumTrait.requestValidationModeMember?.getOrNull() != null) &&
+                (input?.memberNames?.any { it == httpChecksumTrait.requestValidationModeMember.get() } == true)
+        }
+
+        override fun render(ctx: ProtocolGenerator.GenerationContext, op: OperationShape, writer: KotlinWriter) {
+            val middlewareSymbol = RuntimeTypes.Http.Middlware.FlexibleChecksumsResponseMiddleware
+            writer.addImport(middlewareSymbol)
```
aajtodd, line 43: **fix**: remove. `#T` will auto import symbols as needed

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsResponse.kt
```diff
@@ -0,0 +1,55 @@
...
+
+            val httpChecksumTrait = op.getTrait<HttpChecksumTrait>()!!
+            val requestValidationModeMember = ctx.model.expectShape<StructureShape>(op.input.get())
+                .members()
+                .first { it.memberName == httpChecksumTrait.requestValidationModeMember.get() }
+
+            writer.withBlock("input.#L?.let {", "}", requestValidationModeMember.defaultName()) {
```
aajtodd, line 50: **correctness**: What is this withBlock actually doing? It looks like it's generating code such as:

```kotlin
input.fooRequestValidationmember?.let {
    op.install(FlexibleChecksumsResponseMiddleware())
}
```

Which isn't actually doing anything with `requestValidationModeMember`

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsRequest.kt
```diff
@@ -0,0 +1,56 @@
...
+            val middlewareSymbol = RuntimeTypes.Http.Middlware.FlexibleChecksumsRequestMiddleware
+            writer.addImport(middlewareSymbol)
+
+            val httpChecksumTrait = op.getTrait<HttpChecksumTrait>()!!
+
+            val checksumAlgorithmMember = ctx.model.expectShape<StructureShape>(op.input.get())
+                .members()
+                .first { it.memberName == httpChecksumTrait.requestAlgorithmMember.get() }
+
+            writer.withBlock("input.#L?.let {", "}", checksumAlgorithmMember.defaultName()) {
```
aajtodd, line 51: **correctness**: Same comment as below. What is this actually doing?

Why do we need the `withBlock`? Can't this just be:

```kotlin
op.install("#T(input.#L.value)", ...)
```

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsResponse.kt
lauzadis: `.defaultName()` will return the value of the `requestValidationModeMember`, so it will generate something like this:

```kotlin
input.checksumAlgorithm?.let {
    op.install(FlexibleChecksumsResponseMiddleware())
}
```

I did it this way to only install the middleware if the input member defined by `requestValidationModeMember` is not `null`. Is there a better way to do this?

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsRequest.kt
lauzadis: `input.#L` may be `null`, and the middleware shouldn't be installed in that case. We can only determine if the input `checksumAlgorithmMember` is `null` during runtime, right?

ianbotsf (approved):

aajtodd (approved):

PR #801, opened 2022-12-29
feat: flexible checksums design

lauzadis:
This PR adds the flexible checksums design doc.

## Description of changes

### DIFF
--- a/docs/design/flexible-checksums.md
+++ b/docs/design/flexible-checksums.md
@@ -34,14 +34,13 @@ If a checksum is required, and the user does not opt-in to using flexible checks
 of injecting the `Content-MD5` header.
 ## Checksum Algorithms
 
-The SDK needs to support the following checksum algorithms: CRC32C, CRC32, SHA-1, SHA-256
-
+The SDK needs to support the following checksum algorithms: CRC32C, CRC32, SHA-1, SHA-256.
 All of them are [already implemented for JVM](https://github.com/awslabs/smithy-kotlin/tree/5773afb348c779b9e4aa9689836844f21a571908/runtime/hashing/jvm/src/aws/smithy/kotlin/runtime/hashing).
 
 As part of this feature, CRC32C was implemented in **smithy-kotlin** [PR#724](https://github.com/awslabs/smithy-kotlin/pull/724). 
 This algorithm is essentially the same as CRC32, but uses a different polynomial under the hood. 
 The SDK uses [`java.util.zip`'s implementation of CRC32](https://docs.oracle.com/javase/8/docs/api/java/util/zip/CRC32.html), 
-but this package only began shipping CRC32C in Java 9. The SDK wants to support Java 8 at a minimum, so this was implemented
+but this package only began shipping CRC32C in Java 9. The SDK requires Java 8, so this was implemented
 rather than imported as a dependency (which is also [what the Java SDK did](https://github.com/aws/aws-sdk-java-v2/blob/ecc12b43a4aedc433c39742a2ae1361bd8d17991/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/checksums/factory/SdkCrc32C.java)).
 
 ### Header Name
@@ -60,11 +59,8 @@ During an HTTP request, the SDK first needs to check if the user has opted-in to
 but the operation has the `requestChecksumRequired` property set, the SDK will fall back to the legacy behavior of computing the MD5 checksum.
 
 ### Middleware
-A new middleware is introduced at the `mutate` stage. This stage of the HTTP request lifecycle represents the last chance
-to modify the request before it is sent on the network.
-
-There are many middleware which operate at the `mutate` stage. It is important that this new middleware come before 
-[`AwsSigningMiddleware`](https://github.com/awslabs/smithy-kotlin/blob/5773afb348c779b9e4aa9689836844f21a571908/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt#L26)
+A new middleware is introduced at the `mutate` stage. There are many middleware which operate at this stage. 
+It is important that this new middleware come before [`AwsSigningMiddleware`](https://github.com/awslabs/smithy-kotlin/blob/5773afb348c779b9e4aa9689836844f21a571908/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt#L26)
 because it is dependent on the header values set in this new middleware (specifically `x-amz-trailer`).
 
 The SDK exposes an `order` integer parameter which is used to model dependencies between middleware.
@@ -135,8 +131,8 @@ An exception will be thrown if the algorithm can't be parsed or if it's not supp
 Note that because users select an algorithm from a code-generated enum, accidentally selecting an unsupported algorithm is unlikely.
 
 ### Computing and Injecting Checksums
-Next the SDK will compute and inject the checksum. If the body is smaller than the `aws-chunked` threshold ([1MB today](https://github.com/awslabs/smithy-kotlin/blob/9b9297c690d9a01777447f437f0e91562e146bf9/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt#L38)), 
-the checksum will be immediately computed and injected under the appropriate header name.
+Next the SDK will compute and inject the checksum. If the body is smaller than the `aws-chunked` threshold ([1MB today](https://github.com/awslabs/smithy-kotlin/blob/9b9297c690d9a01777447f437f0e91562e146bf9/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt#L38)) 
+and replayable, the checksum will be immediately computed and injected under the appropriate header name.
 
 Otherwise, if the request body is large enough to be uploaded with `aws-chunked`, the SDK will append the checksum header name to the `x-amz-trailer` header.
 
@@ -191,23 +187,24 @@ AwsChunkedReader(
 )
 ```
 
-#### CompletingBody
+#### Completing Sources and Channels
 
 When using `CompletableDeferred`, the `.complete()` method must be called to mark the deferred value as complete.
 
-Because the checksum computation [is done while the request body is being sent](#computing-and-injecting-checksums),
-a new type of `Source`/`ByteReadChannel` is introduced, called `CompletingSource` and `CompletingByteReadChannel` respectively
+Because the checksum computation [is done while the request body is being sent](#computing-and-injecting-checksums), 
+new types of `Source`/`ByteReadChannel` are introduced, called `CompletingSource` and `CompletingByteReadChannel` respectively.
 
 Below is the signature of `CompletingSource`:
 ```kotlin
 CompletingSource(
-  private val deferredChecksum: CompletableDeferred<String>,
-  private val source: HashingSource
+    private val deferredChecksum: CompletableDeferred<String>,
+    private val source: HashingSource
 )
 ```
 
 This will be used to wrap the `HashingSource`. When the source is fully exhausted, 
-the calculated checksum will be digested and used to `.complete()` the `completableDeferred`. The same will be done for HashingByteReadChannels using a `CompletingByteReadChannel`.
+the calculated checksum will be digested and used to `.complete()` the `CompletableDeferred`. 
+The same will be done for `HashingByteReadChannel` using a `CompletingByteReadChannel`.
 
 #### Mutating Trailing Headers
 
@@ -345,24 +342,23 @@ Instead of `CompletableFuture`, `LazyAsyncValue` and `RunnableFuture` were evalu
 ### `LazyAsyncValue`
 
 Pros:
-- already exists in the SDK
-- lazy execution
-  - the checksum future can be computed only when it is ready / needed
+- Already exists in the SDK
+- Lazy execution
+  - The checksum future can be computed only when it is ready / needed
 
 Cons:
-- no concept of "completion"
-  - there is no way to indicate that the lazy value is ready for consumption. the underlying code block will be executed 
+- No concept of "completion"
+  - There is no way to indicate that the lazy value is ready for consumption. The underlying code block will be executed 
 when `.get()` is called, which could happen earlier than intended
   
 
 ### `RunnableFuture`
 Pros:
-- execution can be delayed until the result is actually needed
-  - fixes the issue with eager execution in `CompletableFuture`
+- Execution can be delayed until the result is actually needed
+  - Fixes the issue with eager execution in `CompletableFuture`
 Cons:
-- there is no concept of completion
-  - calling `.get()` on a `Future` / `RunnableFuture` will block until it's complete. 
-it can throw exceptions if the computation was cancelled, threw an exception, or if the thread was interrupted while blocking
+- There is no concept of completion
+  - Calling `.get()` on a `Future` / `RunnableFuture` will block until it's complete
 
 Ultimately, `CompletableFuture` was chosen because it provides a way to model completion and the value can be set in a non-blocking manner.
 


aajtodd (changes requested):
This is a great start, couple comments/fixes suggested.

-----

On the preferred choice of `LazyAsyncValue` I'm not sure I agree. In part because `LazyAsyncValue` isn't really intended for this use case. It's just a `suspend` equivalent of [lazy](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/lazy.html). The main use case of `LazyAsyncValue` was/is to defer computing a value that is produced by some suspending computation. I can see where this definition _almost_ fits this use case. 

The crux of my issue is that `LazyAsyncValue` places absolutely no timing constraints on the consumer. You can call it whenever you want and it should execute and produce a result. The AWS chunked implementation cannot invoke it until the entire body is read. This works for this use case because the trailer is related to the body consumption but what if it wasn't? I think a better alternative here is to use [Deferred](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-deferred/) (which can be created by the producer with [CompletableDeferred](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-completable-deferred/)). A `Deferred` value has the advantage that it decouples the producer and consumer timing. A consumer can immediately invoke `await` and it won't complete until the producer sets a value for it. The other advantage is that deferred values support cancellation. 

I'm not sure if this is the alternative that was rejected, it wasn't clear but this would be my preferred choice for representing trailer header values. We should discuss as a group probably.

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+        "md5" -> Md5()
+        else -> throw RuntimeException("$this is not a supported hash function")
+    }
+}
+```
+Note that MD5 is included here, but it is not a supported flexible checksum algorithm.
+
+There is a secondary validation to ensure that the user-specified algorithm is allowed to be used in flexible checksums:
+```kotlin
+private val HashFunction.isSupported: Boolean get() = when (this) {
```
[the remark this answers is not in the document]
lauzadis: I changed the implementation to return null instead of throwing an exception -- I will update the doc

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+val getObjectRequest = GetObjectRequest {
+    bucket = "bucket"
+    key = "key"
+    checksumMode = ChecksumMode.Enabled
+}
+```
+
+## Alternative Designs Considered
+
+### Lazy Headers
```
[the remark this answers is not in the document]
lauzadis: Ok, that sounds better. I chose the name "Lazy Headers" to refer back to the original section where the design choice was introduced

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+# Design
+
+This feature can be broken up into two new middleware -- one for calculating checksums for requests, and one for
+validating checksums present in responses.
+
+# Requests
+
+## Overview
+During an HTTP request, the SDK first needs to check if the user has opted-in to sending checksums. If they have not opted-in,
+but the operation has the `requestChecksumRequired` property set, the SDK will fall back to the legacy behavior of computing the MD5 checksum.
```
ianbotsf, line 64: **Style**: Sections without content besides other sections feel unnecessarily over-structured. In this case, the **Overview** subsection seems like the perfect introduction for the **Requests** section. I'd suggest just removing `## Overview`.

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+The header name used to set the checksum value is `x-amz-checksum-<checksum_algorithm_name>`. For example, if the checksum was computed 
+using SHA-256, the header containing the checksum will be `x-amz-checksum-sha256`.
+
+# Design
+
+This feature can be broken up into two new middleware -- one for calculating checksums for requests, and one for
+validating checksums present in responses.
+
+# Requests
+
```
ianbotsf, line 61: **Comment**: It feels like **Requests** and **Responses** should be subsections of ~~**Design**~~ **Implementation**.

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+## Checksum Header Name
+
+The header name used to set the checksum value is `x-amz-checksum-<checksum_algorithm_name>`. For example, if the checksum was computed 
+using SHA-256, the header containing the checksum will be `x-amz-checksum-sha256`.
+
+# Design
+
+This feature can be broken up into two new middleware -- one for calculating checksums for requests, and one for
+validating checksums present in responses.
```
ianbotsf, line 58: **Nit**: This entire document is a "design". This section should probably be titled something like **Implementation** or **Specification**.

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+## Middleware
+A new middleware is introduced at the `mutate` stage. This stage of the HTTP request lifecycle represents the last chance
+to modify the request before it is sent on the network.
+
+There are many middleware which operate at the `mutate` stage. It is important that this new middleware come before 
+[AwsSigningMiddleware](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt#L26)
+because that middleware is dependent on the header values set in this new middleware (specifically `x-amz-trailer`).
+
+The SDK exposes an `order` integer parameter which is used to model dependencies between middleware.
+The `order` of AwsSigningMiddleware has already been set to 126, which ensures it will be executed towards the end of the mutate middleware stack,
```
ianbotsf, line 75: **Nit**: AwsSigningMiddleware → `AwsSigningMiddleware`

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+### Normal Requests
+For all normal requests, the checksum should be injected into the header.
+
+### Streaming Requests
+For streaming requests which are either streaming-signing or unsigned, the checksum must be sent as a trailing header via `aws-chunked` encoding.
+
+To indicate that a trailing header will be sent, the SDK sets the `x-amz-trailer` header to a string of comma-delimited trailing header names.
+The service uses this header to parse the trailing headers that are sent later.
+
+For flexible checksums, we append the [checksum header name](#checksum-header-name) to the `x-amz-trailer` header.
```
ianbotsf, line 100: **Style**: Avoid "we" in technical writing. It might be clearer here to say that "the checksum header name _will be appended_ to the `x-amz-trailer` header".

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+
+## Input Checksum
+The user may pre-calculate the checksum and provide it as input. The SDK automatically parses this input
+and adds it to the request headers. When this header is present, the rest of the flexible checksums request workflow is skipped.
```
ianbotsf, line 104: **Style**: Elsewhere in this document (and generally) we refer to "requests" rather than "input". For consistency, I recommend following that convention in this section too.

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+
+Note: the user must still specify the `ChecksumAlgorithm` even if the checksum itself is supplied as input.
+If the input checksum and checksum algorithm do not match, the input checksum will be ignored and the checksum will be calculated
+internally.
+
+## Validating Input Algorithms
+
+When a user sets the `requestAlgorithmMember` property, they are choosing to opt-in to sending request checksums.
```
ianbotsf: **Question**: I'm confused—`requestAlgorithmMember` is a property of a trait in a Smithy model and I do not see how users can "set" this value directly. Does this actually mean "when a user sets a value in the member identified by the `requestAlgorithmMember` property"?

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+> Content-Encoding: aws-chunked
+> Content-Length: 1238
+> ...
+>
+> 400;chunk-signature=<chunk-signature> + \r\n + [1024 bytes of payload data] + \r\n
+> 0;chunk-signature=<signature> + \r\n
+> x-amz-checksum-crc32c:AAAAAA== + \r\n
+> x-amz-trailer-signature:<trailer-signature> + \r\n
+> \r\n
+```
```
ianbotsf, line 162: **Style**: These `>` prefixes are unnecessary given this is already a preformatted block.

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+Note that users select an algorithm from a code-generated enum, so accidentally providing an unsupported algorithm is unlikely.
+
+
+## Computing and Injecting Checksums
+Next the SDK will compute and inject the checksum. If the body is smaller than the aws-chunked threshold (1MB today), 
+the checksum will be immediately computed and injected under the appropriate header name.
+
+## aws-chunked
+Otherwise, if the request body is large enough to be uploaded with `aws-chunked`, the SDK will set the `x-amz-trailer` header
+to the checksum header name.
```
ianbotsf, line 146: **Style**: I believe `aws-chunked` is not the proper name of a feature (which is probably "[Chunked Upload](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html)") but rather a code literal used by systems that send/receive chunked streams. If so, we should probably monospace every mention of `aws-chunked` or use a proper name in its place.

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+    private val signingConfig: AwsSigningConfig,
+    private var previousSignature: ByteArray,
+    private var trailingHeaders: Headers = Headers.Empty
+)
+```
+Note that the trailing headers are provided as a `Headers` object, which is essentially a key-value map of header names to their values.
+
+After sending the body, the checksum needs to be sent as a trailing header.
+
+However, it is wise to avoid tight coupling of the aws-chunked and flexible checksums features. The aws-chunked body should have
```
ianbotsf: **Style**: wise → desirable

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+no knowledge of the HashingSource/HashingByteReadChannel it is reading from. 
+
+So, how will the value of the checksum be passed to the AwsChunked body?
+
+### Lazy Headers (*)
+A concept of deferred, or "lazy" header values is introduced. At initialization, the aws-chunked body needs to know that
+a trailing header *will be sent*, but the value can't be ready until the body has been fully consumed.
+
+LazyAsyncValue is the SDK's pre-existing wrapper around Kotlin's [Lazy type](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-lazy/)
```
ianbotsf: **Nit**: LazyAsyncValue → `LazyAsyncValue`

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+
+To calculate the checksum while the payload is being written, the body will be wrapped in either a HashingSource
+or a HashingByteReadChannel, depending on its type.
+
+These constructs will use the provided checksum algorithm to compute the checksum as the data is being read.
+
+Further down the middleware chain, this hashing body will be wrapped once more in an aws-chunked body. This body is used to format the 
+underlying data source into aws-chunked content encoding.
+
+Today the aws-chunked reader has the following signature:
```
ianbotsf: **Comment**: Our designs generally describe the system as it has been implemented after the design is completed. Using words like "today" to describe the prior state of the system or the now-deprecated design is confusing. It might be better to say "The previous implementation of the `aws-chunked` reader had the following signature" or "A naïve structure for the reader might be"...anything to indicate that what follows is _not_ current.

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+)
+```
+Note that the trailing headers are provided as a `Headers` object, which is essentially a key-value map of header names to their values.
+
+After sending the body, the checksum needs to be sent as a trailing header.
+
+However, it is wise to avoid tight coupling of the aws-chunked and flexible checksums features. The aws-chunked body should have
+no knowledge of the HashingSource/HashingByteReadChannel it is reading from. 
+
+So, how will the value of the checksum be passed to the AwsChunked body?
```
ianbotsf: **Nit**: AwsChunked vs aws-chunked

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+LazyAsyncValue is the SDK's pre-existing wrapper around Kotlin's [Lazy type](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-lazy/) 
+that allows asynchronous initialization.
+
+After calling `get()` on these header values, the underlying block is executed and a value is returned.
+
+Since the AwsChunked body will only call `get()` on the trailing headers after the body has been sent, the checksum will already
+be computed and ready to retrieve from the LazyAsyncValue.
+
+The aws-chunked trailing headers implementation is refactored to use the new `LazyHeaders` class, 
+which maps String -> List<LazyAsyncValue&lt;String>>
```
ianbotsf: **Nit**: String -> List<LazyAsyncValue&lt;String>> → `String -> List<LazyAsyncValue<String>>`

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+
+
+## Middleware
+
+To run this validation process, a new middleware is inserted at the `receive` stage. During an HTTP request lifecycle,
+this stage represents the first opportunity to access the response prior to deserialization into the operation's Response type.
+
+### Deferred Checksum Validation (*)
+
+For efficiency, it is best for the SDK to compute the checksum as the user is consuming the response body.
```
ianbotsf, line 250: **Question**: Does this only apply to streaming responses? I assume so since the caller does not "consume" non-streaming responses...they are returned inline and fully complete. Do we need additional description of what happens for non-streaming responses?

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+- checksum header name: the name of the checksum header to be validated, which allows the user to see if validation 
+occurred and which checksum algorithm was used.
+
+### Notifying the User of Validation
+
+In some cases, a service will not return a checksum even if it is requested. Because of this, the SDK must provide a mechanism for users to verify whether checksum validation occurred, 
+and which checksum algorithm was used for the validation.
+
+The SDK will store the checksum header name in the execution context. Users can then check the execution context for that
+variable, and if it's present, will know that validation occurred.
```
ianbotsf, line 266: **Nit**: variable → value

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+body in a hashing body, in a similar manner to the [request middleware](#aws-chunked). The execution context will be updated
+with:
+
+- expected checksum: the checksum value from response headers
+- calculated checksum: a LazyAsyncValue containing the calculated checksum, which will only be fetched after the entire response body is consumed
+- checksum header name: the name of the checksum header to be validated, which allows the user to see if validation 
+occurred and which checksum algorithm was used.
+
+### Notifying the User of Validation
+
```
ianbotsf, line 261: **Question**: Neither this section nor its surrounding sections discuss what happens when validation yields a mismatch. Will we throw an exception? Leave it up to the user to compare expected and calculated values? Other?

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+The SDK will store the checksum header name in the execution context. Users can then check the execution context for that
+variable, and if it's present, will know that validation occurred.
+
+Today the SDK does not provide a mechanism for users to observe the execution context, but that will be added as part of Interceptors. 
+
+# Appendix
+
+## Request Examples
+
```
ianbotsf, line 273: **Nit**: The following subsections are "appendices"—a collection in which each item is an "appendix". I suggest either renaming this section to `# Appendices` –or– removing this `# Appendix` section and updating the following sections to `# Appendix: Request Examples`, `# Appendix: Response Examples`, etc.

ianbotsf, line 273: **Comment**: It may be helpful to note in these examples that `requestAlgorithmMember` is `checksumAlgorithm` and to describe how the names `checksumSha256`, `checksumCrc32`, etc. are determined.

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+## Alternative Designs Considered
+
+### Lazy Headers
+
+Instead of LazyAsyncValue, CompletableFuture / Future were evaluated for use as a deferred header value.
+
+LazyAsyncValue was ultimately chosen because:
+- it is already used in other similar places by the SDK
+  - it's better to reuse something that already exists
+- CompletableFuture uses coroutines, which adds unnecessary cognitive load for developers
```
ianbotsf: **Correctness**: `CompleteableFuture` _does not_ use coroutines...it uses Java-native concurrency primitives.

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+calculation was being done in a blocking manner in the middleware.
+The response body was being hashed and validated prior to passing the response on to the user.
+
+There are pros and cons to this design choice, but ultimately it was decided against blocking to calculate the checksum.
+
+Pros:
+- The checksum is calculated prior to passing the response to the user
+  - With the accepted design choice, if the checksum is invalid, the user will only know about it after they've consumed
+    the whole response body. If they are sending the response body downstream, for example, they would then have to claw it
+    back since the data is not valid
```
ianbotsf, line 377: **Style**: One cannot "claw back" a sent stream. It's probably more accurate to say a user may have to cancel, invalidate, or otherwise handle a downstream operation.

ianbotsf (changes requested):
Please also add a link to the new document in [docs/design/README.md](https://github.com/awslabs/aws-sdk-kotlin/blob/main/docs/design/README.md).

docs/design/flexible-checksums.md
lauzadis: Good question, I did not specify it here in the design. An exception will be thrown on checksum mismatch.

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,356 @@
...
+### Deprecating `httpChecksumRequired`
+
+Before flexible checksums, services used the `httpChecksumRequired` trait to require a checksum in the request. 
+This is computed using the MD5 algorithm and injected in the request under the `Content-MD5` header. 
+
+This `httpChecksumRequired` trait is now deprecated. Services should use the `httpChecksum` trait's
+`requestChecksumRequired` property instead.
+
+If the `requestChecksumRequired` property is set to `true`, **and** the user opts-in to using flexible checksums,
+the SDK must use flexible checksums. Otherwise, if a request requires a checksum but
```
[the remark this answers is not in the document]
lauzadis: Yes, it is. I can clarify that here

docs/design/flexible-checksums.md
lauzadis: Yes, and that's something I'll need check in my implementation, because I'm not sure what would happen with a non-streaming response.

ianbotsf (commented):
**Question**: Most of this design now seems to be [implemented in **smithy-kotlin**](https://github.com/awslabs/smithy-kotlin/pull/772). How much of this design is specific to **aws-sdk-kotlin**? Should the design be moved to the other repository?

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,428 @@
...
+  
+
+### `RunnableFuture`
+Pros:
+- execution can be delayed until the result is actually needed
+  - fixes the issue with eager execution in `CompletableFuture`
+Cons:
+- there is no concept of completion
+  - calling `.get()` on a `Future` / `RunnableFuture` will block until it's complete. 
+it can throw exceptions if the computation was cancelled, threw an exception, or if the thread was interrupted while blocking
```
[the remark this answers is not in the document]
lauzadis: This is actually a part of the bulleted list, not its own sentence. I just broke it out onto a newline to prevent the Markdown line being too long, though I could capitalize every member of the bulleted lists if that seems better.

lauzadis:
> **Question**: Most of this design now seems to be [implemented in **smithy-kotlin**](https://github.com/awslabs/smithy-kotlin/pull/772). How much of this design is specific to **aws-sdk-kotlin**? Should the design be moved to the other repository?

Sure, if that's the heuristic the team uses to determine locating designs in **aws-sdk-kotlin** vs. **smithy-kotlin**. I thought it would be better here in **aws-sdk-kotlin** since the customization is specific to AWS services.

For what it's worth,`httpChecksum` is [documented](https://smithy.io/2.0/aws/aws-core.html#aws-protocols-httpchecksum-trait) as an "AWS-specific trait"

docs/design/flexible-checksums.md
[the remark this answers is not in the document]
ianbotsf: Is this meant to be a top-level bullet point in the list? If so, it's missing a leading `- `. It's currently rendering as part of the preceding line.

aajtodd:
> 

I think previously anything AWS we were putting in this repo but I think we'll eventually move AWS protocols (including trait support) into `smithy-kotlin` as a separate package. This repo will eventually (hopefully) just contain customization's required for AWS SDK Kotlin (and perhaps a small runtime component, e.g. `aws-config` is pretty specific to AWS SDK's). I could make an argument for both but being forward looking, probably relocate to `smithy-kotlin`, perhaps in a new "aws" subfolder of design or something. We can discuss as a team.

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,428 @@
...
+This feature can be implemented by adding two new middleware: one for calculating checksums for requests, and one for
+validating checksums present in responses.
+
+## Requests
+
+During an HTTP request, the SDK first needs to check if the user has opted-in to sending checksums. If they have not opted-in,
+but the operation has the `requestChecksumRequired` property set, the SDK will fall back to the legacy behavior of computing the MD5 checksum.
+
+### Middleware
+A new middleware is introduced at the `mutate` stage. This stage of the HTTP request lifecycle represents the last chance
```
aajtodd, line 63: nit: `mutate` is not the last chance to modify the request before sent out on the network (that would be `receive`).

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,428 @@
...
+    else -> false
+}
+```
+
+An exception will be thrown if the algorithm can't be parsed or if it's not supported for flexible checksums. 
+Note that because users select an algorithm from a code-generated enum, accidentally selecting an unsupported algorithm is unlikely.
+
+### Computing and Injecting Checksums
+Next the SDK will compute and inject the checksum. If the body is smaller than the `aws-chunked` threshold ([1MB today](https://github.com/awslabs/smithy-kotlin/blob/9b9297c690d9a01777447f437f0e91562e146bf9/runtime/auth/aws-signing-common/common/src/aws/smithy/kotlin/runtime/auth/awssigning/middleware/AwsSigningMiddleware.kt#L38)), 
+the checksum will be immediately computed and injected under the appropriate header name.
```
aajtodd, line 139: question: I'm assuming this is only true for replayable streams/bodies?

docs/design/flexible-checksums.md
```diff
@@ -0,0 +1,428 @@
...
+### Middleware
+
+To run this validation process, a new middleware is inserted at the `receive` stage. During an HTTP request lifecycle,
+this stage represents the first opportunity to access the response prior to deserialization into the operation's response type.
+
+If the request has a non-streaming response, the middleware will compute and validate the response checksum in a blocking
+manner. If there is a checksum mismatch, an exception will be thrown.
+
+#### Checksum Validation
+For non-streaming responses, the entire response is loaded into memory at once. The SDK will validate the checksum prior to deserializing
```
aajtodd, line 258: question: Is this some flag passed to the middleware when constructed based on the model? The middleware otherwise will not be able to tell the difference between streaming and non streaming responses (all HTTP response bodies are basically "streaming"). 

Also keep in mind reading HTTP response bodies can only be done once, you'll have to copy the response into a new (e.g. `HttpBody.Bytes`) instance so that the deserializer can read it "again".

aajtodd (approved):
No blocking comments other than what Ian has already called out.

docs/design/flexible-checksums.md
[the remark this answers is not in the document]
lauzadis: I was intending for it to be part of the preceding line, but in hindsight it's not a useful bit of information so I'll remove it

docs/design/flexible-checksums.md
lauzadis: If that's the case, then yeah I will probably need to pass a flag based on the model. Based on the current feedback on the implementation PR I will already need to refactor this validation middleware, and I'll make sure it works for both streaming and non-streaming responses.

lauzadis:
Moving to **smithy-kotlin** repo: https://github.com/awslabs/smithy-kotlin/pull/788

PR #811, opened 2023-01-11
implement per-op config downstream changes

lucix-aws:
## Description of changes
* Add `ManagedCredentialsProvider` to handle share count tracking for defaulted providers in config.

### DIFF
--- a/.changes/9c5244b5-bf00-4928-bc1d-20e50f8e6c53.json
+++ b/.changes/9c5244b5-bf00-4928-bc1d-20e50f8e6c53.json
@@ -1,0 +1,5 @@
+{
+    "id": "9c5244b5-bf00-4928-bc1d-20e50f8e6c53",
+    "type": "misc",
+    "description": "**Breaking** Remove `Closeable` supertype from `HttpClientEngine` interface. See [this discussion post](https://github.com/awslabs/aws-sdk-kotlin/discussions/818) for more information."
+}
--- a/gradle.properties
+++ b/gradle.properties
@@ -13,7 +13,7 @@ sdkVersion=0.19.6-SNAPSHOT
 smithyVersion=1.26.1
 smithyGradleVersion=0.6.0
 # smithy-kotlin codegen and runtime are versioned together
-smithyKotlinVersion=0.14.3
+smithyKotlinVersion=0.14.4-SNAPSHOT
 
 # kotlin
 kotlinVersion=1.7.10


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
```diff
@@ -98,7 +98,9 @@ public class EcsCredentialsProvider internal constructor(
                 else -> CredentialsProviderException("Failed to get credentials from container metadata service", ex)
             }
         } finally {
-            client.close()
+            if (manageEngine) {
+                client.engine.close()
+            }
```
ianbotsf, line 103: **Question**: Why are we closing the engine after retrieving credentials? Might we have to retrieve credentials again in the future and won't that fail on a closed engine?

Same question for `ImdsClient`.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/core/AwsHttpProtocolClientGenerator.kt
```diff
@@ -37,6 +37,13 @@ open class AwsHttpProtocolClientGenerator(
         super.render(writer)
     }
 
+    override fun renderInit(writer: KotlinWriter) {
+        writer.withBlock("init {", "}") {
+            write("managedResources.#T(config.httpClientEngine)", RuntimeTypes.IO.addIfManaged)
+            write("managedResources.#T(config.credentialsProvider)", RuntimeTypes.IO.addIfManaged)
+        }
+    }
+
```
ianbotsf, line 46: **Suggestion**: This feels like unnecessary duplication with `HttpProtocolClientGenerator.renderInit`. Consider a new function for subclasses to add additional things to the initializer or using sections.

ianbotsf (approved):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
aajtodd: Ya this is wrong, previously we were passing `manageEngine = false` so the underlying engine wasn't closed when `SdkHttpClient` was closed. It seems like `SdkHttpClient` doesn't need `Closeable` anymore and doesn't need to be closed.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/imds/ImdsClient.kt
```diff
@@ -64,6 +64,7 @@ public class ImdsClient private constructor(builder: Builder) : InstanceMetadata
     private val platformProvider: PlatformProvider = builder.platformProvider
     private val sdkLogMode: SdkLogMode = builder.sdkLogMode
     private val httpClient: SdkHttpClient
+    private val manageEngine: Boolean = builder.engine == null
```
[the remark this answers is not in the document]
lucix-aws: Talked offline, and we want to strive for this in the future, but given how the HTTP client is spanned across several providers in the default chain (and then often passed a few levels down within the chain providers) there's a risk of getting the bookkeeping wrong somewhere in that environment, and it'll be safer to land/vet this separately.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/internal/ManagedCredentialsProvider.kt
```diff
@@ -0,0 +1,20 @@
...
+
+private class ManagedCredentialsProvider(
+    private val delegate: CredentialsProvider,
+) : SdkManagedCloseable(delegate), CredentialsProvider by delegate
+
+/**
+ * Wraps a [CredentialsProvider] for internal runtime management by the SDK.
+ */
+@InternalSdkApi
+public fun CredentialsProvider.manage(): CredentialsProvider =
```
aajtodd, line 19: **correctness**: I think this probably actually belongs in `smithy-kotlin` in same package as `CredentialsProvider`.

aajtodd (approved):

ianbotsf (approved):

PR #813, opened 2023-01-12
chore: upgrade dependencies

lauzadis:
Upgrade third party dependencies to their latest verisons
## Issue \#

N/A
## Description of changes

The Kotlin SDK continues to aggressively take the latest versions of dependencies while we are in developer preview.

### DIFF
--- a/aws-runtime/aws-config/jvm/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommandJVM.kt
+++ b/aws-runtime/aws-config/jvm/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommandJVM.kt
@@ -63,13 +63,11 @@ internal actual suspend fun executeCommand(
         }
         reader.close()
 
-        Pair(
-            process.exitValue(),
-            if (process.exitValue() == 0) {
-                output.toString()
-            } else {
-                process.errorStream.bufferedReader().use { it.readText() }
-            },
-        )
+        val commandOutput = when {
+            process.exitValue() == 0 -> output.toString()
+            else -> process.errorStream.bufferedReader().use { it.readText() }
+        }
+
+        process.exitValue() to commandOutput
     }
 }


aws-runtime/aws-config/jvm/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommandJVM.kt
```diff
@@ -65,8 +65,11 @@ internal actual suspend fun executeCommand(
...
         Pair(
             process.exitValue(),
-            if (process.exitValue() == 0) output.toString()
-            else process.errorStream.bufferedReader().use { it.readText() },
+            if (process.exitValue() == 0) {
+                output.toString()
+            } else {
+                process.errorStream.bufferedReader().use { it.readText() }
+            },
         )
```
ianbotsf, line 73: **Comment**: Multiline `if`-`else` block is a bit much for an inline argument. Consider a temp val outside the `Pair` call. Bonus: then you can eliminate the explicit `Pair` construction and just use [`to`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/to.html) instead:

```kotlin
val processOutput = when {
    process.exitValue() == 0 -> output.toString()
    else -> process.errorStream.bufferedReader().use { it.readText() }
}

process.exitValue() to processOutput // Pair(process.exitValue(), processOutput)
```

aws-runtime/aws-config/jvm/test/aws/sdk/kotlin/runtime/config/profile/AWSConfigLoaderFilesystemTest.kt
```diff
@@ -32,13 +32,13 @@ class AWSConfigLoaderFilesystemTest {
 
     @TempDir
     @JvmField
-    val tempDir: Path? = null
+    var tempDir: Path? = null
```
ianbotsf, line 35: **Question**: Why is this change necessary?

ianbotsf (approved):

aws-runtime/aws-config/jvm/test/aws/sdk/kotlin/runtime/config/profile/AWSConfigLoaderFilesystemTest.kt
lauzadis: In the upgraded version of JUnit, `@TempDir` fields cannot be `final`: https://github.com/junit-team/junit5/issues/2839

ianbotsf: That's too bad. In that case, just remove the `requireNotNull(tempDir)` checks in the unit tests. They're no longer doing anything meaningful now that the field is a `var`.

Issue #836, opened 2023-02-08
Eventstream HeaderValue "Unknown HeaderType: X"


ben-zip:
### Describe the bug

When performing SelectObjectContent operation, there is an intermittent error coming from the `HeaderValue` enum class.
Message: `Unknown HeaderType: X`

Where X can be any value (92, 55, 82, 105, 121 etc...)

The call comes from the `deserializeSelectObjectContentOperationBody` function.

Error stack
```
at aws.sdk.kotlin.runtime.protocol.eventstream.HeaderType$Companion.fromTypeId(HeaderValue.kt:33)
  at aws.sdk.kotlin.runtime.protocol.eventstream.HeaderValue$Companion.decode(HeaderValue.kt:119)
  at aws.sdk.kotlin.runtime.protocol.eventstream.Header$Companion.decode(Header.kt:43)
  at aws.sdk.kotlin.runtime.protocol.eventstream.Message$Companion.decode(Message.kt:76)
  at aws.sdk.kotlin.runtime.protocol.eventstream.FrameDecoderKt$decodeFrames$2.invokeSuspend(FrameDecoder.kt:37)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
  at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
  at kotlinx.coroutines.internal.LimitedDispatcher.run(LimitedDispatcher.kt:42)
  at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:95)
  at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:570)
```

### Expected behavior

The  `SelectObjectContent` operation should succeed without error.

### Current behavior

The `fromTypeId` function of the `HeaderValue` enum throws an Exception "Unknown HeaderType: X"
```
        fun fromTypeId(value: Byte): HeaderType =
            requireNotNull(values().find { it.value == value }) { "Unknown HeaderType: $value" }
```

The header type value can be anything 34, 83, 92, 105 etc... and the error is intermittent, trying the operation again can be successful.

### Steps to Reproduce

- Do a `selectObjectContent` on Parquet files (cross-account) with the following request:
```
val r = SelectObjectContentRequest {
        expectedBucketOwner = "ownerAccountId"
        bucket = "bucket"
        key = "key"
        expression = "SELECT * FROM s3object s"
[...]

PR #840, opened 2023-02-08
fix: reading too few bytes when parsing headers in event streams

lauzadis:
This updates the HeaderValue parsing logic to read the full `len` number of bytes. Before this change, the `source` would occasionally read fewer bytes than requested. This would result in a few extra bytes in the stream, and next time the header value is parsed, it would result in an exception claiming `Unknown HeaderType`.

## Issue \#

https://github.com/awslabs/aws-sdk-kotlin/issues/836
## Description of changes

This change fixes the bug described in the issue above.

### DIFF
--- a/aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/Message.kt
+++ b/aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/Message.kt
@@ -58,7 +58,7 @@ public data class Message(val headers: List<Header>, val payload: ByteArray) {
             val messageBuffer = SdkBuffer()
             val computedCrc = run {
                 val crcSource = HashingSource(Crc32(), source)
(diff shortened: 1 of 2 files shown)


aws-runtime/protocols/aws-event-stream/common/src/aws/sdk/kotlin/runtime/protocol/eventstream/HeaderValue.kt
```diff
@@ -128,7 +128,10 @@ public sealed class HeaderValue {
                     val len = source.readShort().toUShort()
                     check(source.request(len.toLong())) { "Invalid HeaderValue; type=$type, len=$len; readRemaining: ${source.buffer.size}" }
                     val bytes = ByteArray(len.toInt())
-                    source.read(bytes)
+                    var rc = 0
```
aajtodd, line 131: I think this loop/allocation could be simplified to just:

```kotlin
val bytes = source.readByteArray(len.toLong())
```

aajtodd (approved):

ianbotsf (approved):

Issue #206, opened 2021-05-12
Support S3 Config File Options


aajtodd:
https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-s3

In particular the common service specific ones like s3 need propagated to the client.

See the `configuration-file` SEP

PR #854, opened 2023-02-17
feature: add sub-property support to config parser

lucix-aws:
In preparation for #206, add support for sub-properties in config.

## Description of changes
* Add support for sub-property continuations, e.g.:
```
[default]
foo = bar ; top-level property
baz = ; parent
  qux1 = 1 ; sub-property. note this "comment" is actually part of the value per configuration-file SEP
  qux2 = 2
```
* Refactor parsing pipeline that transforms input -> config profile map. Workflow is now:
`tokenize (handle all explicit parsing failures) -> build profile mapping (handle and log all dropping of valid-but-malformed values) -> merge`
* Continuations and sub-properties are now explicitly tokenized and handled in the profile mapping, rather than being pre-folded before tokenization
* Tokenization is now context-aware and generally more rigorous, and will explicitly reject otherwise-valid tokens that are incorrectly positioned, rather than deferring and requiring later handling of what should be an explicit fail-fast scenario
* Organize parser components within package

Plumbing for s3-specific properties will be implemented as a follow-on PR.

### DIFF
--- a/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
+++ b/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
@@ -46,7 +46,7 @@ public abstract class AbstractAwsSdkClientFactory<
             defaultTracer(builder.config.clientName)
         }
 
-        coroutineContext.withRootTraceSpan(tracer.createRootSpan("From environment")) {
+        coroutineContext.withRootTraceSpan(tracer.createRootSpan("Config resolution")) {
             builder.config.region = builder.config.region ?: resolveRegion()
             builder.config.retryStrategy = builder.config.retryStrategy ?: resolveRetryStrategy()
         }


aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/config/profile/SpecTestSuites.kt
```diff
@@ -142,20 +142,6 @@ internal const val parserTestSuiteJson = """
         }
       }
     },
-    {
-      "name": "Profiles can contain multiple properties.",
```
lucix-aws, line 146: FYI this is removed because it was a duplicated test case -- the one above it is identical.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/ParseFn.kt
```diff
@@ -0,0 +1,90 @@
...
+ */
+internal fun configurationProfile(input: FileLine, currentProfile: Token.Profile?, lastProperty: Token.Property?): Token.Profile? {
+    if (!input.isProfile()) return null
+
+    val parts = input.content
+        .stripInlineComments()
+        .stripComments()
+        .drop(1)
+        .dropLast(1)
+        .splitWhitespace(limit = 2)
```
ianbotsf, line 27: **Correctness**: This doesn't appear to handle whitespace between `[` and `profile`.

aajtodd (approved):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/Text.kt
```diff
@@ -0,0 +1,45 @@
...
+
+internal fun contextMessage(message: String, lineNumber: Int? = null): String = buildString {
+    append(message)
+    lineNumber?.let { append(" on line $lineNumber.") }
+    append(" $helpText")
+}
+
+// what constitutes a "valid" profile/property identifier is not explicitly stated in the SEP (nor is it consistent
+// across SDKs) but the one universally rejected component is whitespace
+internal fun String.isValidIdentifier(): Boolean = !any { it.isWhitespace() }
```
ianbotsf, line 29: **Nit**: `!any` → `none`

ianbotsf (changes requested):

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/config/profile/SpecTestSuites.kt
```diff
@@ -188,7 +174,7 @@ internal const val parserTestSuiteJson = """
         "configFile": "[profile foo]\n= value"
       },
       "output": {
-        "errorContaining": "Property did not have a name"
+        "errorContaining": "Encountered unexpected token"
```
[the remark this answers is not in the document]
lucix-aws: You're correct that it has generalized all parsing errors, I felt the tradeoff was ideal given that the line of the failed parse is always included for all of these.

I also feel that the SEP is a bit disingenuous in the way it lists out the different "fail fast" cases, and that ultimately, all of them _are_ really more accurately represented as "unexpected token". For example, one of the cases is
- Property definitions without an = sign cause parsing to fail immediately.

But realistically, how would you even determine that something was _trying_ to be a property definition without that = sign? To me that implies there's some additional parsed state of "recognized token, but critically invalid".

For those reasons I would argue it's better to follow the intent of the SEP rather than the letter.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsConfigParser.kt
```diff
@@ -6,185 +6,146 @@
...
+    var currentProperty: Token.Property? = null
+    var currentParentMap: MutableMap<String, String>? = null
+
+    for (token in this@toProfileMap) {
+        when (token) {
+            is Token.Profile -> {
+                currentProfile = token
+                currentProperty = null
+
+                if (containsKey(token)) continue
```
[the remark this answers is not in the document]
lucix-aws: This is allowed per the SEP, it's valid to have two separate profile declarations with their own keys, you basically just treat it as one profile in the output. One of the existing unit tests (not shown in diff) does cover this case.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsConfigParser.kt
```diff
@@ -6,185 +6,146 @@
...
+                    }
+                    continue
+                }
 
-    return mergeProfiles(tokenMap)
+                val profile = this[currentProfile]!!
+                if (profile[currentProperty.key] is AwsConfigValue.String) { // convert newly recognized parent to map
+                    currentParentMap = mutableMapOf()
+                    profile[currentProperty.key] = AwsConfigValue.Map(currentParentMap)
+                }
```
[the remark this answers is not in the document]
lucix-aws: Since the tokens are in declaration order after `tokenize`, the last defined value for the key will always win, which is intentional per the SEP.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
```diff
@@ -14,69 +15,70 @@ import aws.sdk.kotlin.runtime.config.retries.RetryMode
  * @property name name of profile
  * @property properties key/value pairs of properties specified by the active profile, accessible via [Map<K, V>]
  */
+@InternalSdkApi
 public data class AwsProfile(
```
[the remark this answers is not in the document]
lucix-aws: We had discussed that it was most likely incorrectly non-internal before, hence the change. I don't think any of those cases are strictly _off_ the table, though.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsConfigParser.kt
```diff
@@ -6,185 +6,146 @@
...
+                    }
+                    continue
+                }
 
-    val tokenMap = buildMap<Token.Profile, Map<String, String>> {
-        var lastProfile: Token.Profile? = null
+                put(token, mutableMapOf())
+            }
+            is Token.Property -> {
+                currentProfile as Token.Profile
```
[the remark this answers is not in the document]
lucix-aws: It's guaranteed within the scope of this code because the tokens are in declaration order (I believe the documentation for the `tokenize` function mentions this).

FWIW I'm not the world's biggest fan of having this implied context persist across two logically discrete pieces of code. Ultimately, since they're coupled and internal to the parser, I felt it was fine.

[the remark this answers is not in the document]
lucix-aws: I did consider adding profile/parent information to the tokens for properties/sub-properties, that would allow you to ignore the order of the tokens (to some extent, you'd still have to know declaration order for the "last wins" part of building the map) and instead sort them and avoid having to trust that you're not going to incorrectly index at some point, but I just went back to the fact that this code is internal, and the interaction between `tokenize` and `toProfileMap` is static, and left it alone.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/FileLine.kt
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/**
+ * Matches a NON-inline comment line, e.g.:
+ * - `; a comment`
+ * - `# a comment`
+ */
+internal fun FileLine.isComment() = content.trim().let {
+    it.startsWith(Literals.COMMENT_1) || it.startsWith(Literals.COMMENT_2)
+}
```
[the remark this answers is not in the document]
lucix-aws: Not sure what you mean? In the case of `isComment` here it's `trim()`ing the content before it makes that check. Profiles can certainly end with comments, I did note there's not really a clear directive on whether you can have leading whitespace on them, but since there's no difference between the two I didn't see the harm in ignoring it.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/FileType.kt
```diff
@@ -0,0 +1,67 @@
...
+    CONFIGURATION(
+        AwsSdkSetting.AwsConfigFile,
+        listOf(::configurationProfile, ::property, ::continuation, ::subProperty),
+        listOf("~", ".aws", "config"),
+    ),
+    CREDENTIAL(
+        AwsSdkSetting.AwsSharedCredentialsFile,
+        listOf(::credentialProfile, ::property, ::continuation, ::subProperty),
+        listOf("~", ".aws", "credentials"),
+    ),
```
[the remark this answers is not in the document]
lucix-aws: This is ported from what we had before, I'll double-check this.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/ParseFn.kt
lucix-aws: `splitWhitespace` here is a homespun utility in `Text.kt` and it filters out blank (i.e. `isBlank()`) segments.

There is an existing unit test that covers this as well.

lucix-aws: e.g.
```kt
fun main() {
	println(" 		  		 	  			 	profile 	 	 qux".splitWhitespace(limit = 2))
}

fun String.splitWhitespace(limit: Int): List<String> =
    trim().split(" ", "\t", limit = limit).filter(String::isNotBlank).map(String::trim)
```
gives us
`[profile, qux]`

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/config/profile/SpecTestSuites.kt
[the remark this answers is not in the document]
aajtodd: I think I was more concerned with the users ability to track down the issue, if the error contains the line/col number then it's probably good enough

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/Token.kt
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.sdk.kotlin.runtime.config.profile
+
+/**
+ * Tokens representing state declared in AWS configuration files. Other types such as empty lines
+ * or comments are filtered out before parsing.
+ */
+internal sealed class Token {
```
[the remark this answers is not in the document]
aajtodd: Are you suggesting `sealed interface` or just an `interface`? I think there is value in knowing the set of token implementations (aka sealed).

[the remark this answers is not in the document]
ianbotsf: Yes a `sealed interface`, sorry.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsConfigParser.kt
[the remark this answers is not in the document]
ianbotsf: OK, and `tokenize` guarantees that a profile always precedes the first property?

This code seems to muddle the responsibilities of **lexing** (turning a text stream into tokens and categorizing them into types) and **parsing** (turning a token stream into a syntax tree, identifying context between elements). This means implicit contracts are present in disparate parts of the code, which make it difficult to follow and easy to break. I wonder if this code would be simpler and easier to read if those responsibilities were more clearly separated.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsConfigParser.kt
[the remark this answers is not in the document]
ianbotsf: It fees like we should emit a warning here. Users may be surprised if their config values are silently discarded, especially if they just don't scroll down far enough in their file.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/FileLine.kt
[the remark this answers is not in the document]
ianbotsf: Sorry, I missed the `trim` in both of these methods.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/FileType.kt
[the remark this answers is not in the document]
lucix-aws: FYI, these segments are passed through a `normalizePath` method downstream that explicitly handles home directory expansion.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
```diff
@@ -37,8 +39,23 @@ public abstract class AbstractAwsSdkClientFactory<
...
-        builder.config.region = builder.config.region ?: resolveRegion()
-        builder.config.retryStrategy = builder.config.retryStrategy ?: resolveRetryStrategy()
+        val tracer = if (builder is TracingClientConfig.Builder) {
+            if (builder.tracer == null) builder.tracer = defaultTracer()
+            builder.tracer!!
+        } else {
+            defaultTracer()
+        }
+
+        coroutineContext.withRootTraceSpan(tracer.createRootSpan("fromEnvironment")) {
```
lucix-aws, line 49: **Question**: unsure about the exact span hierarchy, but this seemed correct. This root span will then disappear in the built client, and we go back to creating root spans for each operation.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
```diff
@@ -37,8 +39,23 @@ public abstract class AbstractAwsSdkClientFactory<
...
+            coroutineContext.withChildTraceSpan("resolveRegion") {
```
aajtodd, line 50: This seems unnecessary to have two child spans, what are we gaining?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
```diff
@@ -25,7 +27,7 @@ public abstract class AbstractAwsSdkClientFactory<
     TConfigBuilder,
     TClient : SdkClient,
     TClientBuilder : SdkClient.Builder<TConfig, TConfigBuilder, TClient>,
-    > : SdkClientFactory<TConfig, TConfigBuilder, TClient, TClientBuilder>
+    >(private val clientName: String) : SdkClientFactory<TConfig, TConfigBuilder, TClient, TClientBuilder>
```
aajtodd, line 30: Question/Correctness: We have a property [serviceName](https://github.com/awslabs/smithy-kotlin/blob/cbf83d8bbe631ddd309f35fda08d36084bae1c02/runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/SdkClient.kt#L17) off `SdkClient` that isn't configurable. We just [codegen the override directly ](https://github.com/awslabs/smithy-kotlin/blob/a639523d2d09c3c73c6ddd5ad98f1cb19f9295a9/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/ServiceClientGenerator.kt#L164). Should we just move this to a configurable property on `SdkClientConfig.Builder` (with it defaulted the way it is now via codegen)?

Taking a step back though, do we even need this property? If we aren't configured with a tracer we could probably just use the class name as a fallback. 

e.g.

```kotlin
DefaultTracer(LogingTraceProbe, builder::class.simpleName)
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsConfigLoader.kt
```diff
@@ -42,24 +49,30 @@ public suspend fun loadActiveAwsProfile(platform: PlatformProvider): AwsProfile
...
-    mergeProfiles(
-        parse(FileType.CONFIGURATION, platform.readFileOrNull(source.configPath)?.decodeToString()),
-        parse(FileType.CREDENTIAL, platform.readFileOrNull(source.credentialsPath)?.decodeToString()),
-    )
+@InternalSdkApi
+public suspend fun loadAwsProfiles(platform: PlatformProvider, source: AwsConfigurationSource): AwsProfiles =
+    coroutineContext.withChildTraceSpan("loadAwsProfiles") {
+        // merged AWS configuration based on optional configuration and credential file contents
+        mergeProfiles(
+            parse(coroutineContext.traceSpan, FileType.CONFIGURATION, platform.readFileOrNull(source.configPath)?.decodeToString()),
```
aajtodd, line 57: correctness: this is doing file IO on the wrong dispatcher thread potentially

```kotlin
withContext(SdkDispatchers.IO) {
    mergeProfiles (
        ... 
    )
}
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsConfigParser.kt
```diff
@@ -210,191 +177,5 @@ private fun mergeProfiles(tokenIndexMap: Map<Token.Profile, Map<String, String>>
...
-    "$message on line $lineNumber. $helpText"
-
-private fun warn(message: String, lineNumber: Int) {
-    logger.warn(contextMessage(message, lineNumber))
-}
-
-private fun throwParseException(message: String, lineNumber: Int): Nothing {
-    throw AwsConfigParseException(contextMessage(message, lineNumber))
-}
+private fun TraceSpan.warnParse(message: String) =
```
aajtodd, line 180: suggestion: Should probably be `inline` and take the `() -> Any` lambda as an argument instead of forcing the message to be formatted on every call

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
lucix-aws: I mean, these are two different "events" that delegate somewhat deep, it seems useful to have the spans capture that, but no strong opinion.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsConfigParser.kt
```diff
@@ -210,191 +177,5 @@ private fun mergeProfiles(tokenIndexMap: Map<Token.Profile, Map<String, String>>
...
+    warn("AwsConfigParse", content = { contextMessage(message) })
```
ianbotsf, line 181: **Nit**: There's no real class here (although that seems like it'd simplify passing around the `traceSpan`) but the file name is **AwsConfigParser** (with an "r"), which seems like a better choice for source component.

ianbotsf, line 181: **Suggestion**: `contextMessage` has the capability to show a line number where the problem occurred but we don't use that here. I think it'd be especially useful in larger files to see where warnings are occurring. Could we enhance `Token` to remember its line number and then pass the token into this helper method so a line number can be passed to `contextMessage`?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
```diff
@@ -6,77 +6,92 @@
...
+            is AwsConfigValue.String -> {
+                require(subKey == null) { "property '$key' is a string, but caller specified a sub-key" }
+                it.value
+            }
+            is AwsConfigValue.Map -> {
+                require(subKey != null) { "property '$key' has sub-properties, caller must specify a sub-key" }
+                it[subKey]
+            }
+        }
+    }
```
ianbotsf, line 34: **Question**: These `require` calls will throw `IllegalArgumentException`. Is that what we want or would `ConfigurationException` be better?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
ianbotsf: Using the builder class name as a default won't work...it's always `Builder`.

I like the idea of a configurable name retrievable from the builder. If we do, we should _move_ the service name default codegen from [the config-level tracer instantiation](https://github.com/awslabs/smithy-kotlin/blob/39e45f76eca4c4fee4f32202f4d2645bd324f91d/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/util/RuntimeConfigProperty.kt#L139) and then reuse the value in constructing the `DefaultTracer` (rather than codegenning it twice).

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
lucix-aws: Definitely the former, `IllegalArgumentException` is what is says, `ConfigurationException` right now is being used when there's some issue with the config itself (e.g. a retrieved profile value was invalid, we couldn't load a region anywhere, etc.)

aajtodd: If anything wouldn't `IllegalStateException` be more appropriate than `IllegalArgumentException`? (i.e `check` not `require`)

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
```diff
@@ -40,22 +40,18 @@ public abstract class AbstractAwsSdkClientFactory<
...
         }
 
-        coroutineContext.withRootTraceSpan(tracer.createRootSpan("fromEnvironment")) {
-            coroutineContext.withChildTraceSpan("resolveRegion") {
-                builder.config.region = builder.config.region ?: resolveRegion()
-            }
-            coroutineContext.withChildTraceSpan("resolveRetryStrategy") {
-                builder.config.retryStrategy = builder.config.retryStrategy ?: resolveRetryStrategy()
-            }
+        coroutineContext.withRootTraceSpan(tracer.createRootSpan("From environment")) {
```
ianbotsf, line 49: **Nit**: I still don't think "From environment" is a helpful span title. It describes a method name or a strategy, not necessarily a taxonomical lifespan. I suggest something like "Initialization", "Config resolution", etc.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsConfigParser.kt
```diff
@@ -72,27 +68,27 @@ internal fun tokenize(type: FileType, input: String): List<Token> = buildList {
             lastProperty = token
         }
 
-        add(token)
+        add(line to token)
```
ianbotsf, line 71: **Question**: This works but seems like it has a wide impact on type signatures elsewhere in the code. Would it be easier to make `FileLine` a property of `Token`? Then you wouldn't have to worry about `Pair<FileLine, Token>` or explicitly passing the line to `warnParse`.

ianbotsf (approved):

PR #867, opened 2023-03-06
misc: add binary compatibility validation to CI

lauzadis:
Adds the binary-compatibility-validator plugin's `apiCheck` task to CI.
This will run after a build and before any tests are run, failing if there is a binary compatibility error.

## Issue \#

N/A
## Description of changes

The binary compatibility validator is currently only running validation during local builds and not in CI. This is allowing PRs with binary compatibility errors to be merged which is undesired.

### DIFF
--- a/builder.json
+++ b/builder.json
@@ -55,7 +55,6 @@
         "codebuild-svc-check": {
             "!imports": [],
             "!test_steps": [
-                "{gradlew} apiCheck",
                 "{gradlew} test allTests"
             ]
         },


builder.json
```diff
@@ -34,8 +35,11 @@
     "build_steps": [
         "{gradlew} assemble --parallel"
     ],
+    "post_build_steps": [
+        "{gradlew} publishToMavenLocal"
+    ],
```
lauzadis, line 40: **question:** is there any reason this was in `test_steps`? it's in `post_build_steps` in the other repositories

aajtodd: Probably because of the way upstream dependencies work. They don't run the test steps when referencing the upstream dependency and we don't need the publish to maven local here unless we are testing.

builder.json
```diff
@@ -51,6 +55,7 @@
         "codebuild-svc-check": {
             "!imports": [],
             "!test_steps": [
+                "{gradlew} apiCheck",
```
aajtodd, line 58: service check batch shouldn't need this since we don't check generated service APIs

aajtodd (approved):

ianbotsf (approved):

PR #869, opened 2023-03-07
feat: load endpoint and s3 settings from env and shared config

lucix-aws:
## Issue \#
#206 

## Description of changes
* Add support for shared config properties:
  * FIPS and dualstack toggle
  * S3 accelerate and virtualhost toggle
* Make the relevant boolean fields nullable in builder form, such that their presence can be checked when loading from env. Env values will be ignored if an explicit `true` or `false` is set in code.
* Streamline config loading in `fromEnvironment` such that the active profile is only loaded once (if needed).

### DIFF
--- a/aws-runtime/aws-config/api/aws-config.api
+++ b/aws-runtime/aws-config/api/aws-config.api
@@ -1241,8 +1241,8 @@ public final class aws/sdk/kotlin/runtime/config/profile/AwsProfileKt {
 	public static final fun getAwsAccessKeyId (Laws/sdk/kotlin/runtime/config/profile/AwsProfile;)Ljava/lang/String;
 	public static final fun getAwsSecretAccessKey (Laws/sdk/kotlin/runtime/config/profile/AwsProfile;)Ljava/lang/String;
 	public static final fun getAwsSessionToken (Laws/sdk/kotlin/runtime/config/profile/AwsProfile;)Ljava/lang/String;
-	public static final fun getBooleanStrictOrNull (Laws/sdk/kotlin/runtime/config/profile/AwsProfile;Ljava/lang/String;Ljava/lang/String;)Ljava/lang/Boolean;
-	public static synthetic fun getBooleanStrictOrNull$default (Laws/sdk/kotlin/runtime/config/profile/AwsProfile;Ljava/lang/String;Ljava/lang/String;ILjava/lang/Object;)Ljava/lang/Boolean;
+	public static final fun getBooleanOrNull (Laws/sdk/kotlin/runtime/config/profile/AwsProfile;Ljava/lang/String;Ljava/lang/String;)Ljava/lang/Boolean;
+	public static synthetic fun getBooleanOrNull$default (Laws/sdk/kotlin/runtime/config/profile/AwsProfile;Ljava/lang/String;Ljava/lang/String;ILjava/lang/Object;)Ljava/lang/Boolean;
 	public static final fun getCredentialProcess (Laws/sdk/kotlin/runtime/config/profile/AwsProfile;)Ljava/lang/String;
 	public static final fun getMaxAttempts (Laws/sdk/kotlin/runtime/config/profile/AwsProfile;)Ljava/lang/Integer;
 	public static final fun getRegion (Laws/sdk/kotlin/runtime/config/profile/AwsProfile;)Ljava/lang/String;


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AwsSdkSetting.kt
```diff
@@ -143,6 +143,16 @@ public sealed class AwsSdkSetting<T>(
...
+
+    /**
+     * Whether to use FIPS endpoints when making requests.
+     */
+    public object AwsUseFipsEndpoint : AwsSdkSetting<Boolean>("AWS_USE_FIPS_ENDPOINT", "aws.useFipsEndpoint")
+
+    /**
+     * Whether to use dual-stack endpoints when making requests.
+     */
+    public object AwsUseDualStackEndpoint : AwsSdkSetting<Boolean>("AWS_USE_DUALSTACK_ENDPOINT", "aws.useDualstackEndpoint")
```
lucix-aws, line 155: The lowercase 's' in the system property name is intentional and matches java.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/region/ProfileRegionProvider.kt
```diff
@@ -15,8 +17,7 @@ import aws.smithy.kotlin.runtime.util.asyncLazy
  */
 internal class ProfileRegionProvider(
     private val platformProvider: PlatformProvider = PlatformProvider.System,
+    private val profile: LazyAsyncValue<AwsProfile> = asyncLazy { loadActiveAwsProfile(platformProvider) },
 ) : RegionProvider {
-    private val profile = asyncLazy { loadActiveAwsProfile(platformProvider) }
-
     override suspend fun getRegion(): String? = profile.get().region
 }
```
ianbotsf, line 23: **Comment**: `platformProvider` is now unnecessary as a separate parameter. It can be moved into the `asyncLazy` default.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/endpoints/Endpoints.kt
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Attempts to resolve the enabled state of dual-stack endpoints from the environment.
+ */
+@InternalSdkApi
+public suspend fun resolveUseDualStack(
+    provider: PlatformProvider = PlatformProvider.System,
+    profile: LazyAsyncValue<AwsProfile> = asyncLazy { loadActiveAwsProfile(provider) },
+): Boolean? =
+    AwsSdkSetting.AwsUseDualStackEndpoint.resolve(provider)
+        ?: profile.get().useDualStack
```
ianbotsf, line 38: **Comment**: The `provider` parameters are unnecessary. They can be moved into the `asyncLazy` defaults.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
```diff
@@ -95,3 +95,33 @@ public val AwsProfile.retryMode: RetryMode?
...
+        "false" -> false
+        null -> null
+        else -> throw ConfigurationException(
+            buildString {
+                append("Failed to parse config property $key")
+                subKey?.let { append(".$it") }
+                append(" as a boolean")
+            },
+        )
+    }
```
ianbotsf, line 127: **Comment**: In Kotlin, `strict` generally means case-sensitive (e.g., [`toBooleanStrict`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/to-boolean-strict.html)) which doesn't seem like what's desired here since you use `lowercase` to avoid case-sensitivity.

This looks like it can be simplified to:

```kotlin
public fun AwsProfile.getBooleanStrictOrNull(key: String, subKey: String? = null): Boolean? =
    getOrNull(key, subKey)?.toBoolean() ?? throw ConfigurationException(
        buildString {
            append("Failed to parse config property $key")
            subKey?.let { append(".$it") }
            append(" as a boolean")
        },
    )
```

aws-runtime/aws-core/common/src/aws/sdk/kotlin/runtime/client/AwsSdkClientConfig.kt
```diff
@@ -19,12 +19,34 @@ public interface AwsSdkClientConfig : SdkClientConfig {
...
+        /**
+         * Flag to toggle whether to use [FIPS](https://aws.amazon.com/compliance/fips/) endpoints when making requests.
+         */
+        public var useFips: Boolean?
+
+        /**
+         * Flag to toggle whether to use dual-stack endpoints when making requests.
+         * See [https://docs.aws.amazon.com/sdkref/latest/guide/feature-endpoints.html] for more information.
+         */
+        public var useDualStack: Boolean?
```
ianbotsf, line 50: **Suggestion**: It would be helpful for users to indicate what the defaults for these fields are when not provided. Applies to codegen as well.

services/s3/common/src/aws/sdk/kotlin/services/s3/internal/FinalizeConfigImpl.kt
```diff
@@ -0,0 +1,30 @@
...
+}
+
+private val AwsProfile.forcePathStyle: Boolean?
+    get() = getOrNull("s3", "addressing_style")?.let {
+        when (it) {
+            "virtual", "auto" -> false
+            "path" -> true
+            else -> throw ConfigurationException("invalid value '$it' for config property s3.addressing_style")
+        }
+    }
```
ianbotsf, line 27: **Question**: Case-sensitive matching is required here?

ianbotsf (commented):
**Question**: Looks like we'll still load the profile again for credentials. Should we plumb this through to the credentials chain as well?

aws-runtime/aws-core/common/src/aws/sdk/kotlin/runtime/client/AwsSdkClientConfig.kt
```diff
@@ -19,12 +19,34 @@ public interface AwsSdkClientConfig : SdkClientConfig {
...
          * The AWS region (e.g. `us-west-2`) to make requests to. See about AWS
          * [global infrastructure](https://aws.amazon.com/about-aws/global-infrastructure/regions_az/) for more
          * information
          */
         public var region: String?
+
+        /**
+         * Flag to toggle whether to use [FIPS](https://aws.amazon.com/compliance/fips/) endpoints when making requests.
+         */
+        public var useFips: Boolean?
```
aajtodd, line 44: **Question**: Why nullable?

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/CompanionObjectWriter.kt
```diff
@@ -0,0 +1,33 @@
...
+import software.amazon.smithy.kotlin.codegen.core.withBlock
+import software.amazon.smithy.kotlin.codegen.integration.SectionWriter
+import software.amazon.smithy.kotlin.codegen.lang.KotlinTypes
+import software.amazon.smithy.kotlin.codegen.rendering.ServiceClientGenerator
+
+/**
+ * Overrides the service client companion object for how a client is constructed, with the ability to extend inside the
+ * declaration itself.
+ */
+class CompanionObjectWriter(private val extend: (KotlinWriter.() -> Unit)? = null) : SectionWriter {
```
aajtodd, line 14: **nit**: This name doesn't tell me much other than it is a writer for _some_ kind of companion object. Consider renaming to something more specific (e.g. `ServiceClientCompanionObjectWriter`)

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
```diff
@@ -23,22 +31,26 @@ class ClientConfigIntegration : KotlinIntegration {
         model.expectShape<ServiceShape>(settings.service).isS3
 
     companion object {
-        val EnableAccelerateProp: ConfigProperty = ConfigProperty.Boolean(
-            "enableAccelerate",
-            defaultValue = false,
+        val EnableAccelerateProp: ConfigProperty = ConfigProperty {
+            name = "enableAccelerate"
+            useSymbolWithNullableBuilder(KotlinTypes.Boolean)
```
aajtodd, line 36: **Question**: Again why change to nullable?

services/s3/common/src/aws/sdk/kotlin/services/s3/internal/FinalizeConfigImpl.kt
```diff
@@ -0,0 +1,30 @@
+/*
```
aajtodd, line 1: **Nit**: filename `ConfigResolvers.kt` or something would be a little more meaningful (IMHO anyway)

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/endpoints/Endpoints.kt
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.sdk.kotlin.runtime.config.endpoints
```
aajtodd, line 5: **Bikeshed**: Wonder if we shouldn't move all the config resolvers to `aws.sdk.kotlin.runtime.config.resolvers` package (including `resolveRetryStrategy`/`resolveRegion`, etc) 🤔 

The `retries` package could be dissolved in favor of just moving `RetryMode `to the root `aws.sdk.kotlin.config` package. 

Probably not something to worry about in this PR though.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
lucix-aws: It's mentioned in the PR description, but we have to make all of these env-configurable params nullable in the builder, so we only apply the env version if they're unset.

services/s3/common/src/aws/sdk/kotlin/services/s3/internal/FinalizeConfigImpl.kt
lucix-aws: The [public docs](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#:~:text=or%20s3api%20namespaces.-,addressing_style,-Specifies%20which%20addressing) imply that it is:

> Valid values are: path, virtual, and auto.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
lucix-aws: You're right, and furthermore what's more important is the values we _expect_ in the config, which do appear to be case-sensitive based on the public docs (https://docs.aws.amazon.com/sdkref/latest/guide/feature-endpoints.html).

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/endpoints/Endpoints.kt
lucix-aws: The `AwsSdkSetting` resolvers still need the provider, are you suggesting we just hardcore to the system provider there as well?

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
aajtodd: RTFD 🤦 

If we wanted to (and I'm not saying we do necessarily) we could keep them as non-nullable if all we care about is checking that they are set or not.

```kotlin
class Foo {
    val _property: Boolean? = ...
    val property: Boolean
        get() = _property ?: false

}
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
ianbotsf: I don't see any mention of case sensitivity in those docs. The only examples shown are lowercase but I'd argue that mixed/uppercase versions seem very clear in their intent.

services/s3/common/src/aws/sdk/kotlin/services/s3/internal/FinalizeConfigImpl.kt
ianbotsf: I don't see any mention of case sensitivity in those docs nor any reason not to support mixed/uppercase equivalents.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
```diff
@@ -100,19 +100,19 @@ public val AwsProfile.retryMode: RetryMode?
...
 
 /**
- * Parse a config value as a strict boolean.
+ * Parse a config value as a boolean, ignoring case.
  */
-public fun AwsProfile.getBooleanStrictOrNull(key: String, subKey: String? = null): Boolean? =
-    getOrNull(key, subKey)?.let {
+public fun AwsProfile.getBooleanOrNull(key: String, subKey: String? = null): Boolean? =
+    getOrNull(key, subKey)?.lowercase()?.let {
         it.toBooleanStrictOrNull() ?: throw ConfigurationException(
```
ianbotsf: **Nit**: `?.let { it.toBooleanStrictOrNull()` is unnecessary, can be simplified to `?.toBooleanStrictOrNull()`.

ianbotsf (approved):

aajtodd (approved):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
lucix-aws: There's a subtle difference there. As it's written it'll break out and return `null` if the key is missing. Pulling the expression out of the let would cascade back to the exception being thrown if the key was missing, which we don't want.

The lowercase call in the outer chain was a bit odd and served to hide that intent IMO, i moved it inside the `let`.

ianbotsf (approved):

PR #880, opened 2023-03-28
feat: add support for awsQuery-compatible error responses

lucix-aws:
Parse `x-amzn-query-error` header when present on responses from query-compatible servers, exposing the parsed code and type values within service exception metadata.

There are no services released today that consume the `awsQueryCompatible` trait. I recommend applying it to a json-protocol model yourself and codegenning to verify the changed output.

### DIFF
--- a/codegen/sdk/aws-models/proton.json
+++ b/codegen/sdk/aws-models/proton.json
@@ -239,7 +239,6 @@
                 }
             ],
             "traits": {
-                "aws.protocols#awsQueryCompatible": {},
                 "aws.api#service": {
                     "sdkId": "Proton",
                     "endpointPrefix": "proton"


codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/core/AwsHttpBindingProtocolGenerator.kt
```diff
@@ -140,6 +139,20 @@ abstract class AwsHttpBindingProtocolGenerator : HttpBindingProtocolGenerator()
             }
             .write("")
 
+        if (ctx.service.hasTrait<AwsQueryCompatibleTrait>()) {
```
aajtodd, line 142: **correctness**: This seems like it would be better off in `QueryHttpBindingProtocolGenerator` to isolate it to appropriate protocol(s). It looks like for S3 we have to re-implement all of this logic. I wonder if we couldn't get more re-use and allow concrete protocol implementations to override more fine grained sections: 

e.g.

```kotlin
protected open fun renderThrowOperationError(...) {
    writer.declareSection(...) { ... }

    writer.declareSection(...) { ... }

}

// override

override fun renderThrowOperationError(...) {
    // register section overrides
    writer.onSection(...) { ... }

    super.renderThrowOperationError(...)

}
```

Sections isn't the only way to achieve this of course but it's one options.

lucix-aws: I think this is the most generally correct place to put it. To clarify, right now the only protocol that they're supporting the trait for is AwsJson1_0, so if we _were_ to isolate it we would put there.

That said I think that this is general enough of a mechanism that it belongs here (that also makes it more future-proof, read on). There's nothing preventing this query-compatible mechanism from working with any HTTP-based protocol (it's just a header). If they ever enable support for other protocols it will just be available.

aajtodd: > right now the only protocol that they're supporting the trait for is AwsJson1_0, so if we were to isolate it we would put there

Good point.

---- 

> There's nothing preventing this query-compatible mechanism from working with any HTTP-based protocol (it's just a header)

Hmm I guess I can buy that for now but I'd wager this won't be used anywhere other than `awsJson1_0`. Would be nice if it was a mixin of some sort such that we could integrate into just the protocols that need it. I think reduced blast radius of any change is generally better but I'm also not opposed to leaving it here at the moment.

aajtodd (approved):

lauzadis (approved):

Issue #432, opened 2021-11-17
Allow configuration of SDK log mode by environment


ianbotsf:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue, please leave a comment

## Describe the Feature

Currently, the `sdkLogMode` config can be controlled at the service client level (and potentially soon at the shared config level via #431). This requires manual code changes, however, and may not be the most convenient in every scenario. Much like other client-level configuration, this should be manageable via environmental settings (e.g., JVM system properties or environment variables).

## Is your Feature Request related to a problem?

The problem here is that changing logging levels requires a Kotlin code change, recompilation, and redeployment of build artifacts into Lambda, Elastic Beanstalk, etc. If this setting were manageable by environmental settings, a change would be much simpler and quicker.

## Proposed Solution

I propose adding JVM system property and environment variable keys for detecting desired logging modes without explicit configuration in code. A stretch goal is parsing them from **config** profiles.

## Describe alternatives you've considered

None at this time.

## Additional Context

See **Is your Feature Request related to a problem?** above.

- [x] I may be able to implement this feature request

PR #897, opened 2023-04-13
feat!: Resolve `LogMode` from environment

lauzadis:
This PR renames `SdkLogMode` to `LogMode`.

 It also enables resolving `LogMode` from the user's environment when a client is instantiated using `fromEnvironment()`.

Places checked (in priority order):
- `sdk.logMode` JVM system property
- `SDK_LOG_MODE` environment variable

Users can configure the following `LogMode` values: `LogRequest`, `LogResponse`, `LogRequestWithBody`, `LogResponseWithBody`.

Composite `LogMode` values are supported too, they can be configured by adding a pipe `|` between values. For example, `LogRequest|LogResponse`.

Misc. changes:
- Relocate `TestPlatformProvider` to **smithy-kotlin**, delete the `aws-runtime:testing` module

## Issue \#

#432 

## Description of changes

Currently `LogMode` can only be configured in code during SDK client initialization. This change enables users to configure the value without changing any code. 

### DIFF
--- a/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/polly/PollyPresigner.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/polly/PollyPresigner.kt
@@ -4,7 +4,6 @@
  */
 package aws.sdk.kotlin.codegen.customization.polly
 
-import aws.sdk.kotlin.codegen.AwsKotlinDependency
 import aws.sdk.kotlin.codegen.PresignerGenerator
 import software.amazon.smithy.kotlin.codegen.KotlinSettings
 import software.amazon.smithy.kotlin.codegen.core.*
--- a/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3ErrorMetadataIntegration.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3ErrorMetadataIntegration.kt
@@ -4,7 +4,6 @@
  */
 package aws.sdk.kotlin.codegen.customization.s3
 
-import aws.sdk.kotlin.codegen.AwsKotlinDependency
 import aws.sdk.kotlin.codegen.AwsRuntimeTypes
 import software.amazon.smithy.kotlin.codegen.KotlinSettings
 import software.amazon.smithy.kotlin.codegen.core.*


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AwsSdkSetting.kt
```diff
@@ -153,6 +154,11 @@ public sealed class AwsSdkSetting<T>(
      * Whether to use dual-stack endpoints when making requests.
      */
     public object AwsUseDualStackEndpoint : AwsSdkSetting<Boolean>("AWS_USE_DUALSTACK_ENDPOINT", "aws.useDualstackEndpoint")
+
+    /**
+     * Which [SdkLogMode] to use for logging requests and responses, when one is not specified at the client level.
+     */
+    public object SdkLogMode : AwsSdkSetting<aws.smithy.kotlin.runtime.client.SdkLogMode>("AWS_SDK_LOG_MODE", "aws.sdkLogMode")
```
aajtodd, line 161: **fix**: This is AWS SDK Kotlin specific, we shouldn't pollute the `aws.` system property namespace. Java uses their own namespace for this ([example](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkSystemSetting.java#L128)).

So for us it should probably be `AWS_SDK_KOTLIN_SDK_LOG_MODE` and `aws.sdk.kotlin.sdkLogMode`

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
```diff
@@ -108,6 +109,15 @@ public val AwsProfile.useFips: Boolean?
 public val AwsProfile.useDualStack: Boolean?
     get() = getBooleanOrNull("use_dualstack_endpoint")
 
+/**
+ * Which [SdkLogMode] to use for logging requests and responses, when one is not specified at the client level.
+ */
+public val AwsProfile.sdkLogMode: SdkLogMode?
```
aajtodd, line 115: **fix**: Remove, we don't want to add non-standard profile properties to shared config files. Log mode will be via environment or system property only which I think is sufficient.

aajtodd (changes requested):
Looks ok but couple things we need to address.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
lauzadis: Sounds good, [the issue](https://github.com/awslabs/aws-sdk-kotlin/issues/432) originally referenced parsing from config files as a goal, but I can see why it would be good to avoid clashing profile properties

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
```diff
@@ -60,6 +60,7 @@ public abstract class AbstractAwsSdkClientFactory<
             builder.config.retryStrategy = builder.config.retryStrategy ?: resolveRetryStrategy(profile = profile)
             builder.config.useFips = builder.config.useFips ?: resolveUseFips(profile = profile)
             builder.config.useDualStack = builder.config.useDualStack ?: resolveUseDualStack(profile = profile)
+            builder.config.sdkLogMode = builder.config.sdkLogMode ?: resolveSdkLogMode()
```
aajtodd, line 63: **Question**: Should we just have `resolveSdkLogMode()` be the default for the property type such that all generated clients (AWS SDK or not) benefit from this?

e.g.

```kotlin
override val sdkLogMode: SdkLogMode = builder.sdkLogMode ?: resolveSdkLogMode()
```

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AwsSdkSetting.kt
```diff
@@ -153,6 +154,11 @@ public sealed class AwsSdkSetting<T>(
      * Whether to use dual-stack endpoints when making requests.
      */
     public object AwsUseDualStackEndpoint : AwsSdkSetting<Boolean>("AWS_USE_DUALSTACK_ENDPOINT", "aws.useDualstackEndpoint")
+
+    /**
+     * Which [SdkLogMode] to use for logging requests and responses, when one is not specified at the client level.
+     */
+    public object SdkLogMode : AwsSdkSetting<aws.smithy.kotlin.runtime.client.SdkLogMode>("AWS_SDK_KOTLIN_SDK_LOG_MODE", "aws.sdk.kotlin.sdkLogMode")
```
ianbotsf, line 161: **Question**: Repeated words in multi-word identifiers always feel redundant and redundant to me. Can we remove one of the "SDK"s from these identifiers?

Maybe a somewhat larger namespacing question would be: is `AWS_SDK_KOTLIN`/`aws.sdk.kotlin` how we want to identify Kotlin SDK-specific config options? Or should we build towards a future when more (perhaps _all_) SDKs have similar options for configuring logging?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/ResolveSdkLogMode.kt
```diff
@@ -0,0 +1,18 @@
...
+import aws.smithy.kotlin.runtime.client.SdkLogMode
+import aws.smithy.kotlin.runtime.util.PlatformProvider
+
+/**
+ * Attempt to resolve the SDK log mode used to make requests
+ */
+@InternalSdkApi
+public fun resolveSdkLogMode(
+    platformProvider: PlatformProvider = PlatformProvider.System,
+): SdkLogMode = AwsSdkSetting.SdkLogMode.resolve(platformProvider) ?: SdkLogMode.Default
```
ianbotsf, line 18: **Question**: Under what circumstances would the `resolve` call return `null`?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AwsSdkSetting.kt
lauzadis: I generally agree but I'm leaning towards leaving the redundant "SDK" because it reflects the exact name of the field being configured, `SdkLogMode`

> always feel redundant and redundant to me

is this a joke about redundancy? 😄

aws-runtime/aws-config/api/aws-config.api
```diff
@@ -1093,6 +1093,13 @@ public final class aws/sdk/kotlin/runtime/config/AwsSdkSetting$AwsWebIdentityTok
 	public static final field INSTANCE Laws/sdk/kotlin/runtime/config/AwsSdkSetting$AwsWebIdentityTokenFile;
 }
 
+public final class aws/sdk/kotlin/runtime/config/AwsSdkSetting$SdkLogMode : aws/sdk/kotlin/runtime/config/AwsSdkSetting {
+	public static final field INSTANCE Laws/sdk/kotlin/runtime/config/AwsSdkSetting$SdkLogMode;
+}
+
+public final class aws/sdk/kotlin/runtime/config/ResolveSdkLogModeKt {
+}
+
```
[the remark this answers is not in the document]
lauzadis: I took a look around, and I don't think it's possible

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
lauzadis: I tried but this is not easily done due to our dependencies. Adding the `resolveSdkLogMode` function as the default would require adding a dependency on aws-sdk-kotlin to smithy-kotlin to get that function imported.

It's also not possible to relocate the function because it relies on `AwsSdkSetting` which is also in aws-sdk-kotlin

aajtodd: Well we could have an equivalent `SdkSetting` in `smithy-kotlin`. I don't feel super strongly but if we force it to be only consumable `fromEnvironment()` then customers won't be able to enable additional logging without a code change potentially. Maybe thats ok 🤔

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AwsSdkSetting.kt
aajtodd: > Or should we build towards a future when more (perhaps all) SDKs have similar options for configuring logging?

I think this is solvable though by allowing multiple environment variables and system properties. If there are new cross SDK environment settings added that we have kotlin specific ones for we can just set a priority for the new settings.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
aajtodd: Many of the settings in `AwsSdkSetting` take effect regardless of `fromEnvironment()` though. `fromEnvironment()` is mostly there for dealing with the fact that if you aren't explicit we may have to do _some_ work that involves I/O to resolve credentials provider and potentially region. Things that affect debuggability should probably be easy to consider with just an environment change regardless of how the client is constructed. If they set an `SdkLogMode` explicitly it will of course still take precedence.

aws-runtime/aws-core/build.gradle.kts
```diff
@@ -21,7 +21,6 @@ kotlin {
         }
         commonTest {
             dependencies {
-                implementation(project(":aws-runtime:testing"))
             }
         }
```
ianbotsf, line 25: **Nit**: This entire source set block can be deleted now that there are no dependencies.

ianbotsf (approved):

aajtodd (approved):

PR #908, opened 2023-05-02
chore: lint _all_ the things

ianbotsf:
## Issue \#

(none)

## Description of changes

Similar to [smithy-kotlin#847](https://github.com/awslabs/smithy-kotlin/pull/847), update the lint paths by including all ***.kt** and ***.kts** files and then specifically excluding generated code. Fortunately, no regular Kotlin files were impacted (the allowlist was already pretty comprehensive) and the changes are all in Gradle scripts.

No copyright header rule was updated because it doesn't exist in the **aws-sdk-kotlin** repo. That will have to wait until our linting config is commonized into a separate repo.

Most of this PR is the lint fixes themselves. The only interesting file is **build.gradle.kts**.


lauzadis (approved):

build.gradle.kts
```diff
@@ -178,6 +173,7 @@ tasks.register<JavaExec>("ktlint") {
     classpath = configurations.getByName("ktlint")
     main = "com.pinterest.ktlint.Main"
     args = lintPaths
+    jvmArgs("--add-opens", "java.base/java.lang=ALL-UNNAMED")
```
aajtodd, line 176: **Question**: What does this do?

aajtodd (approved):

build.gradle.kts
ianbotsf: I don't know exactly what it does but it was necessary to get around issues when running `gg ktlintFormat`. I saw instances of [`NoClassDefFoundError`](https://github.com/pinterest/ktlint/issues/1195) and [`InaccessibleObjectException`](https://github.com/pinterest/ktlint/issues/1391) and the related GH issues mentioned this as a solution. [KtLint's integrations guide](https://pinterest.github.io/ktlint/install/integrations/#custom-gradle-integration) also mentions this may be necessary, albeit in the context of custom Gradle integration (rather than via a plugin like we're doing).

PR #928, opened 2023-05-24
fix: correct infinite pagination in S3 ListParts

ianbotsf:
## Issue \#

(none)

## Description of changes

This change fixes an infinite pagination bug with S3's `ListParts` API. This API, despite being modeled with the `@paginated` Smithy trait does not signal pagination termination by sending an empty value in the field indicated by the `outputToken` attribute (`ListPartsResponse.nextPartNumberMarker`). Instead, it uses the flag field `isTruncated` which is non-standard behavior. So this PR adds a customization to handle this API correctly.

**Companion PR**: [smithy-kotlin#858](https://github.com/awslabs/smithy-kotlin/pull/858)

### DIFF
--- a/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/TruncatablePaginationIntegration.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/TruncatablePaginationIntegration.kt
@@ -7,11 +7,10 @@ import software.amazon.smithy.kotlin.codegen.model.traits.PaginationTruncationMe
 import software.amazon.smithy.model.Model
 import software.amazon.smithy.model.shapes.MemberShape
 import software.amazon.smithy.model.shapes.ServiceShape
-import software.amazon.smithy.model.shapes.StructureShape
 import software.amazon.smithy.model.transform.ModelTransformer
 
-private val TRUNCATABLE_PAGINATION_OPS = mapOf(
-    "com.amazonaws.s3#ListParts" to "IsTruncated",
+private val TRUNCATION_MEMBER_IDS = setOf(
+    "com.amazonaws.s3#ListPartsOutput\$IsTruncated",
 )
 
 /**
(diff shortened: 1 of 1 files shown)


codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/TruncatablePaginationIntegration.kt
```diff
@@ -0,0 +1,44 @@
...
+import software.amazon.smithy.kotlin.codegen.model.expectShape
+import software.amazon.smithy.kotlin.codegen.model.traits.PaginationTruncationMember
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.MemberShape
+import software.amazon.smithy.model.shapes.ServiceShape
+import software.amazon.smithy.model.shapes.StructureShape
+import software.amazon.smithy.model.transform.ModelTransformer
+
+private val TRUNCATABLE_PAGINATION_OPS = mapOf(
+    "com.amazonaws.s3#ListParts" to "IsTruncated",
```
aajtodd, line 14: **Question**: Could this not just directly refer to the member shape ID?

aajtodd (approved):

aajtodd:
NOTE service check batch is failing

lauzadis (approved):

services/s3/e2eTest/src/PaginatorTest.kt
```diff
@@ -0,0 +1,86 @@
...
+                    uploadId = id
+                }
+                    .transform { it.parts?.forEach { emit(it.partNumber) } }
+                    .toList()
+                    .sorted()
+            }
+
+            assertContentEquals(expectedParts, actualParts)
+        } finally {
+            client.abortMultipartUpload {
```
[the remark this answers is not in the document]
ianbotsf: That's correct. Additionally, [`abortMultipartUpload`](https://docs.aws.amazon.com/AmazonS3/latest/API/API_AbortMultipartUpload.html) is called in the `finally` block, which frees the parts that were already uploaded.

ianbotsf:
> NOTE service check batch is failing

Looks like [that failing CI run](https://github.com/awslabs/aws-sdk-kotlin/actions/runs/5073378862/jobs/9112332439?pr=928) errored during its initial lookup of the CodeBuild status:

```
Build started successfully.
Build id gh-aws-sdk-kotlin-svc-check-batch:787062b5-f459-4e5e-a3c7-006255a41a35
Checking build status.
jq: error (at <stdin>:1): Cannot iterate over null (null)
.github/scripts/run-codebuild-batch-job.sh: line 46: -  - : syntax error: operand expected (error token is "- ")
```

It's likely the response from CodeBuild was in an unexpected format, possibly because we don't pause between starting the CB job and then initially checking its output. I'll follow up with an optimization of the script in a separate PR.

Reran CI and now it's passing.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/TruncatablePaginationIntegration.kt
ianbotsf: Yes and that would make the preprocessing simpler. Great suggestion.

PR #932, opened 2023-05-26
fix: correct the reporting of the service check build CI job

ianbotsf:
## Issue \#

(none)

## Description of changes

The service check build CI job occasionally fails when checking the status of the CodeBuild job too rapidly after starting it. This change moves the 10 second sleep to the start of the polling loop to ensure that the CodeBuild job is fully initialized before the first polling check.

This change also removes filtering on job identifiers starting with `JOB`. Batch identifiers no longer begin with `JOB` but are either a service name (e.g., `SAGEMAKER`) or a numbered batch (e.g., `BATCH13`). This led to finding zero jobs in every polling check, yielding the following output line:

```
Build is still in progress (failed=0; in_progress=0; succeeded=0; total=0), waiting...
```

### DIFF
--- a/.github/scripts/run-codebuild-batch-job.sh
+++ b/.github/scripts/run-codebuild-batch-job.sh
@@ -33,11 +33,12 @@ echo "Build id $BUILD_ID"
 
 BUILD_STATUS="IN_PROGRESS"
 while [ "$BUILD_STATUS" == "IN_PROGRESS" ]; do
-  echo "Checking build status."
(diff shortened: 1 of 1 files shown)


ianbotsf:
Now seeing correct output from the polling loop:

```
Build is still in progress (failed=0; in_progress=18; succeeded=5; total=23), waiting...
```

lauzadis (approved):

Issue #824, opened 2023-01-30
Handle closing credentials providers properly


ianbotsf:
As follow-up work from from making resourced managed, revisit how closing is handled in credentials providers to ensure underlying resources are released properly and that closed providers do not return (or retain) credentials.

PR #936, opened 2023-06-02
misc: update closeability of various types of `CredentialsProvider`

lauzadis:
Updates the closeability of various types of `CredentialsProvider`. In short, this PR removes the implementation of `Closeable` from providers where it doesn't make sense (i.e the provider has no resources which can meaningfully "close").

## Issue \#

https://github.com/awslabs/aws-sdk-kotlin/issues/824

## Description of changes
**Removed `Closeable` from:**
- `EnvironmentCredentialsProvider`
- `SsoCredentialsProvider`
- `StaticCredentialsProvider`
- `StsAssumeRoleCredentialsProvider`
- `StsWebIdentityCredentialsProvider`

---

### DIFF
--- a/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/DefaultChainBearerTokenProvider.kt
+++ b/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/DefaultChainBearerTokenProvider.kt
@@ -9,10 +9,8 @@ import aws.sdk.kotlin.runtime.config.AwsSdkSetting
 import aws.smithy.kotlin.runtime.http.auth.BearerToken
 import aws.smithy.kotlin.runtime.http.auth.BearerTokenProviderChain
 import aws.smithy.kotlin.runtime.http.auth.CloseableBearerTokenProvider
-import aws.smithy.kotlin.runtime.http.engine.DefaultHttpEngine
 import aws.smithy.kotlin.runtime.http.engine.HttpClientEngine
 import aws.smithy.kotlin.runtime.io.Closeable
-import aws.smithy.kotlin.runtime.io.closeIfCloseable
 import aws.smithy.kotlin.runtime.util.Attributes
 import aws.smithy.kotlin.runtime.util.PlatformProvider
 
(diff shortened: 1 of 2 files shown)


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/SsoCredentialsProvider.kt
```diff
@@ -85,7 +84,8 @@ public class SsoCredentialsProvider public constructor(
     public val ssoSessionName: String? = null,
 
     /**
-     * The [HttpClientEngine] to use when making requests to the AWS SSO service
+     * The [HttpClientEngine] instance to use to make requests. NOTE: This engine's resources and lifetime
+     * are NOT managed by the provider. Caller is responsible for closing.
      */
     private val httpClient: HttpClientEngine? = null,
```
ianbotsf, line 90: **Correctness**: This seems wrong. The client's lifetime is managed by the caller _if they provide one_. But if they don't, then `SsoClient` and `SsoTokenProvider`/`SsoOidcClient` will create new clients (different instances!) and we'll never explicitly close them, leaking resources. I think `SsoCredentialsProvider` _should_ be closeable and should default the client to `DefaultHttpEngine` at the provider level before passing to `SsoClient` and `SsoTokenProvider`/`SsoOidcClient`. We should in turn close the underlying `SsoClient` and `SsoTokenProvider`/`SsoOidcClient`.

Similar comment applies to `StsAssumeRoleCredentialsProvider` and `StsWebIdentityCredentialsProvider`.

ianbotsf (changes requested):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/DefaultChainBearerTokenProvider.kt
```diff
@@ -34,14 +36,19 @@ public class DefaultChainBearerTokenProvider(
     private val platformProvider: PlatformProvider = PlatformProvider.System,
     httpClient: HttpClientEngine? = null,
 ) : CloseableBearerTokenProvider {
+    private val manageEngine = httpClient == null
+    private val engine = httpClient ?: DefaultHttpEngine()
```
aajtodd, line 40: I don't think we need to do this, I'll comment on the overall PR about this more.

aajtodd (commented):
> Added default HTTP client engines to ProfileCredentialsProvider and DefaultChainBearerTokenProvider. These providers had an issue because the httpClient parameter in the constructor is nullable but was being treated as non-null. Instead, we should check if it's null, and if it is, create a DefaultHttpEngine which will be managed and closed when the provider is closed. This is [already how it's done in EcsCredentialsProvider](https://github.com/awslabs/aws-sdk-kotlin/blob/e233721ecbc368f5465ee963b7ecb5d880fc69e1/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt#L68-L69) today, just extending it to these other providers.

I don't think this is necessary. The STS and SSO credential providers basically treat the engine as a pass through parameter to the underlying service clients. If it's `null` the service clients will instantiate one for you (expensive), if it's not then it's cheap to instantiate them. The reason ECS creates one if it's `null` is there is no underlying service client so we have to handle the engine ourselves in the credential provider since it's not handled for us by a generated client (same thing in IMDS cred provider).

ianbotsf (approved):

aajtodd (approved):

PR #944, opened 2023-06-09
feat: update user agent string format

ianbotsf:
## Issue \#

(none)

## Description of changes

Updates the user agent header value to the new cross-SDK format. The biggest change is that key-value pairs are now separated by `#` instead of `/` (e.g., `md/foo#bar` instead of `md/foo/bar`). This change also re-enables several metadata which were previously disabled to work around length restrictions which have now been addressed.


lauzadis (approved):
Looks good!

Issue #242, opened 2021-03-04
Route53 Custom Error Unmarshalling


kggilmer:
ChangeResourceRecordSets uses a custom error format: aws/aws-sdk-go@3265860

*internal id*: 176558794

PR #961, opened 2023-06-23
fix: route53 custom error unmarshalling

0marperez:
## Issue \#

Github issue https://github.com/awslabs/aws-sdk-kotlin/issues/242

## Description of changes
Implements route 53 custom error unmarshalling for `InvaildChangeBatchResponse`. Tests included.

### DIFF
--- a/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/core/AwsHttpBindingProtocolGenerator.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/core/AwsHttpBindingProtocolGenerator.kt
@@ -111,7 +111,7 @@ abstract class AwsHttpBindingProtocolGenerator : HttpBindingProtocolGenerator() 
         writer.write("val payload = response.body.#T()", RuntimeTypes.Http.readAll)
             .write("val wrappedResponse = response.#T(payload)", RuntimeTypes.AwsProtocolCore.withPayload)
             .write("")
-            .declareSection(ProtocolErrorDeserialization) { }
+            .declareSection(ProtocolErrorDeserialization)
             .write("val errorDetails = try {")
             .indent()
             .call {


.changes/7f80de87-a59a-420e-a175-c21a89f93fe9.json
```diff
@@ -0,0 +1,8 @@
+{
+    "id": "7f80de87-a59a-420e-a175-c21a89f93fe9",
+    "type": "feature",
+    "description": "Route 53 custom error unmarshalling",
+    "issues": [
+        "awslabs/aws-sdk-kotlin#242"
+    ]
+}
```
ianbotsf, line 8: **Suggestion**: This fixes a bug, right? Customers won't get the correct exception type today?

Also, this description is not very meaningful to customers. I suggest something like ``"Correctly handle and throw `InvaildChangeBatch` responses from Route53"``.

0marperez: Yeah, I wasn't sure what to name it. I'll change it

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/CustomErrorUnmarshalling.kt
```diff
@@ -0,0 +1,243 @@
...
+
+        val builder = ChangeResourceRecordSetsResponse.Builder()
+        val payload = response.body.readAll()
+        if (payload != null) {
+            deserializeChangeResourceRecordSetsOperationBody(builder, payload)
+        }
+        return builder.build()
+    }
+}
+
```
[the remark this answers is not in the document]
0marperez: A good amount, let me see what I can do

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/route53/CustomErrorUnmarshallingIntegration.kt
```diff
@@ -0,0 +1,35 @@
...
+import software.amazon.smithy.kotlin.codegen.core.getContextValue
+import software.amazon.smithy.kotlin.codegen.integration.KotlinIntegration
+import software.amazon.smithy.kotlin.codegen.integration.SectionWriter
+import software.amazon.smithy.kotlin.codegen.integration.SectionWriterBinding
+import software.amazon.smithy.kotlin.codegen.model.expectShape
+import software.amazon.smithy.kotlin.codegen.rendering.protocol.HttpProtocolClientGenerator
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.ServiceShape
+
+class CustomErrorUnmarshallingIntegration : KotlinIntegration {
```
lauzadis, line 18: Could you give this a more precise name like `ChangeResourceRecordSetsUnmarshallingIntegration`? It helps when using the IDE search functionality

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/CustomErrorUnmarshalling.kt
lauzadis, line 1: Generally the file's name should match the class it's defining (in this case, `ChangeResourceRecordSetsOperationDeserializer`)

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/CustomErrorUnmarshalling.kt
```diff
@@ -0,0 +1,243 @@
+package aws.sdk.kotlin.services.route53.internal
+
+import aws.sdk.kotlin.services.route53.model.ChangeResourceRecordSetsResponse
```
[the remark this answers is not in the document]
lauzadis: The patterns seem fine, but I noticed we're missing the custom rules in **aws-sdk-kotlin**, they're [only in **smithy-kotlin**](https://github.com/awslabs/smithy-kotlin/tree/b2643e0f13031dbb6f212d4797f7baaac1997136/ktlint-rules/src/main/kotlin/software/aws/ktlint/rules). I made a [small backlog item](https://github.com/orgs/awslabs/projects/45/views/1?pane=issue&itemId=31624164) to fix this.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/route53/CustomErrorUnmarshallingIntegration.kt
lauzadis, line 1: Can you add a few test cases for this integration? Similar to [this test](https://github.com/awslabs/aws-sdk-kotlin/blob/main/codegen/smithy-aws-kotlin-codegen/src/test/kotlin/aws/sdk/kotlin/codegen/customization/s3/ContinueIntegrationTest.kt) for a custom S3 integration

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/CustomErrorUnmarshalling.kt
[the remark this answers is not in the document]
0marperez: Turns out none of it. Ideally I would do that but a lot of that code as marked as private or internal. The only function that is public (`parseRestXmlErrorResponse`) I ended up modifying.

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/ChangeResourceRecordSetsUnmarshalling.kt
```diff
@@ -0,0 +1,108 @@
...
+    override val code: String? = error?.code
+    override val message: String? = error?.message
+}
+
+// Models "Error" type in https://awslabs.github.io/smithy/1.0/spec/aws/aws-restxml-protocol.html#operation-error-serialization
+internal data class XmlError(
+    override val requestId: String?,
+    override val code: String?,
+    override val message: String?,
+) : RestXmlErrorDetails
```
ianbotsf, line 108: **Question**: Do we need these classes? Couldn't the deserializers return `ErrorDetails?` directly?

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/route53/ChangeResourceRecordSetsUnmarshallingIntegration.kt
```diff
@@ -0,0 +1,38 @@
...
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.model.shapes.ServiceShape
+import software.amazon.smithy.model.transform.ModelTransformer
+
+class ChangeResourceRecordSetsUnmarshallingIntegration : KotlinIntegration {
+    override val sectionWriters: List<SectionWriterBinding> = listOf(
+        SectionWriterBinding(AwsHttpBindingProtocolGenerator.ProtocolErrorDeserialization) { writer, originalValue ->
+            val lines = originalValue?.split("\n")
+            writer.write("#L\naws.sdk.kotlin.services.route53.internal.parseCustomXmlErrorResponse(payload) ?: #L", lines?.get(0) ?: "", lines?.get(1) ?: "")
```
ianbotsf, line 22: **Nit**: Solely referring to the original value with this lines list is pretty inscrutable. I suggest giving each line a clearer name so that it's easier to understand what's going on here. We can also clean up the null checking by just requiring a non-null value up front:

```kotlin
requireNotNull(originalValue) { "Unexpectedly empty original value" }
val (nullCheckLine, parseLine) = originalValue.split("\n", limit = 2)
writer.write("#L", nullCheckLine)
writer.write("aws.sdk.kotlin.services.route53.internal.parseCustomXmlErrorResponse(payload) ?: #L", parseLine)
```

ianbotsf (approved):
Looks pretty good. A few minor points left...fix and ship!

lauzadis (approved):
Great! No comments outside of what Ian already had.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/route53/ChangeResourceRecordSetsUnmarshallingIntegration.kt
```diff
@@ -0,0 +1,40 @@
...
+import software.amazon.smithy.kotlin.codegen.model.expectShape
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.model.shapes.ServiceShape
+import software.amazon.smithy.model.transform.ModelTransformer
+
+class ChangeResourceRecordSetsUnmarshallingIntegration : KotlinIntegration {
+    override val sectionWriters: List<SectionWriterBinding> = listOf(
+        SectionWriterBinding(AwsHttpBindingProtocolGenerator.ProtocolErrorDeserialization) { writer, originalValue ->
+            requireNotNull(originalValue) { "Unexpectedly empty original value" }
```
aajtodd, line 21: **nit**: `checkNotNull` would be more appropriate

`require*` functions throw illegal argument exception whereas `check` throws illegal state exception.

Though neither of these should be required..see my other comment.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/route53/ChangeResourceRecordSetsUnmarshallingIntegration.kt
```diff
@@ -0,0 +1,40 @@
...
+            val (nullCheckLine, parseLine) = originalValue.split("\n", limit = 2)
```
aajtodd, line 22: **fix**: This is making a lot of assumptions. It seems to assume that the format of this will always be:

```kotlin
        checkNotNull(payload){ "unable to parse error from empty response" }
        parseXyz(payload)
```

This isn't true for every protocol nor is it true in general once you introduce a section other section writers could manipulate it. 

This should likely just wholesale replace the section given it's two lines of code.

aajtodd (changes requested):

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/ChangeResourceRecordSetsUnmarshalling.kt
0marperez: One of the deserializers can, the other can't. I can have the one that can return `ErrorDetails?` do that but I would have to leave the other data class in there for the one that can't.

ianbotsf (approved):

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/route53/ChangeResourceRecordSetsUnmarshallingIntegration.kt
```diff
@@ -0,0 +1,34 @@
...
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.ServiceShape
+
+class ChangeResourceRecordSetsUnmarshallingIntegration : KotlinIntegration {
+    override val sectionWriters: List<SectionWriterBinding> = listOf(
+        SectionWriterBinding(AwsHttpBindingProtocolGenerator.ProtocolErrorDeserialization) { writer, _ ->
+            writer.withBlock("payload?.let {", "}\n") {
+                withBlock("aws.sdk.kotlin.services.route53.internal.parseCustomXmlErrorResponse(payload)?.let {", "}") {
+                    write("val builder = aws.sdk.kotlin.services.route53.model.InvalidChangeBatch.Builder()")
+                    write("builder.message = it.message")
```
aajtodd, line 23: **correctness**: This ignores the other fields possible like request ID. We already go through the process of deserializing this type. I think it might be better to just return the built exception type and throw it.

e.g.

```kotlin
parseCustomInvalidChangeBatchErrorResponse(payload)?.let {
     setAseErrorMetadata(it.exception, wrappedResponse, it.errorDetails)
     throw it.exception
}
```

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/ChangeResourceRecordSetsUnmarshalling.kt
```diff
@@ -0,0 +1,108 @@
...
+                }
+            }
+            XmlErrorResponse(messages, requestId ?: messages?.requestId)
+        } catch (e: DeserializationException) {
+            null // return so an appropriate exception type can be instantiated above here.
+        }
+    }
+}
+
+private object InvalidChangeBatchMessageDeserializer {
```
aajtodd, line 52: **question**: This is a nested type right? Can we not just re-use the generated deserializer for this then?

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/ChangeResourceRecordSetsUnmarshalling.kt
```diff
@@ -0,0 +1,108 @@
...
+package aws.sdk.kotlin.services.route53.internal
+
+import aws.smithy.kotlin.runtime.InternalApi
+import aws.smithy.kotlin.runtime.awsprotocol.ErrorDetails
+import aws.smithy.kotlin.runtime.serde.*
+import aws.smithy.kotlin.runtime.serde.xml.XmlDeserializer
+import aws.smithy.kotlin.runtime.serde.xml.XmlSerialName
+
+@InternalApi
+public suspend fun parseCustomXmlErrorResponse(payload: ByteArray): ErrorDetails? {
```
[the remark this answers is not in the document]
aajtodd: As I noted elsewhere we go through the process of deserializing the exception type here let's not lose all that work. Instead just return the built exception and the error details together as a pair or something and then you can just throw it.

aajtodd (changes requested):

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/ChangeResourceRecordSetsUnmarshalling.kt
```diff
@@ -0,0 +1,108 @@
...
+            }
+            XmlError(requestId, code, message)
+        } catch (e: DeserializationException) {
+            null // return so an appropriate exception type can be instantiated above here.
+        }
+    }
+}
+
+// XML Error response parser from RestXMLErrorDeserializer
+private interface RestXmlErrorDetails {
```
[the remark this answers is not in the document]
0marperez: It was there from the copy and pasting I did and was necessary for that code to work. I modified things and it's no longer there anymore

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/ChangeResourceRecordSetsUnmarshalling.kt
0marperez: I modified things and it's not exactly the same so no

aajtodd: Why is the generated deserializer not usable as is?

0marperez: Because of [this](https://github.com/awslabs/smithy-kotlin/blob/3e1d8618b0090b218461cffdcce20619d6b4119b/runtime/protocol/aws-xml-protocols/common/src/aws/smithy/kotlin/runtime/awsprotocol/xml/RestXmlErrorDeserializer.kt#L105C38-L105C38) line I believe. It should be `Messages` instead of `Error`. But I could be wrong

aajtodd: I see now it's not a nested type it's just a list of string. Ok I have further questions now, new comments incoming.

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/ChangeResourceRecordSetsUnmarshalling.kt
```diff
@@ -0,0 +1,94 @@
...
+
+    suspend fun deserialize(deserializer: Deserializer): String? {
+        var messages: String? = null
+
+        return try {
+            deserializer.deserializeStruct(OBJ_DESCRIPTOR) {
+                loop@ while (true) {
+                    when (findNextFieldIndex()) {
+                        MESSAGE_DESCRIPTOR.index ->
+                            if (messages == null) messages = deserializeString() else messages += "\n" + deserializeString()
```
aajtodd, line 73: **correctness**: This is modeled as a list field why are we joining it to a string with `\n`?

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/ChangeResourceRecordSetsUnmarshalling.kt
```diff
@@ -0,0 +1,94 @@
...
+    val details = InvalidChangeBatchDeserializer.deserialize(XmlDeserializer(payload, true)) ?: return null
+    val exception = buildInvalidChangeBatchException(details.messages)
+    val errorDetails = ErrorDetails("InvalidChangeBatch", details.messages, details.requestId)
+    return InvalidChangeBatchErrorResponse(errorDetails, exception)
+}
+
+private fun buildInvalidChangeBatchException(messages: String?): InvalidChangeBatch {
+    messages ?: throw DeserializationException("Missing message in InvalidChangeBatch XML response")
+    val builder = InvalidChangeBatch.Builder()
+    builder.message = messages
```
aajtodd, line 23: **correctness**: Why are we setting `message` to the result of parsing the list field? Also what if it populates both `Message` and `Messages`?

e.g.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<InvalidChangeBatch xmlns="https://route53.amazonaws.com/doc/2013-04-01/">
  <Messages>
    <Message>Message from list</Message>
  </Messages>
  <Message>Some other message</Message>
  <RequestId>b25f48e8-84fd-11e6-80d9-574e0c4664cb</RequestId>
</InvalidChangeBatch>
```

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/ChangeResourceRecordSetsUnmarshalling.kt
0marperez: I see how it looks like the `Messages` field here because of the variable but it's `Message` which is a String. I am assuming a response that looks like this here:

```
<?xml version="1.0" encoding="UTF-8"?>
<InvalidChangeBatch xmlns="https://route53.amazonaws.com/doc/2013-04-01/">
  <Messages>
    <Message>Tried to create resource record set duplicate.example.com. type A, but it already exists</Message>
    <Message>Tried to create resource record set duplicate.example.com. type B, but it already exists</Message>
  </Messages>
  <RequestId>b25f48e8-84fd-11e6-80d9-574e0c4664cb</RequestId>
</InvalidChangeBatch>
```

Is this even right/possible?

services/route53/common/src/aws/sdk/kotlin/services/route53/internal/ChangeResourceRecordSetsUnmarshalling.kt
0marperez: **Why are we setting message to the result of parsing the list field?** : 
I'm setting the `Message` value here which is a String not a String list. That variable name is pretty confusing and makes it seem like the `Messages` field. I'll change it
 
**What if it populates both Message and Messages?** : 
So I don't populate `Messages` right now I only extract the `Message`. I also asked Monday if I had to worry about seeing a `Message` not inside a `Messages` tag and got told that it shouldn't happen. So that's something we have to look out for as well?

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/core/AwsHttpBindingProtocolGenerator.kt
```diff
@@ -108,6 +111,7 @@ abstract class AwsHttpBindingProtocolGenerator : HttpBindingProtocolGenerator()
         writer.write("val payload = response.body.#T()", RuntimeTypes.Http.readAll)
             .write("val wrappedResponse = response.#T(payload)", RuntimeTypes.AwsProtocolCore.withPayload)
             .write("")
+            .declareSection(ProtocolErrorDeserialization) { }
```
aajtodd, line 114: **nit**: you don't need the empty `{ }` it's already the [default](https://github.com/awslabs/smithy-kotlin/blob/a639523d2d09c3c73c6ddd5ad98f1cb19f9295a9/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/AbstractCodeWriterExt.kt#L124)

aajtodd (approved):

PR #970, opened 2023-07-03
feat: make region optional

### LINKED ISSUE #969: Make region optional
AWS service clients need to be configurable without `region`. 99% of the time region will be required and this will result in a runtime error if not configured/resolved. However, there are use cases where region is in fact optional:
* global endpoints (e.g. STS - although STS has moved to favoring regional endpoints)
* Multi region access points (MRAP) where region may be derived from the ARN

NOTE: This should probably be done as part of the effort to support MRAP for S3/Event Bridge

ianbotsf:
## Issue \#

Closes #969 

## Description of changes

Makes region an optional client config parameter to support MRAP and future multi-region use cases. Note that failing to provide a region in single-region cases will still fail but will throw an exception at operation invocation instead of during client construction.

**Companion PR**: [smithy-kotlin#884](https://github.com/awslabs/smithy-kotlin/pull/884)

### DIFF
--- a/services/s3/e2eTest/src/S3TestUtils.kt
+++ b/services/s3/e2eTest/src/S3TestUtils.kt
@@ -45,7 +45,7 @@ object S3TestUtils {
             client.createBucket {
                 bucket = testBucket
                 createBucketConfiguration {
-                    locationConstraint = BucketLocationConstraint.fromValue(client.config.region)
+                    locationConstraint = BucketLocationConstraint.fromValue(client.config.region!!)
                 }
             }
 


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
```diff
@@ -124,7 +124,7 @@ public class ProfileCredentialsProvider(
         }
     }
 
-    private suspend fun LeafProvider.toCredentialsProvider(region: LazyAsyncValue<String>): CredentialsProvider =
+    private suspend fun LeafProvider.toCredentialsProvider(region: LazyAsyncValue<String?>): CredentialsProvider =
```
aajtodd, line 127: **correctness**: Should region be optional everywhere or just at the client config level? I'm not sure these credential provider components ever work correctly without region.

services/s3/common/test/aws/sdk/kotlin/services/s3/CreateClientTest.kt
```diff
@@ -18,12 +16,10 @@ import kotlin.test.assertFailsWith
  */
 @OptIn(ExperimentalCoroutinesApi::class)
 class CreateClientTest {
-
     @Test
     fun testMissingRegion() {
-        assertFailsWith<IllegalArgumentException> {
-            S3Client { }
-        }.message.shouldContain("region is a required configuration property")
+        // Should _not_ throw an exception since region is optional
```
aajtodd, line 21: **question**: What is the behavior now if region isn't set? Should we have a test case for that?

aajtodd (commented):
It's unclear to me if making region optional in the credential providers is correct or not. At what point when invoking an operation do we consider the lack of region being set to be an error?

ianbotsf:
> It's unclear to me if making region optional in the credential providers is correct or not. At what point when invoking an operation do we consider the lack of region being set to be an error?

As written, we will now consider it to be an error when resolving an endpoint during operation invocation. For example, the S3 client will throw an error in the endpoint provider: `EndpointProviderException: A region must be set when sending requests to S3.`.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
ianbotsf: It's now optional in the underlying service clients (i.e., the STS client) so I saw no reason to keep it required in the credentials provider. I'd assumed that STS would default to the global endpoint but I now see that's not the case.

If we leave this as optional in the profile creds provider and it attempts to make a call to STS which fails because of missing region, then the provider will throw an exception. If its used as part of a chain (e.g., the default chain), then that exception will cause the the next provider in the list to be called. I think that's correct since the next provide may not require a region.

aajtodd (approved):
Some CI failing but otherwise looks ok

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/core/AwsHttpProtocolClientGenerator.kt
```diff
@@ -76,30 +78,38 @@ open class AwsHttpProtocolClientGenerator(
     private fun renderMergeServiceDefaults(writer: KotlinWriter) {
         // FIXME - we likely need a way to let customizations modify/override this
         // FIXME - we also need a way to tie in config properties added via integrations that need to influence the context
```
lauzadis, line 80: Are these two FIXMEs solved with interceptors?

ianbotsf: > `// FIXME - we likely need a way to let customizations modify/override this`

I'm unclear what "customization" means in this context. Certainly _interceptors_ can be used to modify the execution context.

> `// FIXME - we also need a way to tie in config properties added via integrations that need to influence the context`

I wouldn't call this fixed by interceptors. Integrations could certainly add an interceptor but it seems like a boilerplate workaround. There's no common solution for it here yet, which I think is what the `FIXME` is about.

Issue #583, opened 2022-04-13
Make user-supplied region available to config resolution providers


ianbotsf:
### Describe the feature

There are several places where config resolution providers require bootstrapping values which are possibly present in user-supplied client config but those providers are unaware of and unable to access said config.

One example is **region** for the **profile credentials provider**. Profile credentials provider uses region in the case of assuming an STS web identity but it relies on the default region provider chain or an explicitly-passed region and _does not_ reuse any supplied value in client config:

```kotlin
val s3 = S3Client.fromEnvironment {
    region = "us-west-2"
    credentialsProvider = ProfileCredentialsProvider() // Won't respect region on prev line
}
```

~~Another example is **HTTP client engine** for the **ECS credentials provider**. ECS credentials provider uses a client engine but it relies on the default engine or an explicitly-passed region and _does not_ reuse any supplied value in client config:~~ (This is now fixed.)

```kotlin
val s3 = S3Client.fromEnvironment {
    region = "us-west-2"
    httpClientEngine = FooEngine()
    credentialsProvider = EcsCredentialsProvider(Platform) // Won't respect engine on prev line
}
```

### Is your Feature Request related to a problem?

Some configuration values require re-specification multiple times when it should only be required once. For example, the region problem above can be solved by specifying region twice:

```kotlin
val s3 = S3Client.fromEnvironment {
    region = "us-west-2"
    credentialsProvider = ProfileCredentialsProvider(region = "us-west-2")
}
```

This is not a user-friendly pattern.

### Proposed Solution

Find some way to make config resolution providers (e.g., credentials providers) aware of user-supplied client config either directly or indirectly.

### Acknowledge

- [ ] I may be able to implement this feature request

### AWS Kotlin SDK version used

0.14.2-beta

### Platform (JVM/JS/Native)

n/a

### Operating System and version

n/a

PR #997, opened 2023-07-26
feat: make user-supplied region available to credentials providers

0marperez:
## Issue \#
[583](https://github.com/awslabs/aws-sdk-kotlin/issues/583) - Make user-supplied region available to config resolution providers

## Description of changes

There are several places where config resolution providers require bootstrapping region, which is possibly present in user-supplied client config. This fixes this issue by making region available via attributes/execution context 

Also look at: https://github.com/awslabs/smithy-kotlin/pull/906

### DIFF
--- a/aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProviderTest.kt
+++ b/aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProviderTest.kt
@@ -5,11 +5,13 @@
 
 package aws.sdk.kotlin.runtime.auth.credentials
 
+import aws.sdk.kotlin.runtime.client.AwsClientOption
 import aws.smithy.kotlin.runtime.auth.awscredentials.Credentials
 import aws.smithy.kotlin.runtime.httptest.TestConnection
 import aws.smithy.kotlin.runtime.httptest.buildTestConnection
 import aws.smithy.kotlin.runtime.net.Host
 import aws.smithy.kotlin.runtime.util.TestPlatformProvider
+import aws.smithy.kotlin.runtime.util.attributesOf
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.test.runTest
 import kotlin.test.Test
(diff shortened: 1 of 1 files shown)


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
```diff
@@ -102,7 +99,7 @@ public class ProfileCredentialsProvider(
 
         // if profile is overridden for this provider, attempt to resolve it from there first
         val profileOverride = profileName?.let { sharedConfig.profiles[it] }
-        val region = asyncLazy { region ?: profileOverride?.getOrNull("region") ?: resolveRegion(platformProvider) }
+        val region = asyncLazy { region ?: profileOverride?.getOrNull("region") ?: attributes.getOrNull(AttributeKey("region")) ?: resolveRegion(platformProvider) }
```
ianbotsf, line 102: **Question**: Does this work? I'd expected we'd need the `"AwsRegion"` attribute, which is already defined as [`AwsClientOption.Region`](https://github.com/awslabs/aws-sdk-kotlin/blob/74c3a2d49632d6ac97177078d85829846f4bf11d/aws-runtime/aws-core/common/src/aws/sdk/kotlin/runtime/client/AwsClientOption.kt#L19).

aajtodd: Yes we need to use the existing attribute, we should _almost_ never inline an attribute key like this.

aajtodd (changes requested):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
aajtodd: Also we should have a new unit test for this

aajtodd (approved):

lauzadis (approved):

PR #1007, opened 2023-07-31
chore: update kotlin to 1.8.22; kotlinx.coroutines to 1.7.3

aajtodd:
## Issue \#

N/A

## Description of changes

* Update Kotlin to 1.8.22
* Update kotlinx.coroutines to 1.7.3

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,13 +6,13 @@ kotlin.native.ignoreDisabledTargets=true
 org.gradle.jvmargs=-Xmx6g -XX:MaxMetaspaceSize=2G
 
 # sdk
-sdkVersion=0.29.2-SNAPSHOT
+sdkVersion=0.30.0-SNAPSHOT
 
 # codegen
 smithyVersion=1.29.0
 smithyGradleVersion=0.6.0
 # smithy-kotlin codegen and runtime are versioned together
-smithyKotlinVersion=0.24.0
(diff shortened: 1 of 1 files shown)


0marperez (approved):

lauzadis (approved):
1.8.22 is not available internally (latest is 1.8.21). Will that affect internal usage of the SDK?

aajtodd:
@lauzadis 

> 1.8.22 is not available internally (latest is 1.8.21). Will that affect internal usage of the SDK?

Looking at the [release notes](https://github.com/JetBrains/kotlin/releases/tag/v1.8.22) I don't think so.

aws-runtime/build.gradle.kts
```diff
@@ -86,28 +86,6 @@ subprojects {
     }
 }
 
-task<org.jetbrains.kotlin.gradle.testing.internal.KotlinTestReport>("rootAllTest") {
```
antohaby, line 89: Hey! Kotlin Gradle Plugin developer here. 
Can you guys confirm that you don't need this code anymore? 
We have this bug that we want to close as "Won't Fix". But instead we want to improve UX here. 
Would you be so kind sharing your Use Case? 

Thanks

lauzadis: Hi, the code you are commenting on was deleted in this PR almost 1 year ago, and we never added it back, so it's safe to say we don't need it anymore.

PR #1021, opened 2023-08-16
misc: cleanup miscellaneous TODOs and FIXMEs

ianbotsf:
## Issue \#

Closes https://github.com/awslabs/smithy-kotlin/issues/191

## Description of changes

While auditing the outstanding TODOs and FIXMEs in **aws-sdk-kotlin**, I found a handful with easy fixes.

### DIFF
--- a/aws-runtime/aws-config/build.gradle.kts
+++ b/aws-runtime/aws-config/build.gradle.kts
@@ -62,6 +62,9 @@ kotlin {
 
                 // atomics
                 implementation("org.jetbrains.kotlinx:atomicfu:$atomicFuVersion")
+
+                // coroutines
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion")
             }
         }
         commonTest {


aws-runtime/build.gradle.kts
```diff
@@ -37,12 +36,6 @@ subprojects {
 
         sourceSets {
             // dependencies available for all subprojects
-            named("commonMain") {
-                dependencies {
-                    // FIXME - refactor to only projects that need this
-                    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion")
```
aajtodd, line 43: **correctness**: This is removed with no corresponding build where it is added. That seems wrong given that we definitely use coroutine core APIs _somewhere_ in the runtime. Which means we are using an API dependency from a smithy-kotlin module most likely. I think we should grep for the `kotlinx.coroutines` namespace and add the dependency. We should be explicit with our dependencies even if they can be picked up transitively.

aajtodd (approved):

lauzadis (approved):

Issue #693, opened 2022-09-02
Add Support for Kinesis SubscribeToShard


aajtodd:
### Describe the feature

Add support for Kinesis [SubscribeToShard](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_SubscribeToShard.html) event stream operation.

This requires adding support for RPC protocols where the initial request/response comes from the event stream rather than the HTTP request/response.

See also https://github.com/awslabs/aws-sdk-kotlin/issues/543

### Is your Feature Request related to a problem?

N/A

### Proposed Solution

N/A

### Describe alternative solutions or features you've considered

N/A

### Acknowledge

- [X] I may be able to implement this feature request

### AWS Kotlin SDK version used

N/A

### Platform (JVM/JS/Native)

N/A

### Operating System and version

N/A

PR #1026, opened 2023-08-18
feat: support `initial-request` and `initial-response` event stream types for RPC-bound protocols 

lauzadis:
This PR adds support for sending `initial-request` and receiving `initial-response` messages in event streams on RPC-bound protocols.

This functionality was previously only supported for HTTP-bound protocols because the initial requests and responses were sent/received in response headers. RPC-bound protocols do not rely on headers and instead encode the initial requests / responses in the event stream itself.  

## Issue \#

Closes [#693](https://github.com/awslabs/aws-sdk-kotlin/issues/693)

## Description of changes

- Render `initial-request` event streams serializers for RPC-bound protocols which also have initial request members 
- Render `initial-response` event streams deserializers for RPC-bound protocols that have initial response members

### DIFF
--- a/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/eventstream/EventStreamParserGenerator.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/eventstream/EventStreamParserGenerator.kt
@@ -83,7 +83,7 @@ class EventStreamParserGenerator(
 
         writer.write("val chan = body.#T() ?: return", RuntimeTypes.Http.toSdkByteReadChannel)
         writer.write("val frames = #T(chan)", RuntimeTypes.AwsEventStream.decodeFrames)
-        if (ctx.protocol.isRpcBoundProtocol && output.initialResponseMembers.isNotEmpty()) {
+        if (ctx.protocol.isRpcBoundProtocol) {
             renderDeserializeInitialResponse(ctx, output, writer)
         } else {
             writer.write("val events = frames")


codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/eventstream/EventStreamParserGenerator.kt
```diff
@@ -60,26 +62,31 @@ class EventStreamParserGenerator(
...
                         streamShape.filterEventStreamErrors(ctx.model).forEach { member ->
                             withBlock("#S -> {", "}", member.memberName) {
                                 renderDeserializeEventVariant(ctx, streamSymbol, member, writer)
                             }
                         }
+
+                        if (ctx.protocol.isRpcBoundProtocol && output.initialResponseMembers.isNotEmpty()) {
+                            withBlock("\"initial-response\" -> {", "}") {
+                                renderDeserializeInitialResponse(ctx, output, writer)
+                                write("#T.SdkUnknown", streamSymbol)
```
lauzadis, line 83: Discussion point: Here I'm emitting `SdkUnknown` after parsing the `initial-response` message so the presence of an initial-response can be detected. If a response type's initial members are *all* optional, it's possible a service won't send an `initial-response` message and checking for `SdkUnknown` on L123 handles that possibility.

The question is: should I generate a custom type instead of using `SdkUnknown`? Something like `#T.InitialResponse`?

tests/codegen/event-stream/event-stream-initial-request-response.smithy
lauzadis, line 1: I had to make a new service model because the existing one uses `restJson1` which is HTTP-bound. This one uses `awsJson1_1` which is RPC-bound

tests/codegen/event-stream/build.gradle.kts
```diff
@@ -19,24 +19,28 @@ dependencies {
...
 
 val tests = listOf(
     EventStreamTest("restJson1", "restJson1"),
+    EventStreamTest("awsJson11", "awsJson1_1", file("event-stream-initial-request-response.smithy")),
 )
 
-fun fillInModel(output: File, protocolName: String) {
-    val template = file("event-stream-model-template.smithy")
+fun fillInModel(output: File, protocolName: String, template: File?) {
+    val template = template ?: file("event-stream-model-template.smithy")
```
aajtodd, line 34: Maybe we just remove the template here and simplify this since we don't have plans to add other protocols right now and we don't seem to be getting anything out of it.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/eventstream/EventStreamParserGenerator.kt
```diff
@@ -167,6 +195,32 @@ class EventStreamParserGenerator(
...
+    private fun renderDeserializeInitialResponse(ctx: ProtocolGenerator.GenerationContext, outputShape: StructureShape, writer: KotlinWriter) {
+        if (outputShape.initialResponseMembers.isNotEmpty()) {
+            // A custom function which only deserializes the initial response members
+            val initialResponseDeserializeFn = sdg.payloadDeserializer(ctx, outputShape, outputShape.initialResponseMembers)
+
+            // Deserialize into `initialResponse`, then apply it to the actual response builder
+            writer.write("val initialResponse = #T(message.payload)", initialResponseDeserializeFn)
+            writer.withBlock("builder.apply {", "}") {
+                outputShape.initialResponseMembers.forEach { member ->
+                    writer.write("#L = initialResponse.#L", member.defaultName(), member.defaultName())
```
aajtodd, line 210: **nit**: You can use numbered formatters e.g. `write("#1L = initialResponse.#1L", member.defaultName())`

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/eventstream/EventStreamParserGenerator.kt
aajtodd: What if we get an unknown event type as the first event message? That will show up as `SdkUnknown` as well and we would be dropping it correct?

lauzadis: Yes, we would be dropping it in that case. It's unlikely but possible. That's why I'm thinking to generate some kind of `InitialResponse` type for RPC-bound, initial-response-containing event streams only. 

Does that seem correct, or is there a better way to handle this?

aajtodd: Well the issue is that it has to be part of the overall sealed class hierarchy right? We don't want to expose an implementation detail like that I wouldn't think. We could probably generate something that is internal only that wraps the two possibilities.

e.g. (just spit balling here)

```kotlin
internal sealed class RpcEventStream<T> {
    // sentinel value indicating we got an initial response
    object InitialResponse : EventStreamWrapperType() 

     // holds an actual decoded event stream type for the public event stream shape
    data class StreamVariant(val value: T): EventStreamWrapperType()
}
```

Haven't thought too much about it yet 🤔

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/eventstream/EventStreamSerializerGenerator.kt
```diff
@@ -76,10 +78,28 @@ class EventStreamSerializerGenerator(
...
+        if (ctx.protocol.isRpcBoundProtocol && initialRequestMembers.isNotEmpty()) {
+            val serializerFn = sdg.payloadSerializer(ctx, input, initialRequestMembers)
+
+            writer.withBlock("val initialRequest = buildMessage {", "}") {
+                writer.write("addHeader(\":message-type\", HeaderValue.String(\"event\"))")
+                writer.write("addHeader(\":event-type\", HeaderValue.String(\"initial-request\"))")
+                writer.write("payload = #T(input)", serializerFn)
+            }
+
+            writer.withBlock("val messages = #T(#T(initialRequest), stream.#T(::#T))", "", RuntimeTypes.KotlinxCoroutines.Flow.merge, RuntimeTypes.KotlinxCoroutines.Flow.flowOf, RuntimeTypes.KotlinxCoroutines.Flow.map, encodeFn) {
```
ianbotsf, line 93: **Nit**: Line length. (How did **ktlint** not catch this?)

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/eventstream/EventStreamSerializerGenerator.kt
```diff
@@ -195,4 +215,23 @@ class EventStreamSerializerGenerator(
...
+    private val ShapeId.isRpcBoundProtocol: Boolean
+        get() = run {
+            val rpcBoundProtocols = setOf(
+                "awsJson1_0",
+                "awsJson1_1",
+                "awsQuery",
+                "ec2Query",
+            )
+            return rpcBoundProtocols.contains(name)
+        }
```
ianbotsf, line 236: **Nit**: Duplicated.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/eventstream/EventStreamParserGenerator.kt
lauzadis: > Couldn't we change the frame transformation into `mapNotNull` and return `null` from this branch?

My implementation did this at some point, but it does not work for a few reasons.

1) We need to expose the initial-response to users *before* the event stream is collected. Using `mapNotNull` and returning `null` in the branch means that `.take(1)` (preprocessing the initial-response) will consume both the initial-response (which is null) as well as the first non-null message in the actual event stream.

2) It's not possible to determine if we've received an initial-response or not. A service can skip sending an initial-response if all the members are optional, and the SDK needs to detect when that happens to avoid reading the first *actual* event stream message. I'm using `SdkUnknown` here as a flag indicating we've received an initial-response.

lauzadis: We spoke offline, I'll try refactoring this to avoid emitting any type at all by processing the first message off `decodeFrames` before entering the `map` block

lauzadis: Refactored. Here is a sample generated deserializer:
```kt
    val chan = body.toSdkByteReadChannel() ?: return
    val frames = decodeFrames(chan)
    val firstMessage = frames.take(1).single()
    val firstMessageType = firstMessage.type()
    val events = if (firstMessageType is MessageType.Event && firstMessageType.shapeType == "initial-response") {
        val initialResponse = deserializeTestStreamOperationWithInitialRequestResponseResponsePayloadb9carVwo(firstMessage.payload)
        builder.apply {
            initial = initialResponse.initial
        }
        frames
    } else {
        merge(flowOf(firstMessage), frames)
    }
        .map { message ->
            when (val mt = message.type()) {
                is MessageType.Event -> when (mt.shapeType) {
                    "MessageWithString" -> {
                        val eb = MessageWithString.Builder()
                        eb.data = message.payload.decodeToString()
                        val e = eb.build()
                        TestStream.MessageWithString(e)
                    }
                    else -> TestStream.SdkUnknown
                }
                is MessageType.Exception -> when (mt.shapeType) {
                    "SomeError" -> {
                        val err = deserializeSomeErrorPayload(message.payload)
                        throw err
                    }
                    else -> throw TestException("error processing event stream, unrecognized errorType: ${mt.shapeType}")
                }
                is MessageType.Error -> throw TestException("error processing event stream: errorCode=${mt.errorCode}; message=${mt.message}")
                is MessageType.SdkUnknown -> throw ClientException("unrecognized event stream message `:message-type`: ${mt.messageType}")
            }
        }

    builder.value = events
```

ianbotsf: Seems better from a type perspective.

To confirm, if there _is no_ initial message, this code will still pull the first message off the wire, waiting if necessary, right? This means that `builder.value` won't be set (and thus `deserializeFooOperationBody` won't return, etc. up the stack) until the first message becomes available?

lauzadis: Yes that's right, it's unfortunate but seems like the best we can do with a `Flow`. Note that this is only code-generated for RPC-bound protocols that could send an initial-response, it's very tightly scoped.

ianbotsf (approved):

aajtodd (approved):
We might consider (in a follow up PR) adding an e2e test for kinesis subscribe to shard. We have e2e tests for several of the other event stream operations and it would make sense to cover this one if it's not too much work since it's the only RPC based protocol with an event stream.

PR #1032, opened 2023-08-24
fix: merging Flows

lauzadis:
`merge` is non deterministic and can make messages appear in the wrong order.
> Merges the given flows into a single flow **without preserving an order** of elements. 

This PR adds a new `mergeSequential` which enforces order by emitting each flow one-by-one. 

## Issue \#

N/A

## Description of changes

This change is required to fix a bug in event streams serializers and deserializers. 

### DIFF
--- a/aws-runtime/aws-core/common/test/FlowUtilTest.kt
+++ b/aws-runtime/aws-core/common/test/FlowUtilTest.kt
@@ -1,3 +1,7 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
 import aws.sdk.kotlin.runtime.InternalSdkApi
 import aws.sdk.kotlin.runtime.mergeSequential
 import kotlinx.coroutines.flow.Flow


ianbotsf (approved):

aws-runtime/aws-core/api/aws-core.api
```diff
@@ -26,6 +26,9 @@ public class aws/sdk/kotlin/runtime/ConfigurationException : aws/sdk/kotlin/runt
 	public fun <init> (Ljava/lang/Throwable;)V
 }
 
+public final class aws/sdk/kotlin/runtime/FlowUtilKt {
```
aajtodd, line 29: seems like we are missing public visibility marker/config like we have for smithy-kotlin [InternalApi](https://github.com/awslabs/smithy-kotlin/blob/main/build.gradle.kts#L110)

aws-runtime/aws-core/common/src/aws/sdk/kotlin/runtime/FlowUtil.kt
```diff
@@ -0,0 +1,15 @@
...
+ */
+package aws.sdk.kotlin.runtime
+
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.emitAll
+import kotlinx.coroutines.flow.flow
+
+// FIXME relocate to smithy-kotlin
+@InternalSdkApi
+public fun <T> mergeSequential(vararg flows: Flow<T>): Flow<T> = flow {
```
aajtodd, line 13: tests?

aajtodd (approved):
Approved with suggestions

aws-runtime/aws-core/api/aws-core.api
lauzadis: The marker works (because `mergeSequential` doesn't appear in this API dump) but the binary compatibility validator tool still includes the file as a wrapper for some reason, and it's not configurable by us.

PR #1035, opened 2023-08-29
feat: add more parameters for fetching STS credentials

### LINKED ISSUE #1000: StsAssumeRoleCredentialsProvider is missing sourceId and policy, preventing scoped down policies
### Describe the bug

Converting Java SDK code to Kotlin SDK; using StsAssumeRoleCredentialsProvider to provide scoped-down policies for SaaS application, where there is a StsAssumeRoleCredentialsProvider per-tenant that has a dynamic policy limiting access to just that tenants resources (for example, limiting S3 access to /<tenant-id/*).

Kotlin SDK StsAssumeRoleCredentialsProvider does not allow for `policy` attribute for `assumeRole` call, nor does it allow for `sourceId` attribute (which is useful to separate application-assume-role calls from other assume-role calls that you may not want to permit, such as a compromised system making calls from the CLI).

### Expected behavior

StsAssumeRoleCredentialsProvider supports all the capabilities of `assumeRole` call.

### Current behavior

As described earlier, StsAssumeRoleCredentialsProvider is missing `policy` and `sourceId` attributes for `assumeRole` call.

### Steps to Reproduce

n/a

### Possible Solution

Workaround: Given the small size of StsAssumeRoleCredentialsProvider create a custom one that offer the full capabilities.

Fix: much like the Java SDK provide a callback for full customization of the assume-role request

### Context

Unable to directly migrate from Java SDK to Kotlin SDK without building glue code (unfortunate for two missing API attributes).

0.29.0-beta

### Platform (JVM/JS/Native)

JVM

MacOs

ianbotsf:
## Issue \#

Closes #1000 

## Description of changes

This change adds new constructor overloads to `StsAssumeRoleCredentialsProvider` and `StsWebIdentityCredentialsProvider` to allow passing more API parameters to the underlying STS operations.

### DIFF
--- a/aws-runtime/aws-config/api/aws-config.api
+++ b/aws-runtime/aws-config/api/aws-config.api
@@ -161,18 +161,6 @@ public final class aws/sdk/kotlin/runtime/auth/credentials/StsWebIdentityCredent
 	public static synthetic fun fromEnvironment-TUY-ock$default (Laws/sdk/kotlin/runtime/auth/credentials/StsWebIdentityCredentialsProvider$Companion;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;JLaws/smithy/kotlin/runtime/util/PlatformProvider;Laws/smithy/kotlin/runtime/http/engine/HttpClientEngine;ILjava/lang/Object;)Laws/sdk/kotlin/runtime/auth/credentials/StsWebIdentityCredentialsProvider;
 }
 
-public final class aws/sdk/kotlin/runtime/auth/credentials/WebIdentityParameters {
-	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;JLjava/lang/String;Ljava/util/List;Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
-	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;JLjava/lang/String;Ljava/util/List;Ljava/lang/String;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
-	public final fun getDuration-UwyO8pc ()J
-	public final fun getPolicy ()Ljava/lang/String;
-	public final fun getPolicyArns ()Ljava/util/List;
-	public final fun getProviderId ()Ljava/lang/String;
-	public final fun getRoleArn ()Ljava/lang/String;
-	public final fun getRoleSessionName ()Ljava/lang/String;
-	public final fun getWebIdentityTokenFilePath ()Ljava/lang/String;
-}
-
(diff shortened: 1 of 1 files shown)


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/StsAssumeRoleCredentialsProvider.kt
```diff
@@ -107,6 +147,44 @@ public class StsAssumeRoleCredentialsProvider(
...
+    public val policyArns: List<String>? = null,
+    public val policy: String? = null,
+    public val tags: Map<String, String>? = null,
+    public val transitiveTagKeys: List<String>? = null,
+    public val serialNumber: String? = null,
+    public val tokenCode: String? = null,
+    public val sourceIdentity: String? = null,
+) {
+    internal val convertedPolicyArns = policyArns?.map { PolicyDescriptorType { arn = it } }
+    internal val convertedTags = tags?.map { Tag { key = it.key; value = it.value } }
```
aajtodd, line 185: Nice! This is exactly the sort of paper cut I was talking about yesterday where we can make things more idiomatic/easier for customers over just generating the model shape.

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/StsAssumeRoleCredentialsProviderTest.kt
```diff
@@ -46,6 +46,26 @@ class StsAssumeRoleCredentialsProviderTest {
         assertEquals(StsTestUtils.expectedCredentialsBase, actual)
     }
 
+    @Test
+    fun testSuccessWithAdditionalParams() = runTest {
```
aajtodd, line 50: **nit**: This doesn't test that the additional parameters are actually plumbed all the way to the assume role invocation. I think thats probably ok but it would make for a better test if it's easy enough to do.

aajtodd (approved):
Looks great overall. Nice improvements.

lauzadis (approved):
Nice! I agree with Aaron about the naming of `WebIdentityProviders`, it would be safer to be more verbose

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/StsAssumeRoleCredentialsProvider.kt
ianbotsf: Yeah, I hadn't really considered the transitive closure of shapes when contemplating making `AssumeRoleRequest` public. I think this is still a problem we'll eventually want to solve in codegen for other use cases but, for parts of our API where we have more ownership (like creds providers), I think I've come around to your point.

aajtodd (approved):

PR #1045, opened 2023-09-13
misc: generate internal-only clients with `internal` visibility

lauzadis:
This PR applies `internal` visibility build settings to the internal-only generated SDK clients. It updates the .api file to remove them from the public API.

This is technically breaking, but since the classes were all already labeled as "internal", it won't be treated as a breaking change. 

## Issue \#

N/A

## Description of changes

This change is required to remove the internal-only clients from the public API.

### DIFF
--- a/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/core/AwsHttpProtocolClientGenerator.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/core/AwsHttpProtocolClientGenerator.kt
@@ -36,7 +36,7 @@ open class AwsHttpProtocolClientGenerator(
         writer.write("\n\n")
         writer.write(
             "#L const val ServiceApiVersion: String = #S",
-            ctx.settings.api.visibility.value,
+            ctx.settings.api.visibility,
             ctx.service.version,
         )
         writer.write("\n\n")


aws-runtime/aws-config/api/aws-config.api
```diff
@@ -205,122 +205,6 @@ public final class aws/sdk/kotlin/runtime/auth/credentials/internal/ManagedBeare
...
-public abstract interface class aws/sdk/kotlin/runtime/auth/credentials/internal/sso/auth/SsoAuthSchemeProvider : aws/smithy/kotlin/runtime/auth/AuthSchemeProvider {
-}
-
-public final class aws/sdk/kotlin/runtime/auth/credentials/internal/sso/endpoints/DefaultSsoEndpointProvider : aws/sdk/kotlin/runtime/auth/credentials/internal/sso/endpoints/SsoEndpointProvider {
-	public fun <init> ()V
-	public fun resolveEndpoint (Laws/sdk/kotlin/runtime/auth/credentials/internal/sso/endpoints/SsoEndpointParameters;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
-	public synthetic fun resolveEndpoint (Ljava/lang/Object;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
-}
-
 public final class aws/sdk/kotlin/runtime/auth/credentials/internal/sso/endpoints/SsoEndpointParameters {
```
lauzadis, line 208: I couldn't get the endpoints to generate with `internal` because the way it's implemented. 

All generated endpoint providers implement the `public` [EndpointProvider](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/smithy-client/common/src/aws/smithy/kotlin/runtime/client/endpoints/EndpointProvider.kt#L11), which means they can't be generated as `internal` (because then the `public` implementation would "expose" the `internal` one).

I can't see a simple way to work around this without changes in the endpoints implementation.

lauzadis: The specific error is `'public' sub-interface exposes its 'internal' supertype argument SsoEndpointParameters`

ianbotsf (approved):

aajtodd (approved):

PR #1059, opened 2023-10-02
feat(codegen): improve nullability of generated types

aajtodd:
## Issue \#

upstream: https://github.com/awslabs/smithy-kotlin/pull/968

## Description of changes

* **refactor**: Remove `BoxServices` preprocessor since `@box` doesn't affect nullability anymore. This is OBE for all services except EC2 at this point (all API GW service models have been fixed). Updated preprocessor to be EC2 specific. and to actually check nullability rather than just trait presence. 
* **fix**: Update error correction tests to assume struct members are nullable. The default check mode is `CLIENT_CAREFUL`, only `CLIENT` mode would allow for non-null structure members.


codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/ec2/EC2MakePrimitivesOptional.kt
```diff
@@ -0,0 +1,35 @@
...
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean =
+        settings.service == ShapeId.from("com.amazonaws.ec2#AmazonEC2")
+
+    override fun preprocessModel(model: Model, settings: KotlinSettings): Model {
+        val updates = mutableListOf<Shape>()
+        for (struct in model.structureShapes) {
+            for (member in struct.allMembers.values) {
+                updates.add(member.toBuilder().addTrait(ClientOptionalTrait()).build())
+            }
+        }
```
ianbotsf, line 32: **Style**: Could be simplified using list transformations:

```kotlin
val updates = model
    .structureShapes
    .flatMap { it.allMembers.values }
    .map { it.toBuilder().addTrait(ClientOptionalTrait()).build() }
```

ianbotsf (approved):

lauzadis (approved):

PR #1061, opened 2023-10-04
chore: use a version catalog

aajtodd:
## Issue \#

n/a

## Description of changes

Refactor build files to use a version catalog. 

Also made one minor update to allow bootstrap to not require `+` prefix which has annoyed the crap out of me forever. You can now just do :

```
/gradlew :codegen:sdk:bootstrap -Paws.services=s3,sts
```

instead of needing `+s3,+sts` which seems entirely redundant. I've left the logic for handling `+` in place so none of our build/release infrastructure breaks (or anyone's local development setup that may rely on it).

### DIFF
--- a/codegen/protocol-tests/build.gradle.kts
+++ b/codegen/protocol-tests/build.gradle.kts
@@ -78,7 +78,7 @@ tasks.named<SmithyBuild>("generateSmithyProjections") {
     addCompileClasspath = true
 
     // ensure the generated clients use the same version of the runtime as the aws aws-runtime
-    val smithyKotlinVersion: String by project
+    val smithyKotlinVersion = libs.versions.smithy.kotlin.version.get()
     doFirst {
         System.setProperty("smithy.kotlin.codegen.clientRuntimeVersion", smithyKotlinVersion)
     }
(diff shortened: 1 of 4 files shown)


gradle/sdk-plugins/src/main/kotlin/aws/sdk/kotlin/gradle/codegen/tasks/SmithyTasks.kt
```diff
@@ -112,10 +113,13 @@ private fun Project.createCodegenConfiguration(): Configuration {
 internal fun Project.createSmithyCliConfiguration(): Configuration {
     // see: https://github.com/awslabs/smithy-gradle-plugin/blob/main/src/main/java/software/amazon/smithy/gradle/SmithyPlugin.java#L119
     val smithyCliConfig = configurations.maybeCreate("smithyCli")
+
+    // FIXME - this is tightly coupled to our project
```
lauzadis, line 117: I don't understand what's FIXME about this

lauzadis (approved):

gradle/sdk-plugins/src/main/kotlin/aws/sdk/kotlin/gradle/codegen/tasks/SmithyTasks.kt
aajtodd: I'd like to be able to pull out the codegen plugin logic here to make it re-usable in say `smithy-kotlin`

PR #1069, opened 2023-10-09
chore: add release checks

aajtodd:
## Issue \#

n/a

upstream: https://github.com/awslabs/aws-kotlin-repo-tools/pull/17

## Description of changes

Adds additional CI checks to try and detect if we depend on unreleased changes. Previously we released the SDK and `smithy-kotlin` together as needed. Automated daily releases requires the repository _always_ be in a releasable state. This means that features that span repos and depend on corresponding `smithy-kotlin` changes will not be able to be merged to `main` until they've made it into an official `smithy-kotlin` release.

Additionally I've refactored the overall GH workflows a bit:
* Added a new composite build to setup our dependencies at the correct versions. This relies on https://github.com/awslabs/aws-kotlin-repo-tools/pull/17.
* remove dependency on CRT builder (at least for GH actions). Using the new re-usable composite action we can fetch the correct version of `smithy-kotlin` and just use it as an included build rather than go through CRT builder to do all this.
* Restructure our CI to build with JDK 17 but to run tests on many JVM versions. The new release check doesn't use the HEAD version of `smithy-kotlin` and the released maven version is built requiring JDK 17. In practice that is our actual min JDK version to build codegen right now. JDK 17 is what we build and release with.
* Improved the ability to cancel tests in-flight. Often I'll forget something after creating a PR or making a change that requires pushing another commit. I've added concurrency of 1 to most workflows such that previous builds are cancelled and new ones started with the latest commit (rather than needlessly running to completion).
    * There is a corresponding CR to grant the `StopBuild` permission as well as a ticket opened against https://github.com/aws-actions/aws-codebuild-run-build/issues/145

NOTE: If you locally use < JDK 17 you'll need to update to JDK 17 after this lands.

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -10,7 +10,7 @@ runs:
       with:
         path: 'aws-kotlin-repo-tools'
         repository: 'awslabs/aws-kotlin-repo-tools'
-        ref: 'gh-action'
+        ref: '0.2.3'
         sparse-checkout: |
           .github
     - name: Checkout smithy-kotlin
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -46,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, macos-latest, windows-2019 ]
+        os: [ ubuntu-latest, macos-latest, windows-latest ]
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4


.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,27 @@
...
+        repository: 'awslabs/aws-kotlin-repo-tools'
+        ref: 'gh-action'
+        sparse-checkout: |
+          .github
+    - name: Checkout smithy-kotlin
+      uses: ./aws-kotlin-repo-tools/.github/actions/checkout-head
+      with:
+        # checkout smithy-kotlin as a sibling which will automatically make it an included build
+        path: 'smithy-kotlin'
+        repository: 'awslabs/smithy-kotlin'
```
ianbotsf, line 21: **Question**: Do we also need to checkout **aws-crt-kotlin**?

ianbotsf (approved):

ianbotsf (commented):
**Question**: Do we also need a version of this PR for **smithy-kotlin** to only used released versions of **aws-crt-kotlin**?

.github/workflows/continuous-integration.yml
```diff
@@ -1,83 +1,79 @@
...
+          ./gradlew -Ptest.java.version=${{ matrix.java-version }} jvmTest --stacktrace
 
-  windows-compat:
-    runs-on: windows-2019
+  all-platforms:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-latest, macos-latest, windows-2019 ]
```
ianbotsf, line 49: **Question**: I notice that the **smithy-kotlin** PR uses `windows-latest` while this explicitly uses `windows-2019` (which is _not_ the latest). Should it be upgraded to `windows-latest` too?

aajtodd: I don't recall why this one is set to 2019, I can try windows-latest though and see what shakes out.

.github/actions/setup-build/action.yml
aajtodd: I think ideally this would be handled by improving this tooling. Right now we don't update `aws-crt-kotlin` often enough or with large enough changes that it's worth considering at the moment. I'm thinking we revisit this when it's a problem and we need it (likely as we pursue other KMP targets).

lauzadis (approved):

PR #1073, opened 2023-10-11
fix: add content type header to empty body rest xml payloads

### LINKED ISSUE #1050: Missing `Content-Type` on empty restXml union payloads
### Describe the bug

For APIs where the top-level request type is a union, sending an empty body (i.e., no union value selected) via restXml does not include the `Content-Type` header, which fails a [new Smithy protocol test](https://github.com/smithy-lang/smithy/pull/1984) and may cause failures in certain service backends.

### Expected behavior

Uncertain exactly. Likely `Content-Type` should be set on any restXml non-blob request body, even when the body is empty.

### Current behavior

The `RestXmlHttpPayloadWithUnsetUnion` Smithy protocol test fails.

### Steps to Reproduce

Run the protocol tests.

### Possible Solution

_No response_

### Context

_No response_

Latest

### Platform (JVM/JS/Native)

Doesn't matter

Doesn't matter

0marperez:
## Issue \#

closes #1050 
## Description of changes

-Empty body rest xml payloads will have the content type header added to them

Related to: 


ianbotsf (approved):

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/RestXml.kt
```diff
@@ -30,6 +30,21 @@ open class RestXml : AwsHttpBindingProtocolGenerator() {
     override val protocol: ShapeId = RestXmlTrait.ID
     override val defaultTimestampFormat: TimestampFormatTrait.Format = TimestampFormatTrait.Format.DATE_TIME
 
+    override fun renderContentTypeHeader(
```
aajtodd, line 33: **suggestion**: We have [other](https://github.com/awslabs/aws-sdk-kotlin/blob/852fe92843713191d4948efeaf610bba2a635e67/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/RestJson1.kt#L62) cases where we end up having to override when/how the content-type header is set. I suggest looking at the generated code for this case and any others and reconciling with the addition of `renderContentTypeHeader`.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/RestXml.kt
```diff
@@ -30,6 +30,21 @@ open class RestXml : AwsHttpBindingProtocolGenerator() {
...
+        ctx: ProtocolGenerator.GenerationContext,
+        op: OperationShape,
+        writer: KotlinWriter,
+    ) {
+        val resolver = getProtocolHttpBindingResolver(ctx.model, ctx.service)
+        val contentType = resolver.determineRequestContentType(op)
+
+        if (!op.isBlobShape) {
```
aajtodd, line 41: **correctness**: `op` is of type `OperationShape`, this expression is never true.

aajtodd (changes requested):

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/RestJson1.kt
```diff
@@ -59,11 +59,21 @@ class RestJson1 : JsonHttpBindingProtocolGenerator() {
                 }
                 // Content-Type still needs to be set for non-structured payloads
                 // https://github.com/awslabs/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/http-content-type.smithy#L174
-                write("builder.headers.setMissing(\"Content-Type\", #S)", resolver.determineRequestContentType(op))
+                renderContentTypeHeader(ctx, op, writer, resolver)
             }
         }
     }
 
+    override fun renderContentTypeHeader(
```
aajtodd, line 67: This isn't right, it's setting the content type header _always_ instead of predicating on whether the body is empty or not like it does in the base class. Looking at this closer I'd ignore my previous [suggestion](https://github.com/awslabs/aws-sdk-kotlin/pull/1073/files#r1356763978) and keep the inline customization to set content type on L62. I thought we could reconcile them but you'll just end up with repeated runtime checks.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/RestXml.kt
```diff
@@ -30,6 +30,18 @@ open class RestXml : AwsHttpBindingProtocolGenerator() {
     override val protocol: ShapeId = RestXmlTrait.ID
     override val defaultTimestampFormat: TimestampFormatTrait.Format = TimestampFormatTrait.Format.DATE_TIME
 
+    override fun renderContentTypeHeader(
+        ctx: ProtocolGenerator.GenerationContext,
+        op: OperationShape,
+        writer: KotlinWriter,
+        resolver: HttpBindingResolver,
+    ) {
+        val contentType = resolver.determineRequestContentType(op)
```
aajtodd, line 39: This is _always_ going to set the content type header when the failing test was restricted to union shapes. We should predicate this on that instead of overriding it for the entire protocol.

aajtodd (changes requested):

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/protocols/RestXml.kt
```diff
@@ -30,6 +33,18 @@ open class RestXml : AwsHttpBindingProtocolGenerator() {
...
     override val defaultTimestampFormat: TimestampFormatTrait.Format = TimestampFormatTrait.Format.DATE_TIME
 
+    // See: https://github.com/awslabs/aws-sdk-kotlin/issues/1050
+    override fun renderContentTypeHeader(
+        ctx: ProtocolGenerator.GenerationContext,
+        op: OperationShape,
+        writer: KotlinWriter,
+        resolver: HttpBindingResolver,
+    ) {
+        if (op.inputIsUnionShape(ctx.model)) {
```
aajtodd, line 43: **fix**: You have overridden this for the entire protocol but are now only handling the new case. You need to be calling `super.renderContentTypeHeader(...)` somewhere in this function as the default code path and only special case the union shape requirement for content type.

aajtodd (changes requested):

aajtodd (approved):

PR #1082, opened 2023-10-21
fix: add smithy-kotlin to our bom and re-export to version catalog

aajtodd:
## Issue \#

n/a

## Description of changes

https://github.com/awslabs/aws-sdk-kotlin/pull/1072 added a BOM and version catalog artifact. This PR adds `smithy-kotlin` to our BOM and re-exports `smithy-kotlin` runtime from the `aws-sdk-kotlin` version catalog for convenience.

e.g.

```kotlin
dependencies {
     implementation(awsSdk.runtime.smithy.kotlin.http.client.engine.crt)
     implementation(awsSdk.services.s3)
}
```

### Alternatives

`smithy-kotlin` also publishes a version catalog. Originally I had wanted to get rid of a redeclaring these in the SDK `libs.versions.toml` and just use that artifact directly for both the `aws-runtime` build files and in the BOM/version catalog from the SDK. Unfortunately version catalogs can only be created from `settings.gradle.kts` (with `libs.versions.toml` being created implicitly by convention). This doesn't play well with composite builds (you can't use the version catalog from an include build of `smithy-kotlin`). 

### DIFF
--- a/bom/build.gradle.kts
+++ b/bom/build.gradle.kts
@@ -53,19 +53,27 @@ fun createBomConstraintsAndVersionCatalog() {
         }
     }
 
+    val ignoredSmithyKotlin = setOf(
+        "smithy.kotlin.codegen",
+        "smithy.kotlin.http.test",
+        "smithy.kotlin.test",
+        "smithy.kotlin.smithy.test",
+        "smithy.kotlin.aws.signing.test",
+    )
 
     // add smithy-kotlin versions to our BOM and allow direct aliasing in the catalog
-    val smithyKotlinCatalog = extensions.getByType<VersionCatalogsExtension>().named("smithyKotlin")
     catalogExt.versionCatalog {
-        smithyKotlinCatalog.libraryAliases.forEach {  alias ->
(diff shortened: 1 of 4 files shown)


bom/build.gradle.kts
```diff
@@ -52,6 +52,28 @@ fun createBomConstraintsAndVersionCatalog() {
             }
         }
     }
+
+    val ignoredSmithyKotlin = setOf(
```
ianbotsf, line 56: **Question**: Is it worth ignoring based on patterns like "contains test" or "contains codegen"? It's less explicit but maybe more likely to prevent the accidental inclusion of some new module...

ianbotsf (approved):

bom/build.gradle.kts
ianbotsf: I like the pattern-based approach more personally. I think it's better to accidentally omit something and add it later than to accidentally include something and remove it later.

lauzadis: I could go either way but this explicit list seems fine given the low potential for impact

lauzadis (approved):

PR #1095, opened 2023-10-30
chore(ci): add workflow to update release branch

aajtodd:
## Issue \#

n/a

## Description of changes
Adds a new GH workflow that incorporates the release checks and merges changes from `main` to `release` and pushes them to the `release` branch.

This addresses a few things:
1. Updates to the `release` branch are fully automated and (when combined with some branch protection rules that restrict updates to the `release` branch to the CI user) only allowed by the CI user 
2. Removes the need to cut a PR to the `release` branch and have it reviewed and merged.
    * We only want to allow merge commits to the `release` branch, this ensures that (GH does not allow you to set merge strategy per/branch which means developers would have to get this right or they risk messing up the commit history between `main` and `release`). 
    * GH merge commits from PRs _always_ create a merge commit, going this way will also enable fast-forward merges to `release`. 

## Testing
I've tested this workflow to the extent we can. You can see workflow runs of what it would look like when:
* [merge isn't needed](https://github.com/awslabs/aws-sdk-kotlin/actions/runs/6696992500/job/18195917324)
* [merge is needed](https://github.com/awslabs/aws-sdk-kotlin/actions/runs/6696923192/job/18195700162)

I can't actually test a push to `release` so we'll find out on first run of this if it actually works or not.

### DIFF
--- a/.github/workflows/update-release-branch.yml
+++ b/.github/workflows/update-release-branch.yml
@@ -31,7 +31,7 @@ jobs:
         shell: bash
         run: |
           git config user.name aws-sdk-kotlin-ci 
-          git config user.email "aws-kotlin-sdk-automation@amazon.com>"
+          git config user.email "aws-kotlin-sdk-automation@amazon.com"
       - name: Configure JDK
         uses: actions/setup-java@v3
         with:


.github/workflows/update-release-branch.yml
```diff
@@ -0,0 +1,90 @@
...
+        run: |
+          echo "merging main @ $main_sha into release @ $release_sha";
+          git switch release;
+          input_message=${{ inputs.commit_message }}
+          message=${input_message:-"Merging main into release"}
+          echo "message=$message"
+          git merge -m "$message" main;
+          if [ "${{ inputs.dry_run }}" == "true" ]; then
+            echo "dry run, skipping push to remote";
+            git log -n 10 --oneline;
```
lauzadis, line 86: question: when would this `git log` output be useful?

.github/workflows/update-release-branch.yml
```diff
@@ -0,0 +1,90 @@
...
+      - uses: actions/checkout@v4
+        with:
+          ref: 'main'
+          fetch-depth: 0
+          token: ${{ secrets.CI_USER_PAT }}
+      - name: Configure Git
+        shell: bash
+        run: |
+          git config user.name aws-sdk-kotlin-ci 
+          git config user.email "aws-kotlin-sdk-automation@amazon.com>"
```
lauzadis, line 34: typo: unnecessary bracket `>`

0marperez (approved):

.github/workflows/update-release-branch.yml
```diff
@@ -0,0 +1,90 @@
...
+          fi
+      - name: Release Check - build
+        if: env.MERGE_NEEDED == 'true'
+        run: |
+          # Our CI is implemented as a "live at HEAD" model where we build against the latest of all our 1P deps (either
+          # main branch or matching branch name). Double check that without this "live at HEAD" mode we still build 
+          # successfully (which is how it is built during release).
+          # This should help prevent the cases where we forgot to bump smithy-kotlin versions and don't catch it
+          # because CI is masking it
+          ./gradlew -Paws.kotlin.native=false test jvmTest
```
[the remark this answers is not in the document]
aajtodd: We aren't using a composite build here, it's testing that the SDK builds on it's own with the versions given.

.github/workflows/update-release-branch.yml
aajtodd: If skipping is ever not expected or if we wanted to see what the history looked like from the point of view of the workflow. I'm not sure it will be but I'm not sure it won't be either. 🤷

ianbotsf (approved):

PR #1099, opened 2023-11-01
fix: add forcePathStyle and enableAccelerate config properties back to the S3 client

### LINKED ISSUE #1098: forcedPathStyle was needed and removed
### Describe the bug

Hi, in the latest version of the AWS SDK forcePathStyle has been removed (https://github.com/awslabs/aws-sdk-kotlin/commit/f6502eb10273ea70bc33712d0d350c1c472266ff) which is a workaround mentioned https://github.com/awslabs/aws-sdk-kotlin/issues/842#issuecomment-1426140171
This has broken the S3 client as described in that bug

### Expected behavior

The s3 workaround to keep working.

### Current behavior

can't use S3 client for CI use case.

### Steps to Reproduce

see original bug for details

### Possible Solution

put back forcedPathStyle

### Context

_No response_

0.33.0-beta 

### Platform (JVM/JS/Native)

JVM

Linux

aajtodd:
## Issue \#
fixes #1098 

## Description of changes
https://github.com/awslabs/aws-sdk-kotlin/commit/f6502eb10273ea70bc33712d0d350c1c472266ff erroneously removed the explicit configuration settings `forcePathStyle` and `enableAccelerate` from S3 client config. The intended change was to remove parsing these from CLI specific shared config.

This PR also fixes finalize config to respect the environment/system property settings for `useArnRegion` and `disableMrap` (matching [Java v2](https://github.com/aws/aws-sdk-java-v2/blob/65266c2afdd188930310720e2d3754c614d0d434/services/s3/src/main/java/software/amazon/awssdk/services/s3/S3SystemSetting.java#L27) for the system properties).

### DIFF
--- a/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
@@ -54,9 +54,10 @@ class ClientConfigIntegration : KotlinIntegration {
             """.trimIndent()
         }
 
+        // FIXME: default signer doesn't yet implement sigv4a, default to mrap OFF until it does
         val DisableMrapProp: ConfigProperty = ConfigProperty {
             name = "disableMrap"
-            useSymbolWithNullableBuilder(KotlinTypes.Boolean, "false")
+            useSymbolWithNullableBuilder(KotlinTypes.Boolean, "true")
             documentation = """
                 Flag to disable [S3 multi-region access points](https://docs.aws.amazon.com/AmazonS3/latest/userguide/MultiRegionAccessPoints.html).
             """.trimIndent()
--- a/services/s3/common/test/aws/sdk/kotlin/services/s3/internal/FinalizeConfigTest.kt
+++ b/services/s3/common/test/aws/sdk/kotlin/services/s3/internal/FinalizeConfigTest.kt
@@ -89,6 +89,8 @@ class FinalizeConfigTest {
 
     @Test
     fun testConfigPropertiesPresent() {
+        // regression test to verify that these config properties are generated and can be set
+        // see https://github.com/awslabs/aws-sdk-kotlin/issues/1098
         val builder = S3Client.builder()
         builder.config.disableMrap = true
         builder.config.useArnRegion = true


codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
```diff
@@ -37,10 +54,9 @@ class ClientConfigIntegration : KotlinIntegration {
             """.trimIndent()
         }
 
-        // FIXME: default signer doesn't yet implement sigv4a, default to mrap OFF until it does
```
lauzadis, line 40: Isn't this still true?

services/s3/common/test/aws/sdk/kotlin/services/s3/internal/FinalizeConfigTest.kt
```diff
@@ -0,0 +1,100 @@
...
+        )
+        val sharedConfig = asyncLazy { loadAwsSharedConfig(platform) }
+
+        finalizeS3Config(builder, sharedConfig, platform)
+        assertEquals(true, builder.config.useArnRegion)
+        assertEquals(false, builder.config.disableMrap)
+    }
+
+    @Test
+    fun testConfigPropertiesPresent() {
```
lauzadis, line 91: What is this testing?

lauzadis (approved):

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
aajtodd: We select auth scheme based on the model not on endpoint rules (yet) and we don't support sigv4a yet no matter what. Whether it's enabled or not isn't going to matter I don't think so I just changed it to the "correct" default.

services/s3/common/test/aws/sdk/kotlin/services/s3/internal/FinalizeConfigTest.kt
aajtodd: That these properties don't accidentally disappear again from codegen.

ianbotsf: That's not immediately clear from the code. Can we leave a comment indicating why we have a test with no assertions?

ianbotsf (changes requested):

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
aajtodd: not a hill worth dying on, reverting.

ianbotsf (approved):

PR #1100, opened 2023-11-01
feat: remove deprecated shapes `since` one date `until` another date

lauzadis:
This PR enables configuring an `until` date to remove deprecated shapes that have a `since` field configured.

## Description of changes

This enables us to clean up the API footprint of some AWS services with deprecated members before v1.0.

### DIFF
--- a/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/transforms/RemoveDeprecatedShapes.kt
+++ b/codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/transforms/RemoveDeprecatedShapes.kt
@@ -16,6 +16,7 @@ import java.time.LocalDate
 import java.time.format.DateTimeFormatter
 import java.util.function.Predicate
 
+// A date formatter used to parse Smithy @deprecated trait's `since` field
 private val DEPRECATED_SINCE_DATE_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd")
 
 class RemoveDeprecatedShapes : ConfigurableProjectionTransformer<RemoveDeprecatedShapes.Config>() {


codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/transforms/RemoveDeprecatedShapes.kt
```diff
@@ -0,0 +1,76 @@
...
+        } catch (e: DateTimeException) {
+            throw IllegalArgumentException("Failed to parse configured `until` date ${config.until}", e)
+        }
+
+        return context.transformer.removeShapesIf(context.model, shouldRemoveDeprecatedShape(until))
+    }
+}
+
+internal fun shouldRemoveDeprecatedShape(removeDeprecatedShapesUntil: LocalDate) = Predicate<Shape> { shape ->
+    val deprecatedSince = shape
```
aajtodd, line 47: **style**: 

```kotlin
val since = shape.getTrait<DeprecatedTrait>()?.since?.getOrNull() ?: return@Predicate false
val deprecatedDate = try { ... } catch(...) { .. }

val remove = deprecatedDate < removeDeprecatedShapesUntil
return remove.also {
    if (it) { ... }
}
```

aajtodd (approved):

.changes/c79af2ed-bfad-4196-bf5c-87d647ecbcd1.json
```diff
@@ -0,0 +1,6 @@
+{
+    "id": "c79af2ed-bfad-4196-bf5c-87d647ecbcd1",
+    "type": "feature",
+    "description": "**BREAKING**: Remove deprecated shapes with a `since` date configured",
```
ianbotsf, line 4: **Comment**: This isn't a very helpful message for users. Most of them won't know what shapes have a `@Deprecated` trait with a `since` attribute. I suggest saying something more like `"Remove operations/fields which were marked as deprecated on or before <date>. See <this discussion> for more details."`

Also, the linked discussion post should probably list all of the operations/fields which are being removed. We don't have to suggest new code for _all_ of them but at least one example of moving from something deprecated to whatever replaces it.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/transforms/RemoveDeprecatedShapes.kt
```diff
@@ -0,0 +1,68 @@
...
+    }
+}
+
+internal fun shouldRemoveDeprecatedShape(removeDeprecatedShapesUntil: LocalDate) = Predicate<Shape> { shape ->
+    val since = shape.getTrait<DeprecatedTrait>()?.since?.getOrNull() ?: return@Predicate false
+
+    val deprecatedDate = try { since.toLocalDate() } catch (e: DateTimeException) {
+        println("Failed to parse `since` field $since as a date, skipping removal of deprecated shape $shape")
+        return@Predicate false
+    }
```
ianbotsf, line 51: **Suggestion**: This `try { ...toLocalDate() } catch (e: DateTimeException)` logic is largely duplicated. I suggest moving it into `toLocalDate` and making that function return a nullable value.

Then you can reference it from your initial `transformWithConfig` call:

```kotlin
val until = requireNotNull(config.until.toLocalDate()) { "Failed to parse configured `until` date ${config.until}" }
```

And your `shouldRemoveDeprecatedShape` call:

```kotlin
val deprecatedDate = since.toLocalDate() ?: return false?.also {
    println("Failed to parse `since` field $since as a date, skipping removal of deprecated shape $shape")
}
```

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/transforms/RemoveDeprecatedShapes.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    val shouldRemove = deprecatedDate < removeDeprecatedShapesUntil
+
+    shouldRemove.also {
+        if (it) {
+            println("Removing deprecated shape $shape since its modeled `since` field $deprecatedDate comes before the configured $removeDeprecatedShapesUntil date.")
+        }
+    }
```
ianbotsf, line 59: **Suggestion**: This seems verbose. Eventually, we may be printing hundreds or thousands of members here. I'd suggest removing it unless we have some way of logging at a different level.

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/transforms/RemoveDeprecatedShapes.kt
```diff
@@ -0,0 +1,68 @@
...
+}
+
+/**
+ * Parses a string of yyyy-MM-dd format to [LocalDate].
+ */
+internal fun String.toLocalDate(): LocalDate {
+    val formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd")
```
ianbotsf, line 66: **Nit**: Should be a file-level constant.

codegen/smithy-aws-kotlin-codegen/src/test/kotlin/aws/sdk/kotlin/codegen/transforms/RemoveDeprecatedShapesTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class RemoveDeprecatedShapesTest {
+    private fun shapeDeprecatedSince(since: String?): Shape {
+        val deprecatedTrait = since?.let {
+            DeprecatedTrait.builder()
+                .since(since)
+                .build()
+        } ?: DeprecatedTrait.builder().build()
```
ianbotsf, line 21: **Nit**: There's no reason to special-case null for `since`. The builder and built class treat null as no value anyway. Just `val deprecatedTrait = DeprecatedTrait.builder().since(since).build()` should suffice.

ianbotsf (approved):

codegen/smithy-aws-kotlin-codegen/src/test/kotlin/aws/sdk/kotlin/codegen/transforms/RemoveDeprecatedShapesTest.kt
lauzadis: Oh nice, I made this change while my IDE was still indexing so I couldn't easily check

codegen/smithy-aws-kotlin-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/transforms/RemoveDeprecatedShapes.kt
lauzadis: Sure. I think currently (29 shapes) it's nice to see what's being removed, but I could get rid of this.

PR #1107, opened 2023-11-07
fix: correctly default compositeProjects build setting

ianbotsf:
## Issue \#

(none)

## Description of changes

Our **settings.gradle.kts** build logic for composite projects doesn't correctly handle the situation where a **local.properties** file exists but _does not_ contain a `compositeProjects` key. This change now handles that case by defaulting to **../smithy-kotlin**.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -81,13 +81,13 @@ file("services").listFiles().forEach {
 val compositeProjectList = try {
     val localProperties = java.util.Properties()
     localProperties.load(File(rootProject.projectDir, "local.properties").inputStream())
-    val filePaths = localProperties.getProperty("compositeProjects")
-        ?.splitToSequence(",") // Split comma delimited string into sequence
-        ?.map { it.replaceFirst("^~".toRegex(), System.getProperty("user.home")) } // expand user dir
-        ?.filter { it.isNotBlank() }
-        ?.map { file(it) } // Create file from path
-        ?.toList()
-        ?: emptyList()
+    val propertyVal = localProperties.getProperty("compositeProjects") ?: "../smithy-kotlin"
+    val filePaths = propertyVal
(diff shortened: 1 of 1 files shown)


settings.gradle.kts
```diff
@@ -81,13 +81,13 @@ file("services").listFiles().forEach {
...
     val localProperties = java.util.Properties()
     localProperties.load(File(rootProject.projectDir, "local.properties").inputStream())
-    val filePaths = localProperties.getProperty("compositeProjects")
-        ?.splitToSequence(",") // Split comma delimited string into sequence
-        ?.map { it.replaceFirst("^~".toRegex(), System.getProperty("user.home")) } // expand user dir
-        ?.filter { it.isNotBlank() }
-        ?.map { file(it) } // Create file from path
-        ?.toList()
-        ?: emptyList()
+    val propertyVal = localProperties.getProperty("compositeProjects") ?: "../smithy-kotlin"
```
lauzadis, line 84: Would it be better to throw an exception if there is no `compositeProjects` property (such as with `checkNotNull { ... }`) so it can fallback to the `catch (e: Throwable)` logic on L96? It would remove the (small) duplication

lauzadis (approved):

settings.gradle.kts
aajtodd: `local.properties` is heavily used for Android development so it's entirely reasonable that the file may not have this key.

aajtodd (approved):

settings.gradle.kts
lauzadis: if it's missing, that's fine, it will still default to `../smithy-kotlin`, just in a different way

ianbotsf: Whoops, shipped as soon as I saw approvals but before I saw comments. Yes, it would've probably been better to clean up the duplication.

PR #1110, opened 2023-11-14
misc: bump dependency versions

lauzadis:
Bump dependencies to their latest available versions **except** for `mockk`. 

The latest version 1.13.8 was compiled with a higher version of JVM than we support, and I got this error when trying to upgrade to that version: `Cannot inline bytecode built with JVM target 11 into bytecode that is being built with JVM target 1.8. Please specify proper '-jvm-target' option`

## Description of changes


aajtodd (approved):

lauzadis:
Merged in #1111

PR #1171, opened 2024-01-05
fix: add customization for models missing SIGV4A trait

### LINKED ISSUE #625: S3 MRAP Support
### Describe the feature

Support for S3 [Multi Region Access Points](https://aws.amazon.com/s3/features/multi-region-access-points/)

Depends on: 
* Sigv4A signer
* Endpoints refactor

### Proposed Solution

_No response_

_No response_

### Acknowledge

- [X] I may be able to implement this feature request

0.14.4-beta

### Platform (JVM/JS/Native)

JVM

N/A

#### Community Note
1. Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
2. Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
3. If you are interested in working on this issue, please leave a comment

0marperez:
## Issue \#

closes #625 

## Description of changes

Changed the default auth scheme for S3 and added a test to verify

### DIFF
--- a/aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/interceptors/UnsupportedSigningAlgorithmInterceptor.kt
+++ b/aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/interceptors/UnsupportedSigningAlgorithmInterceptor.kt
@@ -23,7 +23,7 @@ public class UnsupportedSigningAlgorithmInterceptor : HttpInterceptor {
         context.response.exceptionOrNull()?.let {
             if (it is UnsupportedSigningAlgorithmException && it.signingAlgorithm == AwsSigningAlgorithm.SIGV4_ASYMMETRIC) {
                 return Result.failure(
-                    it, // TODO: Add a message and link pointing to AWS SDK for Kotlin developer guide
+                    it, // TODO: Add a message and link pointing to AWS SDK for Kotlin developer guide.
                 )
             }
         }


codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
```diff
@@ -54,14 +56,22 @@ class ClientConfigIntegration : KotlinIntegration {
...
                 Flag to disable [S3 multi-region access points](https://docs.aws.amazon.com/AmazonS3/latest/userguide/MultiRegionAccessPoints.html).
             """.trimIndent()
         }
+
+        val AuthSchemes = RuntimeConfigProperty
+            .AuthSchemes
+            .toBuilder()
+            .apply {
+                symbol = KotlinTypes.Collections.list(RuntimeTypes.Auth.HttpAuth.AuthScheme, default = "listOf(${RuntimeTypes.Auth.HttpAuthAws.SigV4AsymmetricAuthScheme}(${RuntimeTypes.Auth.Signing.AwsSigningStandard.DefaultAwsSigner}))")
+            }
```
ianbotsf, line 73: **Style**: This type of default specification doesn't use imports and instead pastes large, fully-qualified names into the codegen output. Instead, we can use `additionalImports` to make the resulting generated code cleaner:

```kotlin
val AuthSchemes = RuntimeConfigProperty
    .AuthSchemes
    .toBuilder()
    .apply {
        propertyType = ConfigPropertyType.RequiredWithDefault("SigV4AsymmetricAuthScheme(DefaultAwsSigner)")
        additionalImports = listOf(
            RuntimeTypes.Auth.HttpAuthAws.SigV4AsymmetricAuthScheme,
            RuntimeTypes.Auth.Signing.AwsSigningStandard.DefaultAwsSigner,
        )
    }
```

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
```diff
@@ -54,14 +56,22 @@ class ClientConfigIntegration : KotlinIntegration {
...
+            .build()
     }
```
ianbotsf, line 75: **Question**: I'm a little confused about this addition. Won't users still have to replace the auth scheme with one that uses the CRT signer in order for MRAP to work?

0marperez: Hmm, yes. I'm guessing we could avoid users having any additional configuration to use MRAP's by setting the CRT signer as the default for S3. Are there any implications/side effects we would need to worry about when doing that?

0marperez: The reason for setting the default auth scheme to SIGV4A is because the requests were silently falling back to SIGV4 and would succeed sometimes.

0marperez: So users would get the impression that everything is configured properly when that's not the case

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
```diff
@@ -54,14 +56,22 @@ class ClientConfigIntegration : KotlinIntegration {
...
         val DisableMrapProp: ConfigProperty = ConfigProperty {
             name = "disableMrap"
-            useSymbolWithNullableBuilder(KotlinTypes.Boolean, "true")
+            useSymbolWithNullableBuilder(KotlinTypes.Boolean, "false")
             documentation = """
                 Flag to disable [S3 multi-region access points](https://docs.aws.amazon.com/AmazonS3/latest/userguide/MultiRegionAccessPoints.html).
             """.trimIndent()
         }
+
+        val AuthSchemes = RuntimeConfigProperty
```
aajtodd, line 68: **fix**:

This isn't the right approach. We want to register sigv4a support as a configured auth scheme but it doesn't belong in this list, these are user provided. We just want to register support for sigv4a as an entirely new auth scheme handler in `smithy-kotlin` (e.g. here is the [sigv4 auth scheme handler](https://github.com/awslabs/smithy-kotlin/blob/main/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AuthSchemeIntegration.kt#L139)). 

There are some thorny questions to answer about where we should register this since the current [SigV4AuthSchemeIntegration](https://github.com/awslabs/smithy-kotlin/blob/main/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AuthSchemeIntegration.kt#L40) is handling a bit of both right now and there is some overlap (e.g. sigv4a re-uses the same config property `credentialsProvider`). Thats probably ok though since they resolve to the same types. My vote would be to create a new `Sigv4AsymmetricAuthSchemeIntegration` and migrate sigv4a support/logic there. 


We will also need to update the [AuthIndex](https://github.com/awslabs/smithy-kotlin/blob/main/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/model/knowledge/AuthIndex.kt) to also consider endpoint rules since s3 and eventbridge don't model sigv4a usage, they are considered only from endpoint rule properties. Effectively if the sigv4a scheme name shows up in _any_ endpoint property `authSchemes` array node (see [here](https://github.com/awslabs/smithy-kotlin/blob/main/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/auth/SigV4AuthSchemeIntegration.kt#L193-L211) for what I'm talking about) we would consider sigv4a as being registered at the service level for possibly any operation. 


NOTE: `aws.auth#sigv4a` was added as a known auth scheme/shape id in smithy [1.42.0](https://github.com/smithy-lang/smithy/releases/tag/1.42.0)

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
aajtodd: > CRT signer as the default for S3. Are there any implications/side effects we would need to worry about when doing that?
Yes, we can't use CRT by default.

> Won't users still have to replace the auth scheme with one that uses the CRT signer in order for MRAP to work?

Yes, when testing this though what happens is sigv4 is registered in the model and sigv4a comes from the endpoint rule properties. So we get a prioritized list of auth options `[sigv4a, sigv]` after merging those. During selection we see that sigv4a isn't supported/configured and so don't consider it as an option and end up selecting sigv4. This actually works when the access point points to a bucket in the same region that you are configured to sign for. This means users would think sigv4a is working but it's not and in a failover scenario their requests would fail as they wouldn't be valid for a different region. I figured it was better to register support for sigv4a with the default signer and just fail loudly instead with an exception that the signer doesn't support that auth scheme.

aajtodd (dismissed):

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
```diff
@@ -54,10 +54,10 @@ class ClientConfigIntegration : KotlinIntegration {
             """.trimIndent()
         }
 
-        // FIXME: default signer doesn't yet implement sigv4a, default to mrap OFF until it does
+        // FIXME: default signer doesn't yet implement sigv4a
```
lauzadis, line 57: This FIXME can just be removed, it's not descriptive / in the right place anymore

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/interceptors/UnsupportedSigningAlgorithmInterceptor.kt
lauzadis, line 1: It would be good to have some tests for this interceptor

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/interceptors/UnsupportedSigningAlgorithmInterceptor.kt
```diff
@@ -0,0 +1,32 @@
...
+        context.response.exceptionOrNull()?.let {
+            if (it is UnsupportedSigningAlgorithm && it.isSigV4a) {
+                throw UnsupportedSigningAlgorithm(
+                    it.message!!, // TODO: Add a message and link pointing to AWS SDK for Kotlin developer guide
+                    true,
+                )
+            }
+        }
+        return super.modifyBeforeCompletion(context)
+    }
```
ianbotsf, line 31: **Question**: It seems strange to me to _throw_ an exception from this interceptor rather than _return_ one. Would other following interceptors get the chance to modify the `UnsupportedSigningAlgorithm`? Does _returning_ the exception achieve the same result?

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/interceptors/UnsupportedSigningAlgorithmInterceptor.kt
```diff
@@ -0,0 +1,32 @@
...
+@InternalSdkApi
+public class UnsupportedSigningAlgorithmInterceptor : HttpInterceptor {
+    override suspend fun modifyBeforeCompletion(context: ResponseInterceptorContext<Any, Any, HttpRequest?, HttpResponse?>): Result<Any> {
+        context.response.exceptionOrNull()?.let {
+            if (it is UnsupportedSigningAlgorithm && it.isSigV4a) {
+                throw UnsupportedSigningAlgorithm(
+                    it.message!!, // TODO: Add a message and link pointing to AWS SDK for Kotlin developer guide
+                    true,
+                )
+            }
```
ianbotsf, line 28: **Nit**: Use the original `UnsupportedSigningAlgorithm` exception as the `cause` for the newly-thrown one so the stack trace isn't lost.

ianbotsf (approved):

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/interceptors/UnsupportedSigningAlgorithmInterceptor.kt
0marperez: > Does returning the exception achieve the same result?

Yeah, it does👍 
I changed it so it returns the exception like you said

ianbotsf (approved):
**Nit**: We still haven't added an E2E test here which I think we should do as a follow-up . Please create a backlog item.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/UnsupportedSigningAlgorithmIntegration.kt
```diff
@@ -0,0 +1,42 @@
...
+import software.amazon.smithy.model.shapes.ServiceShape
+
+// FIXME: Remove this once sigV4a is supported by default AWS signer
+/**
+ * Registers an interceptor for S3 to deal with the default signer not supporting sigV4a
+ * See: [aws.sdk.kotlin.runtime.http.interceptors.UnsupportedSigningAlgorithmInterceptor]
+ */
+class UnsupportedSigningAlgorithmIntegration : KotlinIntegration {
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean =
+        model.expectShape<ServiceShape>(settings.service).isS3
```
aajtodd, line 25: **fix**: This doesn't match the set of sdkIds where we apply the sigv4a trait customization

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/SigV4AsymmetricTraitCustomization.kt
```diff
@@ -0,0 +1,52 @@
...
+        ModelTransformer.create().mapShapes(model) { shape ->
+            when (shape.isServiceShape) {
+                true ->
+                    (shape as ServiceShape)
+                        .toBuilder()
+                        .addTraits(
+                            mutableSetOf(
+                                SigV4ATrait
+                                    .builder()
+                                    .name(shape.expectTrait<ServiceTrait>().arnNamespace)
```
aajtodd, line 43: **fix**: This should fallback to the sigv4 trait if the arnNamespace isn't present/doesn't match the sigv4 trait.

> This value SHOULD match the arnNamespace property of the [aws.api#service trait](https://smithy.io/2.0/aws/aws-core.html#aws-api-service-trait) **if present** and the name property of the [aws.auth#sigv4 trait](https://smithy.io/2.0/aws/aws-auth.html#aws-auth-sigv4-trait).

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/SigV4AsymmetricTraitCustomization.kt
```diff
@@ -0,0 +1,52 @@
...
+                                    .build(),
+                                AuthTrait(mutableSetOf(SigV4ATrait.ID, SigV4Trait.ID)),
```
aajtodd, line 45: **fix**: This isn't right. (1) if the service has an `auth` trait applied we need to respect it and modify it rather than replace it. (2) the `auth` trait is a prioritized list, this places sigv4a in front of sigv4 which if we configure sigv4a by default then every operation would fail if endpoint rules doesn't return anything for an auth option. The reason it doesn't seem to for S3 right now is they are returning sigv4 and/or sigv4a from their endpoint provider for pretty much every branch. Nothing in endpoint rules dictates this and indeed Eventbridge doesn't appear to return sigv4 anywhere just sigv4a for specific branches (which we appear to have already triggered a customer [issue](https://github.com/awslabs/aws-sdk-kotlin/issues/1184)). 

See [auth trait spec](https://smithy.io/2.0/spec/authentication-traits.html#smithy-api-auth-trait):

PR #1172, opened 2024-01-08
chore: extract bootstrap into separate project with tests

aajtodd:
## Issue \#

n/a

## Description of changes

**refactor**: Add a new `.brazil.json` file with the transform logic for just `aws-sdk-kotlin`. **NOTE**: This will not take effect until we release a new version of `kat` and update the release pipeline to use it. Until then the "global" transform config from repo tools is still the one being used. I tested this locally already on 5 services (all protocols) + the runtime and diffed the transform output using the one from repo tools and this new localized copy
**refactor**: Extract a lot of the logic used to bootstrap the AWS services codegen tasks into a separate composite build. This allows us to actually add tests for our logic and have higher confidence in changes to the build. There is more to do here but I think this is a decent point to stop and review before continuing with any refactoring or adding additional tests.

### DIFF
--- a/build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/Naming.kt
+++ b/build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/Naming.kt
@@ -28,7 +28,7 @@ fun String.kotlinNamespace(): String = split(".")
     .joinToString(separator = ".") { segment -> segment.filter { it.isLetterOrDigit() } }
 
 /**
- * Convert an sdkID to the module/artifact name to use
+ * Convert an sdkId to the module/artifact name to use
  */
 internal fun sdkIdToArtifactName(sdkId: String): String = sdkId.replace(" ", "").replace("-", "").lowercase()
 
--- a/build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/MembershipTest.kt
+++ b/build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/MembershipTest.kt
@@ -19,6 +19,12 @@ class MembershipTest {
     }
 
     @Test
+    fun testExcludePrecedence() {
+        val unit = Membership(setOf("m1"), setOf("m1"))
+        assertFalse(unit.isMember("m1"))
+    }
+
+    @Test
     fun testParse() {
         val expected = Membership(
             setOf("i1", "i2"),
--- a/build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/NamingTest.kt
+++ b/build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/NamingTest.kt
@@ -24,6 +24,9 @@ class NamingTest {
 
     @Test
     fun testSdkIdToArtifactName() {
+        assertEquals("foobar", sdkIdToArtifactName("foo bar"))
+        assertEquals("foobar", sdkIdToArtifactName("foo-bar"))
+        assertEquals("foobar", sdkIdToArtifactName("fOo -Bar"))
     }
 
     @Test
--- a/build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/AwsService.kt
+++ b/build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/AwsService.kt
@@ -85,7 +85,7 @@ fun fileToService(
         .result
         .get()
     val services: List<ServiceShape> = model.shapes(ServiceShape::class.java).sorted().toList()
-    val service = services.singleOrNull() ?: error("Expected one service per aws model, but found ${services.size} in ${file.absolutePath}: ${services.map { it.id }}")
+    val service = services.singleOrNull() ?: error("Expected one service per aws model, but found ${services.size} in ${file.absolutePath}: ${services.joinToString { it.id.toString() }}")
     val protocolName = service.protocolName()
 
     val serviceTrait = service


build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/Naming.kt
```diff
@@ -0,0 +1,39 @@
...
+fun packageNameForService(sdkId: String): String =
+    sdkId.replace(" ", "")
+        .replace("-", "")
+        .lowercase()
+        .kotlinNamespace()
+
+/**
+ * Get the package namespace for a service from it's `sdkId`
+ */
+fun packageNamespaceForService(sdkId: String): String = "$SDK_PACKAGE_NAME_PREFIX${packageNameForService(sdkId)}"
```
lauzadis, line 22: naming: `namespacedPackageNameForService`?

while reading this, I was expecting `packageNamespace` to just return `$SDK_PACKAGE_NAME_PREFIX`, not the package name as well

build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/BootstrapConfig.kt
```diff
@@ -0,0 +1,34 @@
...
+ * the extension or by their artifact/package name.
+ * @param protocols the names of protocols to bootstrap
+ */
+class BootstrapConfig(
+    services: String? = null,
+    protocols: String? = null,
+) {
+    companion object {
+        /**
+         * A bootstrap configuration that includes everything by default
```
lauzadis, line 25: question: this does not include everything by default, it just does not *exclude* anything right? (i.e no services will be bootstrapped with a `BootstrapConfig.ALL`?)

build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/Membership.kt
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Service and protocol membership for SDK generation
+ */
+data class Membership(val inclusions: Set<String> = emptySet(), val exclusions: Set<String> = emptySet())
+
+fun Membership.isMember(vararg memberNames: String): Boolean =
+    memberNames.none(exclusions::contains) && (inclusions.isEmpty() || memberNames.any(inclusions::contains))
+fun parseMembership(rawList: String?): Membership {
+    if (rawList == null) return Membership()
```
lauzadis, line 15: nit: this early return seems not necessary

build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/Naming.kt
```diff
@@ -0,0 +1,39 @@
...
+const val SDK_PACKAGE_NAME_PREFIX: String = "aws.sdk.kotlin.services."
+
+/**
+ * Get the package name to use for a service from it's `sdkId`
+ */
+fun packageNameForService(sdkId: String): String =
+    sdkId.replace(" ", "")
+        .replace("-", "")
+        .lowercase()
+        .kotlinNamespace()
```
lauzadis, line 17: simplification: `fun packageNameForService(sdkId: String): String = sdkIdToArtifactName(sdkId).kotlinNamespace()`

build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/Util.kt
```diff
@@ -0,0 +1,42 @@
...
+ */
+fun ServiceShape.protocolName(): String =
+    listOf(
+        RestJson1Trait.ID,
+        RestXmlTrait.ID,
+        AwsJson1_0Trait.ID,
+        AwsJson1_1Trait.ID,
+        AwsQueryTrait.ID,
+        Ec2QueryTrait.ID,
+    ).first { hasTrait(it) }.name
```
lauzadis, line 42: comment: I feel like Smithy should provide this list of protocol traits

they don't (I checked) but it seems like this defining this custom list will break something in the future

build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/MembershipTest.kt
```diff
@@ -0,0 +1,38 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.sdk.kotlin.gradle.sdk
+
+import kotlin.test.*
+
+class MembershipTest {
+    @Test
+    fun testIsMember() {
```
lauzadis, line 11: comment: would be good to test an element that is both included and excluded is not a member

build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/NamingTest.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    @Test
+    fun testKotlinNamespace() {
+        assertEquals("foobar", "foo b-ar".kotlinNamespace())
+        assertEquals("foobar", " foo   bar  ".kotlinNamespace())
+        assertEquals("foo.bar", "foo.b-ar".kotlinNamespace())
+    }
+
+    @Test
+    fun testSdkIdToArtifactName() {
```
lauzadis, line 26: note: needs tests

build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/BootstrapConfig.kt
aajtodd: No it includes everything. Look at the `isMember` logic, if inclusions is empty then everything is included by default (thats how `./gradlew :codegen:sdk:bootstrap` bootstraps The World (TM))

build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/Membership.kt
aajtodd: Much of this was ported as is

build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/Util.kt
aajtodd: Yeah this is unfortunate but luckily the list of protocols doesn't change much over time.

.brazil.json
```diff
@@ -0,0 +1,32 @@
...
+    "org.jetbrains.kotlin:kotlin-stdlib:1.9.*": "KotlinStdlib-1.9.x",
+    "org.jetbrains.kotlinx:atomicfu-jvm:0.23.1": "AtomicfuJvm-0.23.1",
+    "org.jetbrains.kotlinx:atomicfu:0.23.1": "Atomicfu-0.23.1",
+    "org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.7.*": "KotlinxCoroutinesCoreJvm-1.7.x",
+    "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.*": "KotlinxCoroutinesCore-1.7.x",
+    "org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:1.7.*": "KotlinxCoroutinesJdk8-1.7.x"
+  },
+  "packageHandlingRules": {
+    "versioning": {
+      "defaultVersionLayout": "{MAJOR}.0.x"
```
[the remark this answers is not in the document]
aajtodd: > note: We will have to remember to change this for our next minor version bump.

If we land where I hope we land (which is to have only one internal major version) then I expect this won't ever change even when we bump minor version externally. This is set for now to preserve the existing major version and not accidentally bump it (introducing a new MV internally).

build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/Naming.kt
aajtodd: I didn't want to change anything from what it was (yet anyway) so much of the naming is ported exactly as it was.

ianbotsf (approved):

0marperez (approved):

lauzadis (approved):

PR #1175, opened 2024-01-10
chore: add smoke test traits dep

milesziemer:
## Description of changes

Adds a dependency on smithy-smoke-test-traits in the sdk code gen package so aws service models that use the smokeTests trait won't fail to build because of unknown traits. Generated a client with a smoke test and confirmed there was no diff.

Adding the dependency in the code gen package since it's the models that need the dependency, not the code generator. When smoke tests are implemented in the generator, we can add the dep there.


ianbotsf:
Duplicated to https://github.com/awslabs/aws-sdk-kotlin/pull/1176 to run CI.

ianbotsf (approved):

milesziemer:
Closing because of https://github.com/awslabs/aws-sdk-kotlin/pull/1176

PR #1182, opened 2024-01-19
feat: allow full URI path to a localhost metadata service to be a hostname

### LINKED ISSUE #476: Allow ECS metadata service to use hostnames
### Describe the feature

The ECS credentials provider added in #475 only allows `127.0.0.1`/`[::1]` IP addresses when `AWS_CONTAINER_CREDENTIALS_FULL_URI` is used with `http`. Metadata services should be allowed to provide any hostname they want as long as all of the IP addresses for said host resolve to the loopback device. 

N/A

### Proposed Solution

We need a KMP compatible DNS host resolver, something like:

```kotlin
fun interface HostResolver {
    suspend fun resolve(hostname: String): List<IpAddr>
}

class IpAddr(...) {
    fun isLoopbackDevice(): Boolean
    ...
}
```

This needs to be added to the validate full URI check (or at construction time) and verify that all IP addresses are the loopback device:

```kotlin
class EcsCredentialsProvider(...) {
    suspend fun validateFullUri(uri: String) {
        ...
        
        val ips = hostResolver.resolve(hostname)
        return ips.all { it.isLoopbackDevice() }
    }
}
```

N/A

### Acknowledge

- [ ] I may be able to implement this feature request

N/A

### Platform (JVM/JS/Native)

N/A

N/A

0marperez:
## Issue \#
closes #476

## Description of changes

### DIFF
--- a/aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProviderTest.kt
+++ b/aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProviderTest.kt
@@ -20,6 +20,9 @@ import aws.smithy.kotlin.runtime.http.request.url
 import aws.smithy.kotlin.runtime.http.response.HttpResponse
 import aws.smithy.kotlin.runtime.httptest.TestConnection
 import aws.smithy.kotlin.runtime.httptest.buildTestConnection
+import aws.smithy.kotlin.runtime.net.HostAddress
+import aws.smithy.kotlin.runtime.net.HostResolver
+import aws.smithy.kotlin.runtime.net.IpAddr
 import aws.smithy.kotlin.runtime.net.url.Url
 import aws.smithy.kotlin.runtime.retries.StandardRetryStrategy
 import aws.smithy.kotlin.runtime.time.Instant
@@ -82,6 +85,21 @@ class EcsCredentialsProviderTest {
             }
         }
         return builder.build()
+    }
+
+    /**
+     * Mock resolver that always resolves to loopback address
+     */
+    private object LocalHostResolver : HostResolver {
+        override suspend fun resolve(hostname: String): List<HostAddress> =
+            listOf(
+                HostAddress(
+                    "localhost",
+                    IpAddr.parse("127.0.0.1"),
+                ),
+            )
+        override fun reportFailure(addr: HostAddress) { }
+        override fun purgeCache(addr: HostAddress?) { }
     }
 
     @Test
@@ -152,7 +170,7 @@ class EcsCredentialsProviderTest {
             env = mapOf(AwsSdkSetting.AwsContainerCredentialsFullUri.envVar to uri),
         )
 
-        val provider = EcsCredentialsProvider(testPlatform, engine)
+        val provider = EcsCredentialsProvider(testPlatform, engine, LocalHostResolver)
         val actual = provider.resolve()
         assertEquals(expectedCredentials, actual)
         engine.assertRequests()


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
```diff
@@ -260,3 +274,14 @@ private val eksV4Addr = IpV4Addr(169u, 254u, 170u, 23u)
...
+/**
+ * Wrapper for KMP implementations of IP addresses / DNS host resolving
+ */
+@InternalApi
+public expect class Ip<InternalRepresentation> private constructor(address: InternalRepresentation) {
+    public companion object {
+        public fun getAllFromHost(host: String): List<Ip<Any>>
+    }
+    public fun isLoopbackAddress(): Boolean
+}
```
ianbotsf, line 287: **Correctness**: We already have an [`IpAddr` class](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/IpAddr.kt) and a separate [`HostResolver` interface](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/HostResolver.kt). We should reuse those here instead of re-inventing the wheel.

We _don't_ currently have an `isLoopbackAddress` feature so that's the novel piece to keep. You _can_ delegate down to the platform implementation but it may be simpler just to embed the knowledge in the `IpAddr` class hierarchy and check manually for `127.0.0.1` or `::1`.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
```diff
@@ -164,10 +164,24 @@ public class EcsCredentialsProvider(
...
+                    )
+                }
+
+                if (ipAddresses.all { it.isLoopbackAddress() }) {
+                    return url
+                } else {
+                    throw ProviderConfigurationException(
+                        "The container credentials full URI ($uri) is specified via a hostname whose IP address(es) do not resolve to the loopback device.",
+                    )
+                }
```
ianbotsf, line 183: **Correctness**: The `all` method behaves in a way we don't want for empty collections—it always returns true (because no items in the collection _don't_ match the predicate). I think you may want to invert this check to `if (!ipAddresses.any { !it.isLoopbackAddress() })`.

ianbotsf (changes requested):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
aajtodd: The `IpAddr` class has an [isLoopBack](https://github.com/awslabs/smithy-kotlin/blob/main/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/net/IpAddr.kt#L39) field already. All of the components to implement this are in place we just need to use them

ianbotsf: Oh yes, you're correct. That's what I get for searching for `isLoopbackAddress` instead of something more general. 😅

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
```diff
@@ -164,10 +160,24 @@ public class EcsCredentialsProvider(
                 )
             }
 
-            // TODO - resolve hostnames
-            is Host.Domain -> throw ProviderConfigurationException(
-                "The container credentials full URI ($uri) is specified via hostname which is not currently supported.",
-            )
+            is Host.Domain -> {
+                val hostAddresses = try {
+                    HostResolver.Default.resolve(url.host.toString())
```
ianbotsf, line 165: **Correctness**: For the sake of testability and modularity we should probably make the specific `HostResolver` instance a constructor parameter which _defaults_ to `HostResolver.Default`.

ianbotsf (changes requested):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
```diff
@@ -65,6 +61,7 @@ private const val PROVIDER_NAME = "EcsContainer"
 public class EcsCredentialsProvider(
     public val platformProvider: PlatformProvider = PlatformProvider.System,
     httpClient: HttpClientEngine? = null,
+    private val hostResolver: HostResolver = HostResolver.Default,
 ) : CloseableCredentialsProvider {
```
ianbotsf, line 65: **Correctness**: Sorry, I should've been more clear in my previous comment. This is [a backwards-incompatible change](https://kotlinlang.org/docs/jvm-api-guidelines-backward-compatibility.html#don-t-add-arguments-to-existing-api-functions) and may cause binary compatibility issues. We'll need to maintain the old signature (probably as a secondary constructor) and add a new constructor that accepts the `hostResolver`.

ianbotsf (changes requested):

ianbotsf (approved):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProvider.kt
aajtodd: As an aside, isn't this what the binary validator is for? Should we have guidelines on reviewing the API diff output for binary compatibility? I feel like relying on intuition and experience here is not the right mechanism.

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/EcsCredentialsProviderTest.kt
```diff
@@ -135,7 +135,42 @@ class EcsCredentialsProviderTest {
         val provider = EcsCredentialsProvider(testPlatform, engine)
         assertFailsWith<ProviderConfigurationException> {
             provider.resolve()
-        }.message.shouldContain("The container credentials full URI (http://amazonaws.com/full) is specified via hostname which is not currently supported.")
+        }.message.shouldContain("The container credentials full URI (http://amazonaws.com/full) is specified via a hostname whose IP address(es) do not resolve to the loopback device.")
+    }
+
+    @Test
+    fun testLocalFullUri() = runTest {
```
aajtodd, line 142: **suggestion**: These tests make use of real host resolver when we could plug in a mock resolver that doesn't actually rely on resolving DNS.

aajtodd (approved):

Issue #1184, opened 2024-01-21
Upgrade to 1.0.38 fails


steamstreetjon:
### Describe the bug

When upgrading to 1.0.38, the following exception occurs:

aws.smithy.kotlin.runtime.auth.awssigning.UnsupportedSigningAlgorithmException: SIGV4_ASYMMETRIC support is not yet implemented for the default signer.

1.0.37 works fine. 

### Expected behavior

The exception should not be thrown.

### Current behavior

Calls with the EventBridge client are failing. Others may be as well.

### Steps to Reproduce

Update AWS KT dependency to 1.0.38 or greater. Make an API call.

### AWS Kotlin SDK version used

1.0.38

### Platform (JVM/JS/Native)

JVM

### Operating System and version

Lambda Java 17

lauzadis:
Thanks for the report, I replicated the failure and working on a fix.

lauzadis:
Hi, this is fixed in the latest version of the SDK, v1.0.42. Thanks again for the report.

PR #1186, opened 2024-01-22
fix: application of sigv4a 

lauzadis:
Addresses an issue with applying the SigV4A trait to services which don't model it 
## Issue \#

#1184 
## Description of changes

### DIFF
--- a/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/cloudfrontkeyvaluestore/BackfillSigV4ACustomization.kt
+++ b/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/cloudfrontkeyvaluestore/BackfillSigV4ACustomization.kt
@@ -31,7 +31,7 @@ class BackfillSigV4ACustomization : KotlinIntegration {
                     builder.addTrait(
                         SigV4ATrait.builder()
                             .name(shape.getTrait<SigV4Trait>()?.name ?: shape.getTrait<ServiceTrait>()?.arnNamespace)
-                        .build(),
+                            .build(),
                     )
 
                     val authTraitValueSet = shape.getTrait<AuthTrait>()?.valueSet ?: mutableSetOf()


ianbotsf (approved):

0marperez (approved):

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/SigV4AsymmetricTraitCustomization.kt
```diff
@@ -33,19 +33,25 @@ class SigV4AsymmetricTraitCustomization : KotlinIntegration {
...
-                                    .builder()
-                                    .name(shape.expectTrait<ServiceTrait>().arnNamespace)
-                                    .build(),
-                                AuthTrait(mutableSetOf(SigV4ATrait.ID, SigV4Trait.ID)),
-                            ),
-                        )
-                        .build()
+                true -> {
+                    val builder = (shape as ServiceShape).toBuilder()
+                    builder.addTrait(
```
aajtodd, line 38: **suggestion**: We should make the predication for `enabledForService` to include not only the `sdkId` check but also that it doesn't have the `sigv4a` trait already. This will help the integration be a bit more forward compatible (e.g. the FIXME on L18 becomes less of a concern).

aajtodd (approved):
We should probably try and get some integration tests around sigv4a and sigv4 for these services.

0marperez (commented):
LGTM

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/SigV4AsymmetricTraitCustomization.kt
```diff
@@ -24,28 +24,34 @@ class SigV4AsymmetricTraitCustomization : KotlinIntegration {
     // Needs to happen before the `SigV4AsymmetricAuthSchemeIntegration` & `SigV4AuthSchemeIntegration` (-50 & -50)
     override val order: Byte = -60
 
+    // services which support SigV4A but don't model it
+    private val unmodeledSigV4aServices = listOf("s3", "eventbridge", "cloudfront keyvaluestore")
```
aajtodd, line 28: **correctness**: I don't think cloudfront KVS needs to be in this integration anymore does it?

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/SigV4AsymmetricTraitCustomization.kt
```diff
@@ -24,28 +24,34 @@ class SigV4AsymmetricTraitCustomization : KotlinIntegration {
...
+                            .build(),
+                    )
+
+                    val authTrait = shape.getTrait<AuthTrait>()?.let {
+                        if (it.valueSet.contains(SigV4ATrait.ID)) {
+                            it
+                        } else {
+                            AuthTrait(it.valueSet + mutableSetOf(SigV4ATrait.ID))
+                        }
+                    } ?: AuthTrait(mutableSetOf(SigV4Trait.ID, SigV4ATrait.ID))
```
aajtodd, line 50: **suggestion**: Probably want a remark in here about why this is ordered this way.

(It's because existing services (cloudfront KVS excluded) that use sigv4a do so through endpoint rules and it's not on the model. We don't know which operations it should apply for in those cases and so the safest thing to do is add it at the end and let endpoint rules change the priority as needed. Adding it this way registers support for the trait to let the rest of codegen Just Work (TM)).

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/CloudFrontKeyValueStoreSigV4APrioritizationCustomization.kt
```diff
@@ -0,0 +1,42 @@
...
+import software.amazon.smithy.model.traits.AuthTrait
+import software.amazon.smithy.model.transform.ModelTransformer
+
+/**
+ * Modify the [AuthTrait] of the CloudFront KeyValueStore service, placing SigV4A before SigV4 in the
+ * prioritized list.
+ */
+class CloudFrontKeyValueStoreSigV4APrioritizationCustomization : KotlinIntegration {
+    // Runs after SigV4AsymmetricTraitCustomization (-60) and before `SigV4AsymmetricAuthSchemeIntegration`(-50) and `SigV4AuthSchemeIntegration` (-50)
+    override val order: Byte = -55
```
aajtodd, line 22: I don't think you'll even need to order this if you remove it from the other integration

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/CloudFrontKeyValueStoreSigV4APrioritizationCustomization.kt
```diff
@@ -0,0 +1,42 @@
...
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.ServiceShape
+import software.amazon.smithy.model.traits.AuthTrait
+import software.amazon.smithy.model.transform.ModelTransformer
+
+/**
+ * Modify the [AuthTrait] of the CloudFront KeyValueStore service, placing SigV4A before SigV4 in the
+ * prioritized list.
+ */
+class CloudFrontKeyValueStoreSigV4APrioritizationCustomization : KotlinIntegration {
```
aajtodd, line 20: **fix**: Other service specific customization's are in their own package directory. 


**suggestion/nit**: This is more of a backfill, I'd consider renaming to `BackfillSigv4A` or something to that affect.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/CloudFrontKeyValueStoreSigV4APrioritizationCustomization.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    override fun preprocessModel(model: Model, settings: KotlinSettings): Model =
+        ModelTransformer.create().mapShapes(model) { shape ->
+            when (shape.isServiceShape) {
+                true -> {
+                    val builder = (shape as ServiceShape).toBuilder()
+                    builder.removeTrait(AuthTrait.ID) // remove existing auth trait
+
+                    // add a new auth trait with SigV4A in front
+                    val authTrait = AuthTrait(mutableSetOf(SigV4ATrait.ID, SigV4Trait.ID))
```
aajtodd, line 34: **fix**: We should retain the existing auth trait in case it has other auth schemes over time (unlikely but no need to bake it in this way). I'd also only add sigv4a if the existing service model doesn't have it on the service shape.

aajtodd (changes requested):

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/SigV4AsymmetricTraitCustomization.kt
lauzadis: It is needed, this integration also applies the SigV4A trait. Do you want the new integration to replicate that and then remove cloudfront KVS from this integration?

aajtodd: I'm not sure I'm following, the new CFKVS integration is applying the sigv4a trait as well so it's redundant here no?

lauzadis: It wasn't when you first reviewed it, I just added that application and removed `cloudfront keyvaluestore` from the `SigV4AsymmetricTraitCustomization`

aajtodd (approved):

PR #1193, opened 2024-01-24
misc: e2e tests for s3 mrap

0marperez:
## Description of changes

### DIFF
--- a/services/s3/e2eTest/src/MutliRegionAccessPointTest.kt
+++ b/services/s3/e2eTest/src/MutliRegionAccessPointTest.kt
@@ -43,7 +43,7 @@ class MutliRegionAccessPointTest {
     private lateinit var usEastBucket: String
 
     @BeforeAll
-    private fun setUpMrapTest(): Unit = runBlocking {
+    private fun setUp(): Unit = runBlocking {
         accountId = getAccountId()
         usWestBucket = getTestBucket(s3West, "us-west-2", accountId)
         usEastBucket = getTestBucket(s3East, "us-east-2", accountId)
@@ -65,7 +65,7 @@ class MutliRegionAccessPointTest {
     }
 
     @AfterAll
-    private fun cleanUpMrapTest(): Unit = runBlocking {
+    private fun cleanUp(): Unit = runBlocking {
         if (multiRegionAccessPointWasCreated(s3Control, multiRegionAccessPoint, accountId)) {
             deleteMultiRegionAccessPoint(s3Control, multiRegionAccessPoint, accountId)
         }
--- a/services/s3/e2eTest/src/S3TestUtils.kt
+++ b/services/s3/e2eTest/src/S3TestUtils.kt
@@ -51,15 +51,14 @@ object S3TestUtils {
             .buckets
             ?.mapNotNull { it.name }
 
-        var testBucket = if (region != null) {
-            buckets?.firstOrNull {
-                client.getBucketLocation {
-                    bucket = it
-                    expectedBucketOwner = accountId
-                }.locationConstraint?.value == region && it.startsWith(prefix)
-            }
-        } else {
-            buckets?.firstOrNull { it.startsWith(prefix) }
+        var testBucket = buckets?.firstOrNull { bucketName ->
+            bucketName.startsWith(prefix) &&
+                region?.let {
+                    client.getBucketLocation {
+                        bucket = bucketName
+                        expectedBucketOwner = accountId
+                    }.locationConstraint?.value == region
+                } ?: true
         }
 
         if (testBucket == null) {
@@ -262,10 +261,8 @@ object S3TestUtils {
     ): Boolean {
         println("Checking if multi region access point was created: $multiRegionAccessPointName")
 
-        val search = s3Control.listMultiRegionAccessPoints {
+        return s3Control.listMultiRegionAccessPoints {
             accountId = testAccountId
-        }.accessPoints?.find { it.name == multiRegionAccessPointName }
-
-        return search != null
+        }.accessPoints?.any { it.name == multiRegionAccessPointName } ?: false
     }
 }


services/build.gradle.kts
```diff
@@ -75,6 +75,10 @@ subprojects {
                             implementation(libraries.kotlin.test.junit5)
                             implementation(project(":tests:e2e-test-util"))
                             implementation(libraries.slf4j.simple)
+                            implementation("aws.sdk.kotlin:s3control:+")
```
aajtodd, line 78: **fix**: We can't just add these to every e2e test. For example `sts` has it's own e2e tests and taking a dependency on it will cause classpath conflicts (or more likely unpredictable class loading behavior). We're going to need to probably give this more thought.

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+    val accountId = sts.getCallerIdentity(
+        GetCallerIdentityRequest { },
+    ).account
+
+    sts.close()
+
+    return accountId ?: throw Exception("Unable to get AWS account ID")
+}
+
+internal suspend fun createS3Bucket(
```
aajtodd, line 151: **question**: Why can't we re-use the existing `S3Utils.getBucket()`?

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
         return connection.responseCode
     }
 }
+
+internal suspend fun getAccountId(): String {
```
aajtodd, line 135: **fix**: Move to function on `S3Utils` so these are all grouped.

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+    println("Getting account ID")
+
+    val sts = StsClient {
+        region = "us-west-2"
+    }
+
+    val accountId = sts.getCallerIdentity(
```
aajtodd, line 142: **style**: this operation has a [default parameter](https://sdk.amazonaws.com/kotlin/api/latest/sts/aws.sdk.kotlin.services.sts/-sts-client/get-caller-identity.html), no need to construct an empty request object

```kotlin
val accountId = sts.getCallerIdentity()
```

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+    regionOneBucket: String,
+    regionTwoBucket: String,
+    testAccountId: String,
+) {
+    println("Creating multi region access point: $multiRegionAccessPointName")
+
+    val createRequestToken = s3ControlClient.createMultiRegionAccessPoint(
+        CreateMultiRegionAccessPointRequest {
+            accountId = testAccountId
+            details =
```
aajtodd, line 181: **style**: use the DSL builder

```kotlin
        s3control.createMultiRegionAccessPoint {
            accountId = "accont"
            details { 
                name = "access-point-name"
                regions = listOf(
                    Region { bucket = "one" },
                    Region { bucket = "two" }
                )
            }
        }

```

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+                                bucket = regionTwoBucket
+                            },
+                        )
+                }
+        },
+    )
+
+    waitUntilMultiRegionAccessPointOperationCompletes(
+        s3ControlClient,
+        createRequestToken.requestTokenArn ?: throw Exception("Unable to get request token ARN"),
```
aajtodd, line 199: **style**

```kotlin
s3ControlClient,
checkNotNull(createRequestToken.requestTokenArn) { "request token arn missing" },
...
```

services/s3/e2eTest/src/MutliRegionAccessPointTest.kt
```diff
@@ -0,0 +1,138 @@
...
+import kotlin.test.assertFailsWith
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class MutliRegionAccessPointTest {
+    private val s3West = S3Client { region = "us-west-2" }
+    private val s3East = s3West.withConfig { region = "us-east-2" }
+    private val s3SigV4a = s3West.withConfig { authSchemes = listOf(SigV4AsymmetricAuthScheme(CrtAwsSigner)) }
+    private val s3Control = S3ControlClient { region = "us-west-2" }
+
+    private val usWestBucket = "${S3TestUtils.TEST_BUCKET_PREFIX}for-aws-kotlin-sdk-us-west-2"
```
aajtodd, line 32: **concern**: Why do we need to hard code all of these instead of generating them with some random suffix? 

One of the nice properties of the existing tests and `S3Utils.getObject` is that if something goes wrong it attempts to re-use existing infrastructure and/or clean it up. If these tests fail for some reason it may leave the account in a bad state for future tests. I think we need to try and retain this kind of resiliency.

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+            return
+        }
+
+        TimeUnit.SECONDS.sleep(10L) // Avoid constant status checks
+    }
+
+    throw Exception("The multi-region-access-point $operation operation exceeded the time limit set ($timeLimit ms)")
+}
+
+internal suspend fun deleteS3Bucket(
```
aajtodd, line 309: **question**: Why do we need this if we have `S3Utils.deleteBucketAndAllContents`?

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+    println("Deleting S3 bucket: $bucketName")
+
+    s3Client.deleteBucket(
+        DeleteBucketRequest {
+            bucket = bucketName
+        },
+    )
+}
+
+internal suspend fun objectWasCreated(
```
aajtodd, line 322: **fix**: I don't see this used anywhere, if not remove.

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+    s3: S3Client,
+    bucketName: String,
+    keyName: String,
+): Boolean {
+    println("Checking if object '$keyName' was created in S3 bucket: $bucketName")
+
+    val search = s3.listObjectsV2(
```
aajtodd, line 329: **fix**: This fails to account for pagination but also using `HeadObject` is a better way to do test this. If you need an object to exist use the waiter `waitUntilObjectExists`.

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+        ListObjectsV2Request {
+            bucket = bucketName
+        },
+    ).contents
+
+    return search?.find { it.key == keyName } != null
+}
+
+internal suspend fun s3BucketWasCreated(
```
aajtodd, line 338: **fix**: I don't see this used anywhere, if not remove.

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+    s3: S3Client,
+    bucketName: String,
+): Boolean {
+    println("Checking if S3 bucket was created: $bucketName")
+
+    val search = s3.listBuckets(
```
aajtodd, line 344: **fix**: Same comments, use `HeadBucket` or `waitUntilBucketExists` depending on what you are trying to do.

services/s3/e2eTest/src/MutliRegionAccessPointTest.kt
```diff
@@ -0,0 +1,138 @@
...
+    private val usEastBucket = "${S3TestUtils.TEST_BUCKET_PREFIX}for-aws-kotlin-sdk-us-east-2"
+    private val multiRegionAccessPoint = "aws-sdk-for-kotlin-test-multi-region-access-point"
+    private val keyForObject = "test.txt"
+
+    @BeforeAll
+    private fun setUpMrapTest(): Unit = runBlocking {
+        println("Setting up MutliRegionAccessPointTest tests")
```
lauzadis, line 39: nit: unnecessary print

services/s3/e2eTest/src/MutliRegionAccessPointTest.kt
```diff
@@ -0,0 +1,138 @@
...
+            multiRegionAccessPoint,
+            usWestBucket,
+            usEastBucket,
+            accountId,
+        )
+    }
+
+    @AfterAll
+    private fun cleanUpMrapTest(): Unit = runBlocking {
+        println("Cleaning up MutliRegionAccessPointTest tests")
```
lauzadis, line 66: nit: unnecessary print

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+}
+
+internal suspend fun multiRegionAccessPointWasCreated(
+    s3Control: S3ControlClient,
+    multiRegionAccessPointName: String,
+    testAccountId: String,
+): Boolean {
+    println("Checking if multi region access point was created: $multiRegionAccessPointName")
+
+    val search = s3Control.listMultiRegionAccessPoints(
```
aajtodd, line 358: **style**: use the DSL builders

**question**: Is there not a better way to do this than just listing access points? I'd imagine either `getMultiRegionAccessPoint` throws some kind of predictable error if the AP doesn't exist.

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+    }
+
+    val accountId = sts.getCallerIdentity(
+        GetCallerIdentityRequest { },
+    ).account
+
+    sts.close()
+
+    return accountId ?: throw Exception("Unable to get AWS account ID")
+}
```
lauzadis, line 149: correctness: because this uses `StsClient` it's not an "S3 test util", I'd recommend moving it

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+    s3Client.createBucket(
+        CreateBucketRequest {
+            bucket = bucketName
+            createBucketConfiguration =
+                CreateBucketConfiguration {
+                    locationConstraint = location
+                }
+        },
+    )
+}
```
lauzadis, line 167: question: This seems really similar to `getBucketWithPrefix`. Can that function be adapted to fit your requirements instead of making a new function that does roughly the same thing?

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+
+    val sts = StsClient {
+        region = "us-west-2"
+    }
+
+    val accountId = sts.getCallerIdentity(
+        GetCallerIdentityRequest { },
+    ).account
+
+    sts.close()
```
aajtodd, line 146: **fix**: You can ensure this is closed always by leveraging `use`

```kotlin
StsClient { region = "us-west-2"}.use { sts ->
      ...
}
```

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+
+    return accountId ?: throw Exception("Unable to get AWS account ID")
```
aajtodd, line 148: **style**: `checkNotNull(accountId) { "..."}`

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+}
+
+internal suspend fun getMultiRegionAccessPointArn(
+    s3ControlClient: S3ControlClient,
+    multiRegionAccessPointName: String,
+    testAccountId: String,
+): String {
+    println("Getting multi region access point arn for: $multiRegionAccessPointName")
+
+    s3ControlClient.listMultiRegionAccessPoints(
```
aajtodd, line 213: **correctness**: Should we not be able to leverage `getMultiRegionAccessPoint`?

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+) {
+    println("Deleting object '$keyName' in S3 bucket: $bucketName")
+
+    s3Client.deleteObject(
+        DeleteObjectRequest {
+            bucket = bucketName
+            key = keyName
+        },
+    )
+}
```
lauzadis, line 251: simplification: I don't think these need to be new functions, they are simple enough to write every time

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+                            },
+                        )
+                }
+        },
+    )
+
+    waitUntilMultiRegionAccessPointOperationCompletes(
+        s3ControlClient,
+        createRequestToken.requestTokenArn ?: throw Exception("Unable to get request token ARN"),
+        1000 * 60 * 10, // 10 minutes
```
aajtodd, line 200: **concern**: Why so long?

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+internal suspend fun waitUntilMultiRegionAccessPointOperationCompletes(
+    s3ControlClient: S3ControlClient,
+    request: String,
+    timeLimit: Int,
+    testAccountId: String,
+    operation: String,
+) {
+    val startTime = System.currentTimeMillis()
+
+    while (System.currentTimeMillis() - startTime < timeLimit) {
```
lauzadis, line 288: simplification: `withTimeout(timeLimit) { ... }`

services/s3/e2eTest/src/S3TestUtils.kt
aajtodd: Not going to mark this everywhere but there are several other spots I'd have the same feedback

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+        ).asyncOperation?.requestStatus
+
+        println("Waiting on $operation operation. Status: $status ")
+
+        if (status == "SUCCEEDED") {
+            println("$operation operation succeeded.")
+            return
+        }
+
+        TimeUnit.SECONDS.sleep(10L) // Avoid constant status checks
```
lauzadis, line 303: simplification: `delay(10.seconds)`

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+        1000 * 60 * 5, // 5 minutes
+        testAccountId,
+        "deleteMultiRegionAccessPoint",
+    )
+}
+
+internal suspend fun waitUntilMultiRegionAccessPointOperationCompletes(
+    s3ControlClient: S3ControlClient,
+    request: String,
+    timeLimit: Int,
```
aajtodd, line 282: **style**: Use duration

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+    bucketName: String,
+) {
+    println("Deleting S3 bucket: $bucketName")
+
+    s3Client.deleteBucket(
+        DeleteBucketRequest {
+            bucket = bucketName
+        },
+    )
+}
```
lauzadis, line 320: Same comment here, this function is simple enough, I recommend removing it

services/s3/e2eTest/src/S3TestUtils.kt
aajtodd, line 288: **fix**: You don't need this, just use [withTimeout](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-timeout.html) (or [withTimeoutOrNull](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-timeout-or-null.html) depending)

services/s3/e2eTest/src/S3TestUtils.kt
aajtodd, line 303: **fix**: It doesn't matter here since these are all `runBlocking` tests but this is not coroutine friendly, use [delay](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/delay.html) instead

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+    s3: S3Client,
+    bucketName: String,
+    keyName: String,
+): Boolean {
+    println("Checking if object '$keyName' was created in S3 bucket: $bucketName")
+
+    val search = s3.listObjectsV2(
+        ListObjectsV2Request {
+            bucket = bucketName
+        },
```
lauzadis, line 332: simplification: pass `prefix = keyName` to this call and S3 will do the searching for you

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -116,3 +131,243 @@ object S3TestUtils {
...
+    return search != null
+}
+
+internal fun closeClients(
+    vararg clients: SdkClient,
+) {
+    clients.forEach { client ->
+        client.close()
+    }
+}
```
lauzadis, line 373: Another function that's simple enough to remove

services/s3/e2eTest/src/MutliRegionAccessPointTest.kt
```diff
@@ -0,0 +1,138 @@
...
+                keyForObject,
+            )
+
+            deleteObject(
+                s3SigV4a,
+                multiRegionAccessPointArn,
+                keyForObject,
+            )
+        } catch (exception: Throwable) {
+            println("Test failed with exception: ${exception.cause}")
```
lauzadis, line 109: nit: unnecessary print

services/s3/e2eTest/src/MutliRegionAccessPointTest.kt
```diff
@@ -0,0 +1,138 @@
...
+                multiRegionAccessPointArn,
+                keyForObject,
+            )
+
+            deleteObject(
+                s3SigV4a,
+                multiRegionAccessPointArn,
+                keyForObject,
+            )
+        } catch (exception: Throwable) {
```
aajtodd, line 108: **style**: This is unnecessary, any exception will fail the test with the given cause.

services/s3/e2eTest/src/MutliRegionAccessPointTest.kt
```diff
@@ -0,0 +1,138 @@
...
+
+            assertFailsWith<UnsupportedSigningAlgorithmException> {
+                createObject(
+                    s3West,
+                    multiRegionAccessPointArn,
+                    keyForObject,
+                )
+            }.message.shouldContain("SIGV4_ASYMMETRIC support is not yet implemented for the default signer.")
+        } catch (exception: Throwable) {
+            println("Test failed with exception: ${exception.cause}")
```
lauzadis, line 134: nit: unnecessary print

services/s3/e2eTest/src/MutliRegionAccessPointTest.kt
```diff
@@ -0,0 +1,138 @@
...
+            usEastBucket,
+            accountId,
+        )
+    }
+
+    @AfterAll
+    private fun cleanUpMrapTest(): Unit = runBlocking {
+        println("Cleaning up MutliRegionAccessPointTest tests")
+
+        val accountId = getAccountId()
```
aajtodd, line 68: **suggestion**: You could get this once in `setupMrapTest` and share it across tests.

e.g.

```kotlin
private lateinit var accountId: String

@BeforeAll
private fun setupMrapTest() : Unit = runBlocking { 
    ...
    accountId = getAccountId()
}

```

services/s3/e2eTest/src/MutliRegionAccessPointTest.kt
```diff
@@ -0,0 +1,138 @@
...
+    private val s3SigV4a = s3West.withConfig { authSchemes = listOf(SigV4AsymmetricAuthScheme(CrtAwsSigner)) }
+    private val s3Control = S3ControlClient { region = "us-west-2" }
+
+    private val usWestBucket = "${S3TestUtils.TEST_BUCKET_PREFIX}for-aws-kotlin-sdk-us-west-2"
+    private val usEastBucket = "${S3TestUtils.TEST_BUCKET_PREFIX}for-aws-kotlin-sdk-us-east-2"
+    private val multiRegionAccessPoint = "aws-sdk-for-kotlin-test-multi-region-access-point"
+    private val keyForObject = "test.txt"
+
+    @BeforeAll
+    private fun setUpMrapTest(): Unit = runBlocking {
```
aajtodd, line 38: **nit**: `setup` and `cleanup` are one word.

aajtodd (changes requested):

lauzadis (commented):
The tests look good. I would recommend keeping new functions out of `S3TestUtils` unless you think other S3 tests will use it. For MRAP-specific things, it makes more sense to define them in your specific test file

services/s3/e2eTest/src/MutliRegionAccessPointTest.kt
ianbotsf: **Counter-nit**: "setup" and "cleanup" are nouns. "Set up" and "Clean up" are verbs, which is [the preferred form of method names](https://kotlinlang.org/docs/coding-conventions.html#choose-good-names).

services/build.gradle.kts
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import aws.sdk.kotlin.gradle.dsl.configurePublishing
-import aws.sdk.kotlin.gradle.kmp.*
+import aws.sdk.kotlin.gradle.kmp.kotlin
```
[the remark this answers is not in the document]
0marperez: IntelliJ does that, it doesn't like the `*` imports. I'll leave it how it was

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -115,4 +145,126 @@ object S3TestUtils {
...
+                }
+
+                if (status == "FAILED") throw Exception("$operation operation failed.")
+
+                delay(10.seconds) // Avoid constant status checks
+            }
+        }
+    }
+
+    internal suspend fun multiRegionAccessPointWasCreated(
```
0marperez, line 257: This _could_ be done using a get operation and then looking at the error we get if the MRAP doesn't exist. The only issue is it returns a rather generic `S3ControlException` and all we would have to differentiate a non existent MRAP from a failure would be the error message. Which doesn't seem right as it could change. Either way I can take that approach if that's what everyone thinks

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -115,4 +145,126 @@ object S3TestUtils {
 
         return connection.responseCode
     }
+
+    internal suspend fun getAccountId(): String {
```
0marperez, line 149: I couldn't think of a better place for this function. It seemed out of place in the tests class instead of here with the other utils. Placing it in a STS test utils folder could be another option but it seems a little odd since it's used for S3

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -115,4 +145,126 @@ object S3TestUtils {
...
+            1000 * 60 * 5, // 5 minutes
+            testAccountId,
+            "deleteMultiRegionAccessPoint",
+        )
+    }
+
+    private suspend fun waitUntilMultiRegionAccessPointOperationCompletes(
+        s3ControlClient: S3ControlClient,
+        request: String,
+        duration: Int,
```
aajtodd, line 232: **nit**: `withTimeout` also accepts a `Duration`, instead of passing this as an integer you could just be writing `5.minutes` in the calling code and accept a `Duration`. Also probably rename to something like `timeoutAfter`

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -115,4 +145,126 @@ object S3TestUtils {
...
+                }.asyncOperation?.requestStatus
+
+                println("Waiting on $operation operation. Status: $status ")
+
+                if (status == "SUCCEEDED") {
+                    println("$operation operation succeeded.")
+                    return@withTimeout
+                }
+
+                if (status == "FAILED") throw Exception("$operation operation failed.")
```
aajtodd, line 250: **style**: 

```
check(status != "FAILED") { "$operation operation failed")
```

services/build.gradle.kts
```diff
@@ -88,6 +88,26 @@ subprojects {
                         description = "Run e2e service tests"
                         group = "verification"
 
+                        if (project.name == "s3") {
```
aajtodd, line 91: I suppose this works but this is getting ugly. I'm thinking we'll be revisiting the structure around e2e tests soon.

aajtodd (approved):

services/build.gradle.kts
0marperez: I agree,I think we could refactor some of this into a functions if it starts getting used more.

services/s3/e2eTest/src/MutliRegionAccessPointTest.kt
lauzadis: nit: My thought on this is to just call it `setup`, the test class is enough of a hint of what you are setting up

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -5,31 +5,61 @@
...
+        var testBucket = if (region != null) {
+            buckets?.firstOrNull {
+                client.getBucketLocation {
+                    bucket = it
+                    expectedBucketOwner = accountId
+                }.locationConstraint?.value == region && it.startsWith(prefix)
+            }
+        } else {
+            buckets?.firstOrNull { it.startsWith(prefix) }
+        }
```
lauzadis, line 62: nit: possible simplification
```kt
var testBucket = buckets?.firstOrNull { bucketName -> 
    bucketName.startsWith(prefix) && 
    region?.let { 
        client.getBucketLocation { 
            bucket = bucketName
            expectedBucketOwner = accountId
        }.locationConstraint?.value == region 
    } ?: true 
}
```

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -115,4 +145,126 @@ object S3TestUtils {
 
         return connection.responseCode
     }
+
+    internal suspend fun getAccountId(): String {
+        println("Getting account ID")
```
lauzadis, line 150: nit: seems like an unnecessary print to me

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -115,4 +145,126 @@ object S3TestUtils {
...
+        multiRegionAccessPointName: String,
+        testAccountId: String,
+    ): Boolean {
+        println("Checking if multi region access point was created: $multiRegionAccessPointName")
+
+        val search = s3Control.listMultiRegionAccessPoints {
+            accountId = testAccountId
+        }.accessPoints?.find { it.name == multiRegionAccessPointName }
+
+        return search != null
```
lauzadis, line 268: nit: simplification

```kt
return s3Control.listMultiRegionAccessPoints {
    accountId = testAccountId
}.accessPoints?.any { it.name == multiRegionAccessPointName }
```

services/s3/e2eTest/src/S3TestUtils.kt
lauzadis: Yeah this seems fine, especially now that S3 E2E tests are guaranteed to have a dependency on STS

lauzadis (approved):

services/s3/e2eTest/src/S3TestUtils.kt
0marperez: There's a couple of other print statements in the utils, why does this one seem unnecessary?

Issue #1189, opened 2024-01-22
Can't sequentially bootstrap more than one SDK client


lauzadis:
### Describe the bug

When trying to bootstrap a second SDK client after one has already been bootstrapped, the task succeeds but there is no client generated.

### Expected behavior

The second client should be bootstrapped. 

### Current behavior

The bootstrap task succeeds but does not actually generate a client.

### Steps to Reproduce

```sh
./gradlew --no-daemon -Paws.kotlin.native=false -Paws.services=lambda :codegen:sdk:bootstrap
# observe lambda client is generated
./gradlew --no-daemon -Paws.kotlin.native=false -Paws.services=ec2 :codegen:sdk:bootstrap
# observe task succeeds, but there is no EC2 client
```

### Possible Solution

The bug is related to Gradle caching the `generateSmithyBuild` task results. I noticed this in the logs: `Task :codegen:sdk:generateSmithyBuild UP-TO-DATE`.

Adding a `--rerun-tasks` flag to the bootstrap command fixes the issue, but this problem should be fixed in the Gradle task itself.

### AWS Kotlin SDK version used

1.0.40-SNAPSHOT

### Platform (JVM/JS/Native)

JVM

### Operating System and version

MacOS

aajtodd:
Yeah I noticed this recently as well I think, this may be fixed in a WIP [branch](https://github.com/awslabs/aws-sdk-kotlin/compare/publishing-qa...publishing-qa-scaffold#diff-85c60e3c4e4de11fd957913876d5cef4f6ee24fc3fa8b3187c2c35cd79ade77fR151).

PR #1202, opened 2024-02-05
chore: add scaffolding task

aajtodd:
## Issue \#

fixes https://github.com/awslabs/aws-sdk-kotlin/issues/1189

## Description of changes

* Adds a new `scaffold` task for service client codegen to generate package metadata once which will be re-used across future build/publish. This allows us to change the logic behind inference over time without breaking existing packages while also serving as a hedge against accidentally breaking existing packages/artifacts.

### DIFF
--- a/build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/tasks/UpdatePackageManifestTaskTest.kt
+++ b/build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/tasks/UpdatePackageManifestTaskTest.kt
@@ -9,7 +9,6 @@ import aws.sdk.kotlin.gradle.sdk.PackageMetadata
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
-import kotlinx.serialization.json.decodeFromStream
 import org.gradle.kotlin.dsl.create
 import org.gradle.testfixtures.ProjectBuilder
 import org.junit.jupiter.api.io.TempDir
@@ -61,7 +60,7 @@ class UpdatePackageManifestTaskTest {
         val task = setupTest(tempDir, "Test Gradle")
         task.updatePackageManifest()
 
-        val updated = json.decodeFromStream<PackageManifest>(tempDir.resolve("packages.json").inputStream())
+        val updated = PackageManifest.fromFile(tempDir.resolve("packages.json"))
         val expectedPackages = initialManifest.packages.toMutableList()
         expectedPackages.add(
             PackageMetadata("Test Gradle", "aws.sdk.kotlin.services.testgradle", "testgradle", "AwsSdkKotlinTestGradle"),
@@ -75,7 +74,7 @@ class UpdatePackageManifestTaskTest {
     fun testManifestNotExistYet(@TempDir tempDir: File) {
         val task = setupTest(tempDir, "Test Gradle", null)
         task.updatePackageManifest()
-        val updated = json.decodeFromStream<PackageManifest>(tempDir.resolve("packages.json").inputStream())
+        val updated = PackageManifest.fromFile(tempDir.resolve("packages.json"))
         val expected = PackageManifest(
             listOf(
                 PackageMetadata("Test Gradle", "aws.sdk.kotlin.services.testgradle", "testgradle", "AwsSdkKotlinTestGradle"),
@@ -121,7 +120,7 @@ class UpdatePackageManifestTaskTest {
         }
 
         task.updatePackageManifest()
-        val updated = json.decodeFromStream<PackageManifest>(tempDir.resolve("packages.json").inputStream())
+        val updated = PackageManifest.fromFile(tempDir.resolve("packages.json"))
         assertEquals(initialManifest, updated)
     }
 
@@ -150,7 +149,7 @@ class UpdatePackageManifestTaskTest {
         }
 
         task.updatePackageManifest()
-        val updated = json.decodeFromStream<PackageManifest>(tempDir.resolve("packages.json").inputStream())
+        val updated = PackageManifest.fromFile(tempDir.resolve("packages.json"))
         val expected = initialManifest.copy(
             initialManifest.packages + listOf(
                 PackageMetadata("Package 3", "aws.sdk.kotlin.services.package3", "package3", "AwsSdkKotlinPackage3"),


build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/tasks/Scaffold.kt
```diff
@@ -0,0 +1,109 @@
...
+        description = "Flag to discover and process only new packages not currently in the manifest. Only applicable when used in conjunction with `model-dir`",
+    )
+    @get:Input
+    public abstract val discover: Property<Boolean>
+
+    @OptIn(ExperimentalSerializationApi::class)
+    @TaskAction
+    fun updatePackageManifest() {
+        check(modelFile.isPresent || modelDir.isPresent) { "one of `model` or `model-dir` is required" }
+        check(!(modelFile.isPresent && modelDir.isPresent)) { "only one of `model` or `model-dir` can be set" }
```
ianbotsf, line 53: **Style**: Could combine:

```kotlin
check(modelFile.isPresent != modelDir.isPresent) { "Exactly one of `model` or `model-dir` must be set" }
```

build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/AwsServiceTest.kt
```diff
@@ -8,22 +8,20 @@ import org.gradle.kotlin.dsl.extra
...
-import kotlin.test.assertEquals
-import kotlin.test.assertNull
+import kotlin.test.*
 
 class AwsServiceTest {
 
     val modelContents = """
-        ${"$"}version: "2.0"
+        ${"$"}version: "2"
         namespace gradle.test
```
ianbotsf, line 17: **Question**: Why did this change?

build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/AwsServiceTest.kt
```diff
@@ -69,12 +81,13 @@ class AwsServiceTest {
             val result = testWith(tempDir, bootstrap)
             val expected = AwsService(
                 "gradle.test#TestService",
-                "aws.sdk.kotlin.services.testgradle",
+                "aws.sdk.kotlin.services.testgradle2",
                 "1.2.3",
```
ianbotsf, line 85: **Question**: Is this rename significant?

build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/PackageManifest.kt
```diff
@@ -0,0 +1,68 @@
...
+    check(errors.isEmpty()) { errors.joinToString(separator = "\n") }
+}
+
+@Serializable
+data class PackageMetadata(
+    public val sdkId: String,
+    public val namespace: String,
+    public val artifactName: String,
+    public val brazilName: String,
+) {
```
ianbotsf, line 54: **Suggestion**: Add KDocs. In particular, it's not clear that `artifactName` means "Maven artifact name (i.e., the 'A' in 'GAV')" solely by looking at the code.

ianbotsf (approved):
I think the term "scaffold" is not clear enough on its own. May I suggest "scaffold-new-packages", "update-package-manifest", etc., as terms which are more clear for a task name or a task class name.

build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/AwsServiceTest.kt
aajtodd: It's proving that the namespace is taken from the package manifest given vs the inferred namespace from `sdkId`.

build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/AwsServiceTest.kt
aajtodd: I've slept and hacked since then and I don't recall. I want to say something about the model discovery and validation triggering something here and so I just fixed it to make smithy happy again.

aajtodd:
> I think the term "scaffold" is not clear enough on its own. May I suggest "scaffold-new-packages", "update-package-manifest", etc., as terms which are more clear for a task name or a task class name.

As the task name? It is qualified by the project `:codegen:sdk:scaffold` when invoked to be fair which does give it some additional context.

Task names are (err should be) camel cased though so it would be one of:

* `scaffold`
* `scaffoldNewPackages`
* `updatePackageManifest`
* etc

I chose `scaffold` as it sets up any scaffolding for new services. Currently that is only updating the package manifest, in the future it could be more though. Not a hill I want to die on though so if you feel strongly (or anyone else does) just weigh in with your preference.

build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/AwsServiceTest.kt
lauzadis: services still model this as `2.0`, so if it breaks in this test, would it also break once we try to use it for a real service?

probably not, because the CI which bootstraps services passes, still weird.

codegen/sdk/build.gradle.kts
```diff
@@ -185,6 +192,11 @@ tasks.register("bootstrap") {
     finalizedBy(stageSdks)
 }
 
+tasks.register<Scaffold>("scaffold") {
+    group = "codegen"
+    description = "Add a new service to packages.json manifest"
```
lauzadis, line 197: nit: it can add or update multiple services

lauzadis:
I'd vote for `updatePackageManifest` for ultimate clarity

lauzadis (approved):

build-support/src/test/kotlin/aws/sdk/kotlin/gradle/sdk/AwsServiceTest.kt
ianbotsf: In [Smithy IDL](https://smithy.io/2.0/spec/idl.html#version-statement), both of these statements are valid but seem to mean the exact same thing—tooling must support an IDL version of anywhere from 2.0 (inclusive) to 3.0 (exclusive). I'd be curious what broke and if there's maybe a bug somewhere else.

ianbotsf (approved):

lauzadis (approved):

PR #1204, opened 2024-02-06
chore: add @InternalSdkApi to binary compatibility checks; introduce new PR check to verify no API-breaking changes

### LINKED ISSUE #1191: Strengthen backwards-compatibility CI checks
Now that we've gone 1.0/GA and we've had some time with the binary compatibility validator, we should strengthen the checking we do to monitor backwards compatibility. Specifically, we should:
* Re-enable BCV scans on `@InternalApi` and `@InternalSdkApi` symbols
* Add a CI check which requires only forward-compatible API diffs, bypassable with a new PR tag
* Review [our version policy](https://github.com/awslabs/aws-sdk-kotlin/blob/main/VERSIONING.md) for correctness and detail about how/when backwards-incompatible changes are made and how that affects versioning and upgrades

ianbotsf:
## Issue \#

Resolves #1191 

## Description of changes

Same as https://github.com/awslabs/smithy-kotlin/pull/1030 but for **aws-sdk-kotlin**:
* Removing `@InternalSdkApi` from the exemptions for the backwards compatibility verifier
  * Also updates the API dumps because a lot of previously-exempted APIs are now in scope
* Adds a PR check which verifies that either no API dump lines have been deleted _or_ the **acknowledge-api-break** label is applied to the PR

### DIFF
--- a/.github/workflows/api-compat-verification.yml
+++ b/.github/workflows/api-compat-verification.yml
@@ -1,0 +1,32 @@
+name: API compatibility verification
+
+on:
+  pull_request:
+    types: [ opened, synchronize, reopened, labeled, unlabeled ]
+    branches: [ main ]
+
+jobs:
+  api-compat-verification:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check for API compatibility
+        if: ${{ !contains(github.event.pull_request.labels.*.name, 'acknowledge-api-break') }}
+        run: |
+          git fetch origin ${{ github.base_ref }} --depth 1 && \
+          git diff remotes/origin/${{ github.base_ref }} --numstat "*.api" | awk '
+            BEGIN { s = 0 }
+            
+            # git diff numstat shows lines deleted in field 2, hence sum up field 2 across all items 
+            { s += $2 }
+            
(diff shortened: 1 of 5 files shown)


lauzadis (approved):

0marperez (approved):

.github/workflows/api-compat-verification.yml
```diff
@@ -0,0 +1,32 @@
+name: API compatibility verification
+
+on:
+  pull_request:
+    types: [ opened, synchronize, reopened, labeled, unlabeled ]
+    branches: [ main ]
+
+jobs:
+  api-compat-verification:
```
aajtodd, line 9: **suggestion**: We could move this to workflow to `aws-kotlin-repo-tools` and share it across repos. It's not a lot of code and not likely to change often but something to consider if we do end up having to make changes to each repo over time.

aajtodd (approved):

.github/workflows/api-compat-verification.yml
ianbotsf: Yes but I'd rather do that in a separate effort. I think a few of our GH workflows are candidates for consolidation which would warrant a new PR.

PR #1207, opened 2024-02-08
feat: codebuild ci workflow for external contributors

0marperez:
## Issue \#
N/A

## Description of changes
Added inputs to workflow. Inputs will be passed along to the codebuild job via env var

### DIFF
--- a/.github/workflows/codebuild-ci.yml
+++ b/.github/workflows/codebuild-ci.yml
@@ -37,14 +37,17 @@ jobs:
   e2e-tests:
     runs-on: ubuntu-latest
     steps:
+      - name: Verify PRs are not running malicious code
+        if: ${{ (inputs.aws-sdk-kotlin-pr != '' || inputs.smithy-kotlin-pr != '') && inputs.check-pr == false }}
+        run: |
+          echo Please verify the PRs are not running any malicious code and mark the checkbox true when running the workflow
+          exit 1
       - name: Configure AWS Credentials
-        if: ${{ ((inputs.aws-sdk-kotlin-pr != '' || inputs.smithy-kotlin-pr != '') && inputs.check-pr == true) || (inputs.aws-sdk-kotlin-pr == '' && inputs.smithy-kotlin-pr == '') }}
         uses: aws-actions/configure-aws-credentials@v4
         with:
           role-to-assume: ${{ secrets.CI_AWS_ROLE_ARN }}
           aws-region: us-west-2
       - name: Run E2E Tests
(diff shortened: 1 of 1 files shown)


.github/workflows/codebuild-ci.yml
```diff
@@ -29,6 +47,7 @@ jobs:
         uses: aws-actions/aws-codebuild-run-build@v1
         with:
           project-name: gh-aws-sdk-kotlin-e2e-tests
+          env-vars-for-codebuild: SDK_PR, SMITHY_PR
```
aajtodd, line 50: **question**: Why doesn't the service check batch job require similar changes? Why does it work (or does it)?

.github/workflows/codebuild-ci.yml
```diff
@@ -5,6 +5,24 @@ on:
...
+      aws-sdk-kotlin-pr:
+        description: aws-sdk-kotlin PR number
+        type: number
+        required: true
+      smithy-kotlin-pr:
+        description: smithy-kotlin PR number (optional)
+        type: number
+        required: false
+      check-pr:
+        description: I verified that the PR's are not running any malicious code
```
ianbotsf, line 18: **Nit**: PR's (possessive) → PRs (plural)

.github/workflows/codebuild-ci.yml
```diff
@@ -5,6 +5,24 @@ on:
...
+        required: true
+        type: boolean
+        default: false
```
ianbotsf, line 21: **Question**: What does this setting actually do? I don't see it used anywhere else.

0marperez: It's meant as a reminder, it can be enforced though. The workflow could not run if the checkbox isn't marked

ianbotsf: I think enforcement is a good idea. A checkbox that has no effect and can be left unchecked with no repercussions seems pointless.

.github/workflows/codebuild-ci.yml
```diff
@@ -38,21 +59,31 @@ jobs:
             echo "cancelling in-progress build: id=$BUILD_ID"
             aws codebuild stop-build --id $BUILD_ID
           fi
+      - name: Verify PRs are not running malicious code
```
aajtodd, line 62: **suggestion**: Move this as the first step and you won't need the other `if` checks on each step. A failure of an earlier step will prevent it from running.

.github/workflows/codebuild-ci.yml
```diff
@@ -62,3 +93,8 @@ jobs:
             echo "cancelling in-progress batch build: id=$BUILD_ID"
             aws codebuild stop-build --id $BUILD_ID
           fi
+      - name: Verify PRs are not running malicious code
```
aajtodd, line 96: Same suggestion, move to earlier and you won't need a conditional on every step

ianbotsf (approved):

aajtodd (approved):

PR #1215, opened 2024-02-12
fix: profile credentials provider needing opt in to internal annotation

### LINKED ISSUE #1208: ProfileCredentialsProvider isn't available without internal modules
### Describe the feature

One cannot provide explicitly a provider of class `ProfileCredentialsProvider`, since some of its constructor parameters are explicitly marked with `aws.sdk.kotlin.runtime.InternalSdkApi`. So, compared to Java v2 SDK, one couldn't reliably use it:
1. It raises warnings on unstable non-published API (correctly),
2. It requires bringing in `aws.smithy.kotlin:http-client` dependency (that might pull something else), where `aws.sdk.kotlin:aws-config` would probably suffice.

Please consider custom configuration of a client that would like to share a bit of code like this:
```kt
import aws.sdk.kotlin.runtime.auth.credentials.ProfileCredentialsProvider
import aws.sdk.kotlin.runtime.client.AwsSdkClientConfig
import aws.smithy.kotlin.runtime.auth.awscredentials.CachedCredentialsProvider
import aws.smithy.kotlin.runtime.auth.awscredentials.CredentialsProviderConfig

fun <T> T.localSetup()
    where
        T: AwsSdkClientConfig.Builder,
        T: CredentialsProviderConfig.Builder
    = apply {
        region = "eu-west-1"
        credentialsProvider = CachedCredentialsProvider(
            ProfileCredentialsProvider("my-awesome-profile")
        )
    }
```
While it looks innocent it actually compiles against internal API signatures.

### Proposed Solution

Don't know any ideas, sorry. It's clearly implementable, but probably that artifact separation was intended. A lo of Jars out there! Without purely collapsing the whole hierarchy with a dependency or building a stable HttpClient / runtime API I can't think of a solution myself.

_No response_

### Acknowledge

- [ ] I may be able to implement this feature request

1.0.52

### Platform (JVM/JS/Native)

JVM OpenJDK 17

Ubuntu 22.04 amd64

0marperez:
## Issue \#
closes #1208 

## Description of changes

One of the constructor parameters was marked as internal SDK API. Which was causing the profile credentials provider to be marked as internal API. 

### DIFF
--- a/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
+++ b/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
@@ -5,6 +5,7 @@
 
 package aws.sdk.kotlin.runtime.auth.credentials
 
+import aws.sdk.kotlin.runtime.InternalSdkApi
 import aws.sdk.kotlin.runtime.auth.credentials.profile.LeafProvider
 import aws.sdk.kotlin.runtime.auth.credentials.profile.ProfileChain
 import aws.sdk.kotlin.runtime.auth.credentials.profile.RoleArn
@@ -77,13 +78,27 @@ import kotlin.coroutines.coroutineContext
  * @param configurationSource An optional configuration source to use for loading shared config. If not provided,
  * it will be resolved from the environment.
  */
-public class ProfileCredentialsProvider(
+public class ProfileCredentialsProvider @InternalSdkApi constructor(
     public val profileName: String? = null,
     public val region: String? = null,
     public val platformProvider: PlatformProvider = PlatformProvider.System,
     public val httpClient: HttpClientEngine? = null,
     public val configurationSource: AwsConfigurationSource? = null,
 ) : CloseableCredentialsProvider {
+
+    public constructor(
+        profileName: String? = null,
+        region: String? = null,
+        platformProvider: PlatformProvider = PlatformProvider.System,
+        httpClient: HttpClientEngine? = null,
+    ) : this (
+        profileName,
+        region,
+        platformProvider,
+        httpClient,
+        null,
+    )
+
     private val namedProviders = mapOf(
         "Environment" to EnvironmentCredentialsProvider(platformProvider::getenv),
         "Ec2InstanceMetadata" to ImdsCredentialsProvider(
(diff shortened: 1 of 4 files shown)


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsConfigLoader.kt
```diff
@@ -76,7 +76,6 @@ private fun mergeSectionMaps(maps: List<SectionMap>): SectionMap = buildMap {
 /**
  * Specifies the active profile and configured (may not actually exist) locations of configuration files.
  */
-@InternalSdkApi
 public data class AwsConfigurationSource(val profile: String, val configPath: String, val credentialsPath: String)
```
ianbotsf, line 79: **Correctness**: I believe we still want this type to be `@InternalSdkApi`. I think instead we should add to `ProfileCredentialsProvider` a new constructor overload that doesn't have an `AwsConfigurationSource` parameter. Then we can mark the current constructor as `@InternalSdkApi`.

ianbotsf (dismissed):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsConfigLoader.kt
0marperez: What about if the user wants to configure the `AwsConfigurationSource`?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
```diff
@@ -77,13 +78,27 @@ import kotlin.coroutines.coroutineContext
...
-public class ProfileCredentialsProvider(
+public class ProfileCredentialsProvider @InternalSdkApi constructor(
     public val profileName: String? = null,
     public val region: String? = null,
     public val platformProvider: PlatformProvider = PlatformProvider.System,
     public val httpClient: HttpClientEngine? = null,
     public val configurationSource: AwsConfigurationSource? = null,
 ) : CloseableCredentialsProvider {
+
+    public constructor(
```
aajtodd, line 89: **question**: Have you tried this out by publishing it to maven local and trying to use it in a project? I tested the issue originally to verify and could see intellij correctly flagged it, we should verify this actually fixes the issue for customers by comparing before and after if you haven't already.

aajtodd (commented):
Looks good just one question on whether we verified the behavior change.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
0marperez: Yes, I published it to maven local and gave it a try. I was able to re-create the issue first as well.

0marperez: I could see the warning go away and there was no longer a need to import `aws.smithy.kotlin:http-client`

aajtodd (approved):

lauzadis (approved):

Issue #1217, opened 2024-02-15
CodeArtifact .publishPackageVersion fails for large generic artifact with 'InvalidSignatureException'


cloudshiftchris:
### Describe the bug

Uploading a large (937MB) file to AWS CodeArtifact as a generic artifact fails with this exception:
```
x-amzn-errortype: InvalidSignatureException:http://internal.amazon.com/coral/com.amazon.coral.service/
x-amzn-requestid: 9610da2f-c4ca-4b81-b59d-cd2964dc9fe7
```
The SHA256 has been validated as correctly calculated.

Uploading smaller files works (have not tested for size boundaries).
Forcing a SHA256 mismatch results in a different, specific exception.

Furthermore, perhaps related (or not), the same attempt using AWS CLI fails for different reasons (SSL/EOF errors - seems like CodeArtifact terminates the connection).

937MB is within the [5GB limit](https://docs.aws.amazon.com/codeartifact/latest/ug/service-limits.html) for an individual CodeArtifact asset.

AWS Kotlin SDK 1.0.56

### Expected behavior

When invoked with valid data AWS SDK correctly uploads large files into CodeArtifact.

### Current behavior

Fails with this exception: 
```
09:24:10.461 [DefaultDispatcher-worker-3] DEBUG httpTraceMiddleware - HttpResponse:
HTTP 403: Forbidden
content-length: 192
content-type: application/json
date: Thu, 15 Feb 2024 17:20:33 GMT
x-amzn-errortype: InvalidSignatureException:http://internal.amazon.com/coral/com.amazon.coral.service/
x-amzn-requestid: 9610da2f-c4ca-4b81-b59d-cd2964dc9fe7


09:24:10.463 [DefaultDispatcher-worker-3] TRACE aws.smithy.kotlin.runtime.http.engine.okhttp.OkHttpEngine - response body available
09:24:10.463 [DefaultDispatcher-worker-3] TRACE aws.smithy.kotlin.runtime.http.engine.okhttp.OkHttpEngine - response body finished: bytesConsumed=192
[...]

PR #1224, opened 2024-02-19
fix: only enable `aws-chunked` content encoding for S3

lauzadis:
This PR disables `aws-chunked` content encoding for all services by default, except S3. It also adds support for a new `EnableAwsChunked` signing attribute which controls enabling `aws-chunked` content encoding. 

## Issue \#

closes https://github.com/awslabs/aws-sdk-kotlin/issues/1217

## Description of changes

This change is required because S3 is the only service which explicitly supports `aws-chunked` content encoding.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -9,8 +9,8 @@ coroutines-version = "1.7.3"
 atomicfu-version = "0.23.1"
 
 # smithy-kotlin codegen and runtime are versioned separately
-smithy-kotlin-runtime-version = "1.0.14"
-smithy-kotlin-codegen-version = "0.30.15"
+smithy-kotlin-runtime-version = "1.0.15"
+smithy-kotlin-codegen-version = "0.30.16"
 
 # codegen
 smithy-version = "1.42.0"


codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3SigningConfig.kt
```diff
@@ -20,7 +21,7 @@ import software.amazon.smithy.model.shapes.ServiceShape
  * Set default signing attributes in the execution context (which ultimately becomes the signing context) for S3.
  */
 class S3SigningConfig : KotlinIntegration {
-    // auth schemes are de-duped by taking the last one registered
+    // integrations are de-duped by taking the last one registered
```
lauzadis, line 24: question: is this comment now correct? I wasn't sure why auth schemes were relevant here

ianbotsf: No, I don't think this comment is correct for integrations. All of the registered integrations are "taken" and there is no possibility of duplication.

The comment was correct for auth schemes and I believe it was referring to the `order` value defined below, indicating that whatever auth scheme this integration registered should be the one that's taken because it'll be the last one (unless a user overrides it even later in client config). That said, I don't see this integration registering an auth scheme so I'm not sure how it's relevant.

ianbotsf (approved):

0marperez (approved):

PR #1225, opened 2024-02-19
fix: codebuild ci workflow options

0marperez:
## Issue \#
N/A

## Description of changes
The script called from the workflow had a bug. If the workflow was not called from a PR then the SDK PR number would be used as a source. 

The fix was to label the options, and add logic for empty options. This fixes the bug and should avoid confusion.

### DIFF
--- a/.github/scripts/run-codebuild-batch-job.sh
+++ b/.github/scripts/run-codebuild-batch-job.sh
@@ -29,7 +29,7 @@ while [[ $# -gt 0 ]]; do
             shift 2
           fi
           ;;
-        --sdk)
+        --sdk-pr)
           if [[ "$2" == --* ]]; then
             shift 1
           else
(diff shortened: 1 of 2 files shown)


.github/scripts/run-codebuild-batch-job.sh
```diff
@@ -5,11 +5,52 @@ error_exit() {
...
+          ;;
+        --source)
+          if [[ "$2" == --* ]]; then
+            shift 1
+          else
+            SOURCE_VERSION="$2"
+            shift 2
+          fi
+          ;;
+        --sdk)
```
ianbotsf, line 32: **Nit**: The flag names `--sdk` and `--smithy` aren't super clear. I suggest `--sdk-pr` or something similar.

ianbotsf (approved):

lauzadis (approved):

Issue #416, opened 2021-11-12
Track Build Metrics


aajtodd:
It would be nice to track build metrics like artifact size over time (and potentially other things like benchmarks, etc). This is a placeholder ticket for doing this work.

Below are some ideas I've thrown together for inspiration (not necessarily prescriptive).

## Tracking artifact size with gradle task

```kotlin
    tasks.findByName("jvmJar")?.let {
        println("${project.name} has jvmJar")
        tasks.create("metrics") {
            val jvmJar = tasks.getByName<Jar>("jvmJar")
            dependsOn(jvmJar)
            val reportsBuildDir = Paths.get(project.buildDir.absolutePath, "reports", "metrics").toFile()
            val csv = Paths.get(reportsBuildDir.path, "artifacts.csv").toFile()
            outputs.file(csv)

            doLast {
                reportsBuildDir.mkdirs()
                val length = jvmJar.archiveFile.get().asFile.length()
                println("${jvmJar.archiveBaseName.get()} is $length bytes")
                csv.writeText("${jvmJar.archiveBaseName.get()},$length\n")
            }
        }
    }
```

Top level task to aggregate:

```kotlin
tasks.create("metrics") {
    val rootMetrics = this
    val outputFiles = mutableListOf<File>()
    subprojects {
        afterEvaluate {
            val metricsTask = tasks.findByName("metrics")
            if (metricsTask != null) {
                rootMetrics.dependsOn(metricsTask)
                outputFiles.add(metricsTask.outputs.files.singleFile)
            }
        }
    }

    val reportsBuildDir = Paths.get(project.buildDir.absolutePath, "reports", "metrics").toFile()
    val csv = Paths.get(reportsBuildDir.path, "artifacts.csv").toFile()
    outputs.file(csv)
    doLast {
        val content = outputFiles.joinToString(separator = "") { it.readText() }
        csv.writeText(content)
    }
}
```

This generates a file:

```
aws-config,222163
aws-core,9783
aws-endpoint,14982
aws-http,60454
aws-signing,50560
aws-types,17076
crt-util,31746
http-client-engine-crt,56920
testing,6690
aws-event-stream,44084
aws-json-protocols,16550
aws-xml-protocols,44620
```

NOTE: This is some quick and dirty gradle I through together. More than likely we'll want this to be a plugin to handle multiplatform artifacts, etc.

## Tracking in cloudwatch metrics

```
[...]

PR #1230, opened 2024-02-21
chore: track artifact sizes

aajtodd:
## Issue \#

https://github.com/awslabs/aws-sdk-kotlin/issues/416

## Description of changes

* Adds a new task to gather the JVM jar sizes and write them to a JSON file in the format expected by CloudWatch.
* Adds a new task to compare the current sizes to historical data in CW
* Adds a new GH workflow to publish the results of the comparison to the PR.
    * On release the artifact sizes will be updated and new data pushed to CW

### TODO
* Still need to test this in PR
* The workflow should probably fail if we are x % delta above prior averages (or whatever threshold we want to define)
* Manual workflow dispatch would be nice

### Questions
* Is this the "right" way? Should we just use `GetMetricData` and compute average from that
* This doesn't handle any of `smithy-kotlin` runtime directly but it does by computing the dependency closure size of a few services. 
* The dependency closure includes everything needed at runtime (stdlib, smithy-kotlin, aws-runtime + the service JAR size). Is that what we want to track?

### DIFF
--- a/.github/workflows/artifact-size-check.yml
+++ b/.github/workflows/artifact-size-check.yml
@@ -51,7 +51,7 @@ jobs:
         with:
           script: |
             const fs = require('node:fs');
-            const content = fs.readFileSync('./aws-sdk-kotlin/build/reports/metrics/artifact-summary.md');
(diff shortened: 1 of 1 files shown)


lauzadis:
Implemented in #1290

PR #1235, opened 2024-02-27
feat: configuration options for sigv4aSigningRegionSet

### LINKED ISSUE #1234: Multi-Auth and Sigv4A trait implementation

0marperez:
## Issue \#
closes #1234 
SigV4a as a trait is implemented in the following PR: smithy-kotlin#1023

## Description of changes
The `sigv4aSigningRegionSet` config options are added in this PR and are routed to the relevant signer in smithy-kotlin#1043

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -9,8 +9,8 @@ coroutines-version = "1.7.3"
 atomicfu-version = "0.23.1"
 
 # smithy-kotlin codegen and runtime are versioned separately
-smithy-kotlin-runtime-version = "1.0.17"
-smithy-kotlin-codegen-version = "0.30.18"
+smithy-kotlin-runtime-version = "1.0.18"
+smithy-kotlin-codegen-version = "0.30.19"
 
 # codegen
 smithy-version = "1.42.0"


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/sigv4a/ResolveSigningRegionSet.kt
```diff
@@ -0,0 +1,32 @@
...
+        check(isValidListFormat(it)) { "Config setting sigv4aSigningRegionSet is not formatted as a list of strings" }
+        it.split(",")
+    }
+}
+
+/**
+ * Makes sure that a string can be parsed into a list of non-empty strings
+ */
+internal fun isValidListFormat(input: String): Boolean =
+    Regex("^\\s*([^,\\s]+\\s*,\\s*)*[^,\\s]+\\s*\$").matches(input)
```
ianbotsf, line 32: **Question**: Is it _important_ that the list contains no empty strings? Or could they just be filtered out after splitting?

```kotlin
rawString
    ?.split(",")
    ?.map { it.trim() }
    ?.filter { it.isNotEmpty }
```

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/AwsServiceConfigIntegration.kt
```diff
@@ -116,6 +116,17 @@ class AwsServiceConfigIntegration : KotlinIntegration {
...
             )
         }
+
+        val Sigv4aSigningRegionSet: ConfigProperty = ConfigProperty {
+            name = "sigv4aSigningRegionSet"
+            symbol = KotlinTypes.Collections.list(KotlinTypes.String).asNullable()
+            baseClass = AwsRuntimeTypes.Core.Client.AwsSdkClientConfig
+            useNestedBuilderBaseClass()
+            documentation = """
+                The set of regions to use when signing a request with sigV4a. If not provided this will automatically be set by the SDK.
```
ianbotsf, line 126: **Nit**: `sigV4a` → `SigV4a`

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/AwsServiceConfigIntegration.kt
```diff
@@ -116,6 +116,17 @@ class AwsServiceConfigIntegration : KotlinIntegration {
...
+            """.trimIndent()
+            order = 100
```
ianbotsf, line 128: **Question**: Why `order = 100`? If it's important that this appears before/after something else, we should leave a clarifying comment.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/sigv4a/ResolveSigningRegionSet.kt
```diff
@@ -0,0 +1,32 @@
...
+ * @return The sigv4aSigningRegionSet if found, null if not
+ */
+@InternalSdkApi
+public suspend fun resolveSigningRegionSet(platform: PlatformProvider = PlatformProvider.System, profile: LazyAsyncValue<AwsProfile>): List<String>? {
+    val rawString = AwsSdkSetting.AwsSigv4aSigningRegionSet.resolve(platform) ?: profile.get().sigv4aSigningRegionSet
+    return rawString?.let {
+        check(isValidListFormat(it)) { "Config setting sigv4aSigningRegionSet is not formatted as a list of strings" }
+        it.split(",")
+    }
+}
```
ianbotsf, line 26: **Style**: Since this is a region _set_ it feels odd to pass it around in a `List`, especially since the signer implementation in **smithy-kotlin** will turn it into a `Set` eventually. I suggest making the type `Set<String>` in config, resolution, etc.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/AwsServiceConfigIntegration.kt
```diff
@@ -116,6 +116,17 @@ class AwsServiceConfigIntegration : KotlinIntegration {
...
+        }
```
ianbotsf, line 129: **Correctness**: This adds the new property for every service client whether that service supports SigV4a or not. I think we should scope this more narrowly to only services which _actually_ use it—by model trait or known list.

ianbotsf (changes requested):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/sigv4a/ResolveSigningRegionSet.kt
0marperez: The design document says it "must be an optional, non-empty list of non-empty strings". I don't think it matters much but I would like to have consistency with the other SDK's.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/sigv4a/ResolveSigningRegionSet.kt
0marperez: This is also something that I did because of the design document. The name doesn't make any sense but I think it would be nice to have consistency with the other SDK's

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/AwsServiceConfigIntegration.kt
0marperez: This was just a copy paste leftover oops. The order shouldn't matter here

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AwsSdkSetting.kt
```diff
@@ -202,6 +202,12 @@ public object AwsSdkSetting {
      */
     public val AwsRequestMinCompressionSizeBytes: EnvironmentSetting<Long> =
         longEnvSetting("aws.requestMinCompressionSizeBytes", "AWS_REQUEST_MIN_COMPRESSION_SIZE_BYTES")
+
+    /**
+     * The set of regions to use when signing a request with sigV4a.
+     */
+    public val AwsSigv4aSigningRegionSet: EnvironmentSetting<String> =
+        strEnvSetting("aws.sigv4aSigningRegionSet", "AWS_SIGV4A_SIGNING_REGION_SET")
```
lauzadis, line 210: naming: `aws.sigV4aSigningRegionSet` (capital V)

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/sigv4a/ResolveSigningRegionSet.kt
```diff
@@ -0,0 +1,32 @@
...
+import aws.sdk.kotlin.runtime.InternalSdkApi
+import aws.sdk.kotlin.runtime.config.AwsSdkSetting
+import aws.sdk.kotlin.runtime.config.profile.AwsProfile
+import aws.sdk.kotlin.runtime.config.profile.sigv4aSigningRegionSet
+import aws.smithy.kotlin.runtime.config.resolve
+import aws.smithy.kotlin.runtime.util.LazyAsyncValue
+import aws.smithy.kotlin.runtime.util.PlatformProvider
+
+/**
+ * Attempts to resolve sigv4aSigningRegionSet from specified sources.
```
lauzadis, line 16: nit/suggestion: be more specific about "from specified sources"

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/sigv4a/ResolveSigningRegionSet.kt
lauzadis: I'd also vote to go against the spec here and make it a `Set<String>`. I had [the same comment](https://github.com/awslabs/smithy-kotlin/pull/1043#discussion_r1507918750) in your smithy-kotlin PR and didn't realize it was discussed here already. 

This is probably something we should discuss and see if we can get the spec changed

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/sigv4a/ResolveSigningRegionSet.kt
lauzadis, line 32: suggestion: I'd recommend using Kotlin functions to filter this list instead of a regex, it's way more maintainable.

lauzadis: I think the spec just calls out what you can expect the values to be. I don't think we have to validate + blow up / throw an exception if the set has an empty string in it

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AbstractAwsSdkClientFactory.kt
```diff
@@ -79,6 +80,7 @@ public abstract class AbstractAwsSdkClientFactory<
             config.useFips = config.useFips ?: resolveUseFips(profile = profile)
             config.useDualStack = config.useDualStack ?: resolveUseDualStack(profile = profile)
             config.applicationId = config.applicationId ?: resolveUserAgentAppId(platform, profile)
+            config.sigv4aSigningRegionSet = config.sigv4aSigningRegionSet ?: resolveSigningRegionSet(platform, profile)
```
aajtodd, line 83: **correctness**: This doesn't apply to every service, only services using sigv4a. We shouldn't register config properties that are unused/meaningless.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AwsSdkSetting.kt
aajtodd: Whatever we use for JVM system properties needs to be consistent with Java. I don't see a [setting](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/SdkSystemSetting.java) defined by them yet so talk with someone on Java team to confirm.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/sigv4a/ResolveSigningRegionSet.kt
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.sdk.kotlin.runtime.config.sigv4a
```
aajtodd, line 5: I'm not a fan of proliferating packages with single things in them. We should have grouped resolvers somewhere probably but we can at least do better going forward. This should be either in [ResolveRegion](https://github.com/awslabs/aws-sdk-kotlin/blob/main/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/region/ResolveRegion.kt) or in a more generic package like `aws.sdk.kotlin.runtime.config.auth` such that new auth related settings/resolvers end up grouped.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/sigv4a/ResolveSigningRegionSet.kt
aajtodd: Agreed, I think being tolerant here is probably a better experience.

aws-runtime/aws-core/common/src/aws/sdk/kotlin/runtime/client/AwsSdkClientConfig.kt
```diff
@@ -46,6 +46,11 @@ public interface AwsSdkClientConfig : SdkClientConfig {
      */
     public val applicationId: String?
 
+    /**
+     * The set of regions to use when signing a request with sigV4a. If provided it will override endpoints metadata.
+     */
+    public val sigv4aSigningRegionSet: List<String>?
```
aajtodd, line 52: **fix**: Again this doesn't belong here, this should be registered as part of the integration. We can define interfaces for it if we want/need to but the config setting should be added as part of codegen.

aajtodd (changes requested):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AwsSdkSetting.kt
ianbotsf: The Java SDK has not yet implemented this feature and doesn't have a system property name selected yet so they will use whatever we decide in this instance. Given that their typical pattern is a camel-cased version of the environment variable name, I think `aws.sigV4aSigningRegionSet` is good and probably what they would've selected anyway.

aajtodd (approved):

lauzadis (approved):

PR #1245, opened 2024-03-06
fix: add UnsupportedSigningAlgorithmInterceptor to all sigV4a services

0marperez:
## Description of changes
The `UnsupportedSigningAlgorithmInterceptor`, which deals with the default signer not supporting sigV4a was only being applied to a hard coded list of services, now the interceptor is applied based on the sigV4a trait.

### DIFF
--- a/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/UnsupportedSigningAlgorithmIntegration.kt
+++ b/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/UnsupportedSigningAlgorithmIntegration.kt
@@ -21,19 +21,17 @@ import software.amazon.smithy.model.shapes.OperationShape
  * See: [aws.sdk.kotlin.runtime.http.interceptors.UnsupportedSigningAlgorithmInterceptor]
  */
 class UnsupportedSigningAlgorithmIntegration : KotlinIntegration {
-
-    // Needs to happen after the SigV4AsymmetricTraitCustomization (-60), which adds the sigV4a trait to models missing it
-    override val order: Byte = -50
-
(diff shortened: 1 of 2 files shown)


ianbotsf (approved):

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/UnsupportedSigningAlgorithmIntegration.kt
```diff
@@ -5,30 +5,35 @@
...
-            "s3", "eventbridge", "cloudfront keyvaluestore" -> true
-            else -> false
-        }
+
+    // Needs to happen after the SigV4AsymmetricTraitCustomization (-60), which adds the sigV4a trait to models missing it
+    override val order: Byte = -50
+
+    // Set to true rather than looking for sigV4a in the model here because this integration would be filtered out
+    // before the application of the trait is done for some services via SigV4AsymmetricTraitCustomization.
+    // See: CodegenVisitor
```
lauzadis, line 30: This should be a simple fix and I see we have a FIXME already in `CodegenVisitor`, if you have the time I would recommend including it in this PR. Then we don't need the hardcoded `true` and we can tighten the conditional.

`CodegenVisitor.kt`
```
        integrations = ServiceLoader.load(KotlinIntegration::class.java, classLoader)
            .onEach { integration -> LOGGER.info("Loaded KotlinIntegration: ${integration.javaClass.name}") }
            .sortedBy(KotlinIntegration::order)
...
        for (integration in integrations) {
            if (!integration.enabledForService(resolvedModel, settings)) { continue }
            resolvedModel = integration.preprocessModel(resolvedModel, settings)
        }
```

lauzadis (approved):

PR #1246, opened 2024-03-06
fix: add cloudfront keyvaluestore to sigV4a trait customization

0marperez:
## Description of changes
Added `cloudfront keyvaluestore` to the `SigV4AsymmetricTraitCustomization`


ianbotsf (approved):

lauzadis (changes requested):
This trait is already added in [BackfillSigV4ACustomization](https://github.com/awslabs/aws-sdk-kotlin/blame/main/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/cloudfrontkeyvaluestore/BackfillSigV4ACustomization.kt), but to the front of the auth trait list. 

We'd need to investigate whether it should be added to the front or back (probably back) and make the appropriate change. But with this change, it will be added to both the front *and* the back.

PR #1247, opened 2024-03-07
misc: bump smithy to 1.45.0

milesziemer:
Also exclude a new protocol test which is failing in the sdk. The test fails because it expects a blob members with a default of "" to have the default populated when a value isn't explicitly provided, but the sdk ignores defaults of "" for blob members. Fixing this is non-trivial because using the default now would change the nullability of members with this default (nullable -> non-nullable). Keeping blobs with "" default nullable but using the default also doesn't work because it means that migrating @required -> @default("") would be a breaking change (granted, it still is now). I can open an issue describing this behavior as well if needed.


codegen/smithy-aws-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/aws/protocols/core/AwsHttpBindingProtocolGenerator.kt
```diff
@@ -39,6 +39,11 @@ abstract class AwsHttpBindingProtocolGenerator : HttpBindingProtocolGenerator()
         // The following can be used to generate only a specific test by name.
         // val targetedTest = TestMemberDelta(setOf("RestJsonComplexErrorWithNoMessage"), TestContainmentMode.RUN_TESTS)
 
+        val ignoredTests = TestMemberDelta(setOf(
+            // This test requires populating blob members with a default value of "", which the sdk doesn't do
```
lauzadis, line 43: Yeah, opening an issue and linking it here would be appreciated!

lauzadis (approved):

0marperez (approved):

Issue #76, opened 2021-03-05
DynamodDB Enhanced Client Design


aajtodd:
Consider implementing an "enhanced" DDB client like offered by [Java](https://aws.amazon.com/blogs/developer/introducing-enhanced-dynamodb-client-in-the-aws-sdk-for-java-v2/)

Some considerations:
* We might be able to do this as a compiler plugin and generate rather optimized code so that we don't have to rely on annotation processing (which I believe is JVM only at the moment) OR on reflection capabilities.

ianbotsf:
Resolving in favor of #472.

PR #1266, opened 2024-03-22
implement mapper pipeline

ianbotsf:
## Issue \#

Part of #76 

## Description of changes

This changes adds an implementation of a pipeline for orchestrating/executing mapper operations and interceptors. The `Interceptor` interface KDocs should (hopefully) describe well enough how interceptors, hooks, and the overall pipeline works...if not, leave comments so that I can fix that!

### Feedback solicited

Any modified file is open for feedback except where `TODO`s disclaim parts of the code. Things I'd particularly like feedback on include:
* **The implementation of `Operation`**: I considered and partly implemented a higher-level abstraction of this which involved modeled stages (e.g., `InitializationStage`, `SerializationStage`, etc.) which combined into a pipeline inside `Operation`. It felt too complex though and introduced even more generics than are present already. That's not to say it's definitively wrong, though, or that my implementation couldn't be clarified in other ways.
* **The ordering of the pipeline and interceptors**: I designed the pipeline to have distinct steps (e.g., initialization, serialization, etc.) surrounded by phases during which interceptor hooks would run. I wanted read and modify hooks before each step and a read hook after. Interceptors will accumulate errors in read hooks but not modify hooks...those halt subsequent interceptor modify hooks and proceed to the next read hook stage. After a read stage terminates, any present exception gets thrown.
  
  This design leads to a situation which may limit its usefulness: Any exception thrown in a modify hook, read hook, or pipeline step will not have an opportunity to be caught by a subsequent modify hook and changed. Maybe some logic could convert serialization or invocation exceptions into some result representing success?
* **The number/type of generics**: `Interceptor`, all the context classes, and `Operation` are awash in generic type parameters now. I _think_ they make sense but additional eyes will confirm/disconfirm.

### Other minor changes

* Separated `Table` interface into `TableSpec` and `TableOperations`. This had a convenient property of minimizing the context that interceptors and the pipeline needed to include (e.g., mocking an `Operation` no longer requires a mock `Table` implementation including all operations). It also addresses what I believe will be a future codegen concern...namely, having a dedicated space (`TableOperations`) in which to generate operation methods (e.g., `getItem`, `createTable`, etc.) that's separate from the _hand-maintained_ parts of table metadata (`TableSpec`).
* Made a mapper's config a `public val` on the mapper itself. This will allow callers to access values they passed in to us such as the underlying DDB client, the list of interceptors, etc. I _think_ this is okay but let me know if there are concerns.
* Promoted `keyAttributeNames` from an implementation detail to a `public val` in `KeySpec`. It simplifies connecting operations to the pipeline and may also be useful for introspection by users.

### DIFF
--- a/hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/Interceptor.kt
+++ b/hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/Interceptor.kt
@@ -45,8 +45,8 @@ import aws.sdk.kotlin.services.dynamodb.model.GetItemResponse as LowLevelGetItem
  * interceptors hooks in the same phase. Only after _all_ interceptors' read-only hooks for a given phase have completed
  * will any exception be thrown. For example, if a mapper has two interceptors **A** and **B** registered, and **A**'s
  * [readAfterSerialization] hook throws an exception, it will be added to the context passed to **B**'s
- * [readAfterSerialization] hook. After **B**'s [readAfterSerialization] hook has completed, the exception be thrown
- * back to the caller.
+ * [readAfterSerialization] hook. After **B**'s [readAfterSerialization] hook has completed, the exception will be
+ * thrown back to the caller.
  *
  * **Modify hooks** are invoked before each step in the pipeline (except _before_ **Initialization**). They offer the
  * ability to see and modify a high-level operation in progress. They can be used to customize behavior and data in ways
@@ -178,6 +178,6 @@ public interface Interceptor<T, HReq, LReq, LRes, HRes> {
 }
 
 /**
- * A "universal" interceptor which acts on any type of objects, requests, and responses
+ * A universal interceptor which acts on any type of high-level objects, requests, and responses
  */
-public typealias UniversalInterceptor = Interceptor<*, *, *, *, *>
+public typealias InterceptorAny = Interceptor<*, *, *, *, *>
--- a/hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/SerializeInput.kt
+++ b/hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/SerializeInput.kt
@@ -14,22 +14,6 @@ import aws.sdk.kotlin.hll.dynamodbmapper.pipeline.internal.SerializeInputImpl
  * @param HReq The type of high-level request object (e.g., [GetItemRequest])
  */
 public interface SerializeInput<T, HReq> {
-    public companion object {
-        /**
-         * Creates a new [SerializeInput]
-         * @param T The type of objects being converted to/from DynamoDB items
-         * @param HReq The type of high-level request object (e.g., [GetItemRequest])
-         * @param highLevelRequest The high-level request object which is to be serialized into a low-level request
-         * object
-         * @param serializeSchema The [ItemSchema] to use for serializing objects into items
-         */
-        public operator fun <T, HReq> invoke(
-            highLevelRequest: HReq,
-            serializeSchema: ItemSchema<T>,
-        ): SerializeInput<T, HReq> =
-            SerializeInputImpl(highLevelRequest, serializeSchema)
-    }
-
     /**
      * The high-level request object which is to be serialized into a low-level request object
      */
@@ -40,3 +24,13 @@ public interface SerializeInput<T, HReq> {
      */
     public val serializeSchema: ItemSchema<T>
 }
+
+/**
+ * Creates a new [SerializeInput]
+ * @param T The type of objects being converted to/from DynamoDB items
+ * @param HReq The type of high-level request object (e.g., [GetItemRequest])
+ * @param highLevelRequest The high-level request object which is to be serialized into a low-level request object
+ * @param serializeSchema The [ItemSchema] to use for serializing objects into items
+ */
+public fun <T, HReq> SerializeInput(highLevelRequest: HReq, serializeSchema: ItemSchema<T>): SerializeInput<T, HReq> =
+    SerializeInputImpl(highLevelRequest, serializeSchema)
--- a/hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/Operation.kt
+++ b/hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/Operation.kt
@@ -12,7 +12,7 @@ internal class Operation<T, HReq, LReq, LRes, HRes>(
     private val serialize: (HReq, ItemSchema<T>) -> LReq,
     private val lowLevelInvoke: suspend (LReq) -> LRes,
     private val deserialize: (LRes, ItemSchema<T>) -> HRes,
-    interceptors: Collection<UniversalInterceptor>,
+    interceptors: Collection<InterceptorAny>,
 ) {
     private val interceptors = interceptors.map {
         // Will cause runtime ClassCastExceptions during interceptor invocation if the types don't match. Is that ok?
--- a/hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/PipelineStage.kt
+++ b/hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/PipelineStage.kt
@@ -1,5 +1,0 @@
-/*
- * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
- * SPDX-License-Identifier: Apache-2.0
- */
-package aws.sdk.kotlin.hll.dynamodbmapper.pipeline.internal
--- a/hll/ddb-mapper/dynamodb-mapper/api/dynamodb-mapper.api
+++ b/hll/ddb-mapper/dynamodb-mapper/api/dynamodb-mapper.api
@@ -7,8 +7,8 @@ public abstract interface class aws/sdk/kotlin/hll/dynamodbmapper/DynamoDbMapper
 }
 
 public final class aws/sdk/kotlin/hll/dynamodbmapper/DynamoDbMapper$Companion {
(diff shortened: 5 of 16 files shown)


hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/SerializeInput.kt
```diff
@@ -0,0 +1,42 @@
...
+    public companion object {
+        /**
+         * Creates a new [SerializeInput]
+         * @param T The type of objects being converted to/from DynamoDB items
+         * @param HReq The type of high-level request object (e.g., [GetItemRequest])
+         * @param highLevelRequest The high-level request object which is to be serialized into a low-level request
+         * object
+         * @param serializeSchema The [ItemSchema] to use for serializing objects into items
+         */
+        public operator fun <T, HReq> invoke(
```
aajtodd, line 26: **style**: Why `invoke` vs a [factory function](https://kotlinlang.org/docs/coding-conventions.html#function-names)

```kotlin
fun <T, HReq> SerializeInput(highLevelRequest: HReq, serializeScheme: ItemSchemea<T>): SerializeInput = SerializeInputImpl(highLevelRequest, serializeScheme)
```

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/Interceptor.kt
```diff
@@ -4,49 +4,180 @@
...
+    // -----------------------------------------------------------------------------------------------------------------
 
-    public fun readAfterInvocation(ctx: LResContext<T, HReq, LReq, LRes>)
-    public fun modifyBeforeDeserialization(ctx: LResContext<T, HReq, LReq, LRes>): DeserializeInput<T, LReq>
-    public fun readBeforeDeserialization(ctx: LResContext<T, HReq, LReq, LRes>)
+    /**
+     * A **read-only** hook that runs _after_ the **Initialization** step
+     * @param ctx Context containing the high-level request
+     */
+    public fun readAfterInitialization(ctx: HReqContext<T, HReq>) { }
```
aajtodd, line 88: **Question**: What happens in "initialization"? `after` implies to me something happened before this but if this is the first step before _anything_ else happens then I think `readBeforeExecution` was preferable.

hll/ddb-mapper/dynamodb-mapper/build.gradle.kts
```diff
@@ -10,5 +10,12 @@ kotlin {
                 api(project(":services:dynamodb"))
             }
         }
+
+        commonTest {
+            dependencies {
+                implementation(libs.mockk)
```
aajtodd, line 16: **concern**: mockk does not support KMP really from what I can see. This will make many tests JVM only effectively and perhaps result in some refactoring later to move test from `commonTest` -> `jvmTest`.


I think some tests being JVM only are likely ok since verification of behavior of common code in one platform source set is still pretty good coverage (any issues from that point would indicate the compiler has a bug I would think).

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/PipelineStage.kt
```diff
@@ -0,0 +1,5 @@
+/*
```
aajtodd, line 1: **Question**: What is the purpose of this file?

aajtodd (commented):
Looks good overall! Nice work.

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/Operation.kt
```diff
@@ -0,0 +1,112 @@
...
+
+internal class Operation<T, HReq, LReq, LRes, HRes>(
+    private val initialize: (HReq) -> HReqContextImpl<T, HReq>,
+    private val serialize: (HReq, ItemSchema<T>) -> LReq,
+    private val lowLevelInvoke: suspend (LReq) -> LRes,
+    private val deserialize: (LRes, ItemSchema<T>) -> HRes,
+    interceptors: Collection<UniversalInterceptor>,
+) {
+    private val interceptors = interceptors.map {
+        // Will cause runtime ClassCastExceptions during interceptor invocation if the types don't match. Is that ok?
```
lauzadis, line 18: I think it's ok, but why can't we have the constructor parameter `interceptors` be the narrower type  `Collection<Interceptor<T, HReq, LReq, LRes, HRes>>`?

lauzadis (approved):

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/Interceptor.kt
```diff
@@ -4,49 +4,180 @@
...
+ * [modifyBeforeSerialization] hook will not be invoked. Interceptors **A** and **B**'s [readAfterSerialization] hook
+ * will execute, after which the exception will be thrown back to the caller.
+ *
+ * # Registration and execution order
+ *
+ * Interceptors are registered on [DynamoDbMapper] as configuration. Multiple interceptors may be registered on a single
+ * mapper. The order in which interceptors are given in mapper config determines the order in which they will be
+ * executed:
+ * * For phases _before_ the **Low-level invocation** step, hooks will be executed _in given order_
+ * * For phases _after_ the **Low-level invocation** step, hooks will be executed _in reverse order_
```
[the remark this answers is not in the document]
ianbotsf: The intent is to support predictable lifetimes for interceptor events. (Whether it's reasonable to do so is still up for debate.) For an example, suppose there are two interceptors registered for an operation, **A** and **B** in that order.

* **A** has:
  * A `modifyBeforeSerialization` hook that normalizes timestamps in the high-level request from local times to UTC
  * A `modifyBeforeCompletion` hook that localizes timestamps in the high-level response from UTC to local times
* **B** has:
  * A `modifyBeforeSerialization` hook that logs the high-level request
  * A `modifyBeforeCompletion` hook that logs the high-level response

The current logic would run them in this order:

* `A.modifyBeforeSerialization` (times to UTC)
* `B.modifyBeforeSerialization` (log request with UTC timestamp)
* ...
* `B.modifyBeforeCompletion` (log response with UTC timestamp)
* `A.modifyBeforeCompletion` (times to local)

This sequence allows interceptors to rely on the ordering of "set up" and "tear down" events (e.g., **B** knows that whatever **A** has changed on the way "down" will be undone in a predictable order on the way back "up").

I will say that this execution order differs from what we do in low-level interceptors. Those all run in registration sequence regardless of where they sit in the pipeline. This divergence may be fine or it may confuse people, although I'd say that registering custom interceptors in the low-level _and_ high-level clients is a pretty advanced use case.

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/Interceptor.kt
ianbotsf: "Initialization" refers to the `Operation.doInitialize` method, which delegates to the `initialize` constructor parameter passed in by individual operations. My strawman `GetItem` implementation uses `initialize` to create the initial `HReqContext` (see **GetItem.kt:62**). I imagine it's _possible_ that some operations we codegen may need additional initialization steps but I cannot say for sure.

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/Interceptor.kt
```diff
@@ -4,49 +4,180 @@
...
+    // -----------------------------------------------------------------------------------------------------------------
+    // Completion step
+    // -----------------------------------------------------------------------------------------------------------------
 }
 
-public interface MapperContext // TBD
+/**
+ * A "universal" interceptor which acts on any type of objects, requests, and responses
+ */
+public typealias UniversalInterceptor = Interceptor<*, *, *, *, *>
```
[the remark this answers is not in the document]
ianbotsf: Ooh yes, that's much better.

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/SerializeInput.kt
ianbotsf: No particular reason so I'm not tied to the choice. Are there any tradeoffs between `invoke` vs factory function?

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/Operation.kt
ianbotsf: It might be able to but then the business of converting to more specific types would fall to the callers of `Operation`. Interceptors are registered at the mapper level right now and can apply to any operation. Consequently, the mapper-level config for interceptors is also a collection of `InterceptorAny`.

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/PipelineStage.kt
ianbotsf: 😖 Accidentally included file from my aborted attempt to further abstract the pipeline. Will remove!

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/Interceptor.kt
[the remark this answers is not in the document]
lauzadis: Ok that's a good example where this is nice to have. It stood out to me because it's inconsistent with the low-level interceptors, but since it's an advanced use case, I think it's ok

hll/ddb-mapper/dynamodb-mapper/build.gradle.kts
aajtodd: Option 1 seems fine to me.

PR #1270, opened 2024-03-29
misc: replace Snyk with GitHub Dependabot

lauzadis:
This PR replaces use of Snyk with GitHub Dependabot. Dependabot is already enabled on our repositories but does not track our Gradle dependencies by default. 

To enable tracking those, we need to [manually submit our dependencies to Dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#gradle) for scanning, which is done using [a GitHub Action called `gradle/actions/dependency-submission@v3`](https://github.com/marketplace/actions/build-with-gradle#the-dependency-submission-action). 

This action only needs to run on pushes to the `main` branch. Running the action on any other branch will result in a no-op with this message: `The snapshot was accepted, but it is not for the default branch. It will not update dependency results for the repository. `

## Description of changes

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,7 +1,8 @@
 name: Dependabot Dependency Submission
 
-on: [ push ]
-# branches: [ main ]  # TODO Uncomment after testing. We only want to run Dependabot dependency submission on pushes to main
+on:
+  push:
+    branches: [ main ]
 
 permissions:
   contents: write


.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,17 @@
+name: Dependabot Dependency Submission
+
+on:
+  push:
+    branches: [ main ]
+
```
0marperez, line 6: question: we're not running this during pull requests like snyk, was that intentional?

lauzadis: Yeah, the PR description explains this, Dependabot only accepts dependency submissions from the default branch (`main`)

ianbotsf (approved):

PR #1279, opened 2024-04-10
fix: remove `bootstrapAll` task from service benchmarks project; move project inclusion logic to root settings.gradle.kts

ianbotsf:
## Issue \#

(none)

## Description of changes

The `bootstrapAll` task of **:tests:benchmarks:service-benchmarks** was broken during the refactor of our codegen bootstrapping logic into **aws-kotlin-repo-tools**. Rather than repair it, this change removes the ability to automatically bootstrap altogether. Instead, now the decision of whether to include the service benchmarks project lies down in **settings.gradle.kts**. This more closely matches how HLLs will be included/excluded based on bootstrapped service clients.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -41,22 +41,40 @@ include(":aws-runtime:aws-endpoint")
 include(":aws-runtime:aws-http")
 include(":services")
 include(":tests")
-include(":tests:benchmarks:service-benchmarks")
 include(":tests:codegen:event-stream")
 include(":tests:e2e-test-util")
 
 // generated services
-fun File.isServiceDir(): Boolean {
-    if (isDirectory) {
-        return toPath().resolve("build.gradle.kts").toFile().exists()
-    }
-    return false
-}
+val File.isServiceDir: Boolean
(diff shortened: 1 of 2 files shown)


settings.gradle.kts
```diff
@@ -41,24 +41,42 @@ include(":aws-runtime:aws-endpoint")
...
 file("services").listFiles().forEach {
-    if (it.isServiceDir()) {
+    if (it.isServiceDir) {
         include(":services:${it.name}")
     }
 }
 
+// Service benchmarks project
+val benchmarkServices = listOf(
+    // keep this list in sync with tests/benchmarks/service-benchmarks/build.gradle.kts
```
lauzadis, line 62: I was looking at a way to make this list sharable between `settings.gradle.kts` and `build.gradle.kts`, but couldn't find anything. I was hoping the `by settings` delegated property would work but it doesn't seem to be what we want here. https://docs.gradle.org/current/userguide/kotlin_dsl.html#project_properties

lauzadis (approved):

0marperez (approved):

PR #1282, opened 2024-04-15
fix: respect AWS_ENDPOINT_URL_SERVICE environment variables in service client based identity providers

### LINKED ISSUE #1210: StsWebIdentityCredentialsProvider does not respect AWS_ENDPOINT_URL_STS
### Describe the bug

The `StsWebIdentityCredentialsProvider` that is part of the default credentials provider chain does not respect the `AWS_ENDPOINT_URL_STS` environment variable when fetching the credentials from STS.
Instead it will simply call the default AWS URL.

### Expected behavior

When setting `AWS_ENDPOINT_URL_STS` I expect the SDK to use the specified STS URL for fetching credentials.

### Current behavior

It uses the default AWS URL https://sts.us-west-2.amazonaws.com

### Steps to Reproduce

1. Set the `AWS_ENDPOINT_URL_STS` environment variable to a URL that should be called
2. Set `AWS_WEB_IDENTITY_TOKEN_FILE` to a file containing some JWT
3. Set `AWS_ROLE_ARN` to any valid role ARN
4. Use SDK to call any service via STS Web Identity Token 
5. See that the URL specified above is not used for authentication

### Possible Solution

When [creating the STS client](https://github.com/awslabs/aws-sdk-kotlin/blob/6a9c91ffae83c65c24a5fc129eeff011c4d43c10/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/StsWebIdentityCredentialsProvider.kt#L119) the `finalizeConfig` function of the `StsClient` is not called, which would take the endpointUrl from the environment variables.

Alternatively, allow passing my own StsClient to the credentials provider (this is possible in Java SDK)

### Context

Our backend services use a service-mesh to call AWS services and we can only call them via HTTP, not HTTPS. Thus, we have to change the endpointUrl to use HTTP.

1.0.51

### Platform (JVM/JS/Native)

JVM

MacOS 14.3 (M1)

0marperez:
## Issue \#
closes #1210
## Description of changes
By changing the STS clients to use `fromEnvironment` the `AWS_ENDPOINT_URL_STS` environment variable is recognized, fixing the issue. Same for SSO clients and `AWS_ENDPOINT_URL_SSO`, `AWS_ENDPOINT_URL_SSO_OIDC `

### DIFF
--- a/.changes/7699105e-2f52-4239-abff-631aa6dba005.json
+++ b/.changes/7699105e-2f52-4239-abff-631aa6dba005.json
@@ -1,7 +1,7 @@
 {
     "id": "7699105e-2f52-4239-abff-631aa6dba005",
     "type": "bugfix",
-    "description": "STS based identity providers now respect AWS_ENDPOINT_URL_STS environment variable",
+    "description": "Service client based identity providers now respect AWS_ENDPOINT_URL_SERVICE environment variables",
     "issues": [
         "awslabs/aws-sdk-kotlin#1210"
     ]
--- a/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/SsoCredentialsProvider.kt
+++ b/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/SsoCredentialsProvider.kt
@@ -98,7 +98,7 @@ public class SsoCredentialsProvider public constructor(
         }
 
         val telemetry = coroutineContext.telemetryProvider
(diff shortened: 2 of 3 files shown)


ianbotsf (approved):

lauzadis (changes requested):
Can you also do the same thing for the `SsoOidcClient` in the `SsoTokenProvider`?

ianbotsf (approved):

lauzadis (approved):

PR #1284, opened 2024-04-16
fix: apply clock skew interceptor to clients created via `invoke`

### LINKED ISSUE #1211: Clock skew interceptor is only applied when constructing clients via `fromEnvironment`
### Describe the bug

The `ClockSkewInterceptor` is only applied via `finalizeConfig` which is only invoked when constructing clients via `fromEnvironment`

### Expected behavior

`ClockSkew` should always be applied when constructing a client.

### Current behavior

Clock skew may or may not be applied

### Steps to Reproduce

N/A

### Possible Solution

_No response_

### Context

_No response_

1.0.51

### Platform (JVM/JS/Native)

N/A

N/A

ianbotsf:
## Issue \#

Addresses #1211

## Description of changes

This change adds the clock skew interceptor to clients created via `invoke` (as opposed to `fromEnvironment`). It pulls in upstream changes from **smithy-kotlin** which add a `finalizeConfig` method to the abstract superclass from which `AbstractAwsSdkClientFactory` extends. This causes a naming conflict so the SDK's `suspend` `finalizeConfig` method has been renamed `finalizeEnvironmentalConfig`.

**Companion PR**: https://github.com/smithy-lang/smithy-kotlin/pull/1067

### DIFF
--- a/.changes/f164451a-1adc-4220-9a7d-1765959c23d3.json
+++ b/.changes/f164451a-1adc-4220-9a7d-1765959c23d3.json
@@ -1,8 +1,9 @@
 {
     "id": "f164451a-1adc-4220-9a7d-1765959c23d3",
     "type": "bugfix",
-    "description": "Apply clock skew interceptor to clients created via `invoke`",
+    "description": "⚠️ **IMPORTANT**: Apply clock skew interceptor to clients created via `invoke`",
     "issues": [
         "awslabs/aws-sdk-kotlin#1211"
-    ]
+    ],
+    "requiresMinorVersionBump": true
 }


ianbotsf:
FYI I added the **ackhowledge-api-break** label to this PR because I think the breaks are safe but please chime in if you have differing opinions.

codegen/aws-sdk-codegen/src/test/kotlin/aws/sdk/kotlin/codegen/ServiceClientCompanionObjectWriterTest.kt
```diff
@@ -23,18 +23,14 @@ class ServiceClientCompanionObjectWriterTest {
         ServiceClientCompanionObjectWriter().write(writer, null)
 
         val expected = """
-            public companion object : AbstractAwsSdkClientFactory<Config, Config.Builder, TestGeneratorClient, Builder>() {
```
lauzadis, line 26: This is a test for the companion object but there's no longer a companion object being generated, is that because there's no protocol? Can we add a protocol or maybe rename the test?

lauzadis (approved):
Approved, but let's talk about the API break

codegen/aws-sdk-codegen/src/test/kotlin/aws/sdk/kotlin/codegen/ServiceClientCompanionObjectWriterTest.kt
ianbotsf: It's no longer `ServiceClientCompanionObjectWriter`'s job to _create_ the companion object, merely to _append_ things to sections of it, which the updated test reflects. In my view the class and test names are still correct.

ianbotsf:
Discussed offline and this change is breaking enough that we don't want to push it directly to **main**. Updated base branch to **v1.2**. Will pursue a temporary fix strictly within **aws-sdk-kotlin** codegen until such time as we have another minor version bump.

PR #1290, opened 2024-04-24
feat: artifact size metrics

0marperez:
## Issue \#
N/A

## Description of changes
Added artifact size metrics plugin and misc changes to support the addition.

**_Note: The plugin currently won't run correctly because of the database being bootstrapped in my personal AWS account and me being unable to change a repository secret to point to my AWS account because it would break CI._**

I plan to fix this by merging the changes that add the actual database first then re-running CI.


.github/workflows/codebuild-ci.yml
```diff
@@ -78,14 +83,112 @@ jobs:
...
+            --external-contributor-smithy-pr-number ${{ inputs.smithy-kotlin-pr }}
+      - name: Cancel build
+        if: ${{ cancelled() }}
+        env:
+          BUILD_ID: ${{ steps.svc-check-batch.outputs.aws-build-id }}
+        run: |
+          if [ ! -z "$BUILD_ID" ]; then
+            echo "cancelling in-progress batch build: id=$BUILD_ID"
+            aws codebuild stop-build --id $BUILD_ID
+          fi
```
ianbotsf, line 106: **Question**: This is great! 🤩 Any idea why weren't we doing this before?

ianbotsf (approved):
See also comments on https://github.com/awslabs/aws-crt-kotlin/pull/102

.github/workflows/codebuild-ci.yml
0marperez: This is all [Aaron](https://github.com/awslabs/aws-sdk-kotlin/blame/main/.github/workflows/codebuild-ci.yml#L89-L97), I just kept it and now it looks like I wrote it😅

lauzadis: `git blame` remembers all 😆

ianbotsf: Ah okay, the diff was so long and oddly formatted so I didn't notice where they were before. 👍

0marperez: Git makes it more confusing than it has to sometimes😵‍💫 It "moved" the original block down here:
https://github.com/awslabs/aws-sdk-kotlin/blob/ac643dc3f502da3436cfb1cdee4294696ec81235/.github/workflows/codebuild-ci.yml#L223-L231
When that block is actually the new one

ianbotsf (approved):

Issue #1330, opened 2024-06-13
DynamoDbClient.scanPaginated ignores exclusiveStartKey


thake:
### Describe the bug

The method DynamoDbClient.scanPaginated ignores the initial exclusiveStartKey.

### Expected behavior

The inital passed exclusiveStartKey is honored.

### Current behavior

The inital passed exclusiveStartKey is ignored.

### Steps to Reproduce

```kotlin 
val table = "table-to-scan"
val lastKey = mapOf("primaryKey" to AttributeValue.S("lastKeyPk"))
val result = client.scanPaginated {
    tableName = table
    exclusiveStartKey = lastKey
}
val simpleScanResult = client.scan {
    tableName = table
    exclusiveStartKey = lastKey
}
//results should have the same item at pos 0, but they don't as exclusiveStartKey is ignored.
```

### AWS SDK for Kotlin version

2

### Platform (JVM/JS/Native)

JVM

### Operating system and version

Mac OS 14

0marperez:
Hello @thake, the fix for this issue was released. I'll mark this issue as resolved but if you're still having issues feel free to @ me.

0marperez:
Fixed by https://github.com/smithy-lang/smithy-kotlin/pull/1101

PR #1332, opened 2024-06-14
misc: dynamodb e2e test

0marperez:
## Issue \#

N/A

## Description of changes

Added e2e tests to make sure we don't regress on this behavior: #1330 

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -9,8 +9,8 @@ coroutines-version = "1.7.3"
 atomicfu-version = "0.23.1"
 
 # smithy-kotlin codegen and runtime are versioned separately
-smithy-kotlin-runtime-version = "1.2.8"
-smithy-kotlin-codegen-version = "0.32.8"
+smithy-kotlin-runtime-version = "1.2.9"
+smithy-kotlin-codegen-version = "0.32.9"
 
 # codegen
 smithy-version = "1.49.0"


0marperez:
CI is being weird and I'll need to merge `smithy-kotlin` [changes](https://github.com/smithy-lang/smithy-kotlin/pull/1101) first before we see e2e tests pass in CI (they pass locally)

services/dynamodb/e2eTest/src/DynamoDbTestUtils.kt
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.sdk.kotlin.services.dynamodb
+
+/**
+ * Checks if a Dynamo DB table exists based on its name
+ */
+internal suspend fun String.exists(client: DynamoDbClient): Boolean =
+    client.listTables {}.tableNames?.contains(this) ?: false
```
lauzadis, line 11: nit/style: I think a better function signature would be `DynamoDbClient.tableExists(name: String): Boolean`.

`String.exists` is a pretty generic name and might be misused in the future. Though this function is `internal` to the dynamodb e2e tests, so it's not likely

services/dynamodb/e2eTest/src/PaginatorTest.kt
```diff
@@ -0,0 +1,114 @@
...
+import org.junit.jupiter.api.BeforeAll
+import org.junit.jupiter.api.TestInstance
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.time.Duration.Companion.seconds
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class PaginatorTest {
+    private val client = DynamoDbClient { region = "us-west-2" }
+    private val table = "testTable"
```
lauzadis, line 22: suggestion: We should probably add a random suffix to this table name so we don't end up in a bad state if DynamoDb ever has an error deleting the table during cleanup

services/dynamodb/e2eTest/src/PaginatorTest.kt
```diff
@@ -0,0 +1,114 @@
...
+
+        client.putItem {
+            tableName = table
+            item = mapOf(
+                "Artist" to AttributeValue.S("Foo"),
+                "SongTitle" to AttributeValue.S("Qux"),
+            )
+        }
+
+        client.scanPaginated {
```
lauzadis, line 100: suggestion: We should add `limit=1` to ensure pagination works (fetch 2 pages instead of just 1)

lauzadis (approved):

PR #1335, opened 2024-06-18
feat: support `implicitGlobalRegion`

lauzadis:
Adds a new partitions property `implicitGlobalRegion`

note: MRAP e2e tests had to be disabled for this to pass CI. Check out internal issue SDK-KT-214 for more context

## Description of changes

### DIFF
--- a/services/s3/e2eTest/src/S3TestUtils.kt
+++ b/services/s3/e2eTest/src/S3TestUtils.kt
@@ -14,8 +14,6 @@ import aws.sdk.kotlin.services.s3.model.LifecycleRuleFilter
 import aws.sdk.kotlin.services.s3.paginators.listObjectsV2Paginated
 import aws.sdk.kotlin.services.s3.waiters.waitUntilBucketExists
 import aws.sdk.kotlin.services.s3.waiters.waitUntilBucketNotExists
-import aws.sdk.kotlin.services.s3control.*
-import aws.sdk.kotlin.services.s3control.model.*
 import aws.smithy.kotlin.runtime.http.request.HttpRequest
 import aws.smithy.kotlin.runtime.text.ensurePrefix
 import kotlinx.coroutines.*


aws-runtime/aws-endpoint/common/src/aws/sdk/kotlin/runtime/endpoint/functions/Functions.kt
```diff
@@ -61,12 +62,13 @@ public data class Partition(
...
 @InternalSdkApi
-public data class PartitionConfig(
+public data class PartitionConfig @JvmOverloads constructor(
     public val name: String? = null,
     public val dnsSuffix: String? = null,
     public val dualStackDnsSuffix: String? = null,
     public val supportsFIPS: Boolean? = null,
     public val supportsDualStack: Boolean? = null,
+    public val implicitGlobalRegion: String? = null,
 ) {
```
0marperez, line 72: **correctness**: I think if you also add a 2nd constructor it might save us from having to break the API
```kotlin
public data class PartitionConfig (
    public val name: String? = null,
    public val dnsSuffix: String? = null,
    public val dualStackDnsSuffix: String? = null,
    public val supportsFIPS: Boolean? = null,
    public val supportsDualStack: Boolean? = null,
    public val implicitGlobalRegion: String? = null,
) {
    @JvmOverloads
    public constructor (
        name: String? = null,
        dnsSuffix: String? = null,
        dualStackDnsSuffix: String? = null,
        supportsFIPS: Boolean? = null,
        supportsDualStack: Boolean? = null,
    ) : this(
        name,
        dnsSuffix,
        dualStackDnsSuffix,
        supportsFIPS,
        supportsDualStack,
        null,
    )

```

0marperez (changes requested):

aws-runtime/aws-endpoint/common/src/aws/sdk/kotlin/runtime/endpoint/functions/Functions.kt
lauzadis: Which specific API break are you concerned about? I added `@JvmOverloads` so there are new constructors generated for each parameter. I think that nullifies the API "break" reported by the validator

lauzadis: I don't really want to just add a second constructor because:
1. This is `@InternalSdkApi`
2. This will continue being an issue as long as we use data classes. But I think `@JvmOverloads` help with this. I made an internal issue SDK-KT-215 to research adding this to all our data classes

0marperez: Oh okay, nvm then. I was worried about adding unnecessary ones and the ones from line 52 & 53. You're also right that it's internal anyway

0marperez (approved):

PR #1344, opened 2024-07-01
fix: Apply accessibility violation fixes

lauzadis:
Fixes various docs accessibility issues:
1. Ensure pages reflow without requiring two-dimensional scrolling without loss of content or functionality
	- We were violating this with via inline code blocks. Each symbol in these blocks are rendered separately, so it's not a simple fix to just enable wrapping on these elements.
	- This was fixed by adding a hamburger button to all code blocks which controls visibility. The code blocks are automatically hidden on small screens (<550px)
	- "Moving content to an accessible show/hide control, such as a hamburger menu, is acceptable."
1. Ensure custom controls provide proper textual name, role, and state information
    - We were violating this by not setting `aria-label`, `aria-controls`, `aria-expanded` attributes on the left sidebar arrow button. The fix was to add these attributes to each `navButton`. 
1. Ensure all interactive functionality is operable with the keyboard
    - Users could not expand the left sidebar entries using **only** the keyboard. This was fixed by making the navButton a separate togglable button.
1. Provide a mechanism for skipping past repetitive content
    - Users could not skip past the left sidebar, they would have to tab through each entry before reaching the main content of the page. This was fixed by adding an invisible button immediately after the first left sidebar entry. When tabbed into, the button becomes visible and provides a link to the main content of the page.

## Issue \#

N/A

## Description of changes

### DIFF
--- a/docs/dokka-presets/scripts/accessibility.js
+++ b/docs/dokka-presets/scripts/accessibility.js
@@ -1,7 +1,7 @@
 /**
  * Apply "skip to main content" buttons after each active left sidebar `sideMenuPart`.
  * These are invisible and only accessible via keyboard
- * Fix for accessibility violation: "Provide a mechanism for skipping past repetitive content"
+ * Fixes accessibility violation: "Provide a mechanism for skipping past repetitive content"
  */
 function applySkipLinks() {
     function insertSkipLink(element) {
@@ -50,7 +50,7 @@ if (document.readyState === "interactive" || document.readyState === "complete" 
 
 /**
  * Ensure `navButton` elements are interactable and have proper accessibility properties
- * Fix for accessibilty violation: "Ensure all interactive functionality is operable with the keyboard"
+ * Fixes accessibilty violation: "Ensure all interactive functionality is operable with the keyboard"
  */
 function ensureNavButtonInteractable() {
     const navButtons = document.querySelectorAll('.navButton');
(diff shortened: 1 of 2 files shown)


docs/dokka-presets/scripts/accessibility.js
```diff
@@ -0,0 +1,144 @@
...
+        const anchor = document.createElement('a');
+        anchor.classList.add('skip-to-content');
+        anchor.href = '#content';
+        anchor.innerHTML = 'Skip to Main Content';
+        anchor.setAttribute("tabindex", "0");
+        skipLink.appendChild(anchor);
+        if (element.children.length > 1) {
+            element.insertBefore(skipLink, element.children[1]);
+        } else {
+            element.appendChild(skipLink);
```
0marperez, line 21: question: do we have elements with 0 children?

docs/dokka-presets/scripts/accessibility.js
```diff
@@ -0,0 +1,144 @@
...
+            } else {
+                element.style.display = 'block';
+                toggleContent.setAttribute('aria-expanded', 'true');
+            }
+        });
+    });
+}
+
+document.addEventListener('DOMContentLoaded', ensureContentReflow)
+if (document.readyState === "interactive" || document.readyState === "complete" ) { applySkipLinks() }
```
0marperez, line 144: question: is this supposed to be here twice? It's also in line 48

docs/dokka-presets/scripts/accessibility.js
```diff
@@ -0,0 +1,144 @@
...
+ * Fix for accessibility violation: "Ensure pages reflow without requiring two-dimensional scrolling without loss of content or functionality"
+ */
+function ensureContentReflow() {
+    const MIN_WINDOW_SIZE = 550
+
+    // Function to insert 'toggle content' button
+    function insertToggleContentButton(element) {
+        const toggleContent = document.createElement('button');
+        toggleContent.className = 'aws-toggle-content-btn';
+        toggleContent.textContent = '☰';
```
0marperez, line 105: suggestion: we could use `▶` and `▼` elements instead of `☰`

0marperez (approved):

docs/dokka-presets/scripts/accessibility.js
lauzadis: No, it should be `ensureContentReflow`, thanks

docs/dokka-presets/scripts/accessibility.js
lauzadis: Sure, I made the change and it looks good to me

docs/dokka-presets/scripts/accessibility.js
lauzadis: No, the leaf nodes in our documentation have 1 child which is what this condition matches. 

The structure for this condition is like this:
```html
<div sideMenuPart>
	<div overview>
	<div skip-to-main-content> -- added via appendChild
```

The other condition is for structures like like this:
```html
<div sideMenuPart>
	<div overview>
	<div skip-to-main-content> -- added via insertBefore element.children[1]
	<div child sideMenuPart>
		<div overview>
	<div child sideMenuPart>
		<div overview>
	<div child sideMenuPart>
		<div overview>
```

ianbotsf (approved):

PR #1348, opened 2024-07-05
chore: enable Dependabot automatic version upgrade PRs

ianbotsf:
## Issue \#

(none)

## Description of changes

This change should enable Dependabot to send PRs to us when new versions of Smithy are launched. Only Smithy is allowlisted for now—we can see if it makes sense to enable for other dependencies in the future.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+# for information about dependabot.yml format/options
+
+version: 2
+
+updates:
+  - package-ecosystem: gradle
+    directory: "/"
+    schedule:
+      interval: daily
+    allow:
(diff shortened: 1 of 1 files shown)


0marperez (approved):

lauzadis (approved):

ianbotsf:
As noted by @lauzadis this change would stop us from receiving automatic security version bump PRs.

From [documentation of the `allow` parameter](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow):
> Use the `allow` option to customize which dependencies are updated. This applies to both version and security updates.

This leaves us with no way to define separate rules for security updates and non-security version upgrades. Since automated security updates are more important, I'm closing this PR for now until such time as feature requests like https://github.com/dependabot/dependabot-core/issues/6380 or https://github.com/dependabot/dependabot-core/issues/1778 are addressed.

PR #1351, opened 2024-07-08
misc: schedule the update-release-branch workflow

lauzadis:
## Description of changes

### DIFF
--- a/.github/workflows/update-release-branch.yml
+++ b/.github/workflows/update-release-branch.yml
@@ -21,7 +21,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  dry_run: ${{ github.event.inputs.dry_run || 'false' }
+  DRY_RUN: ${{ github.event.inputs.dry_run || 'false' }
 
 jobs:
   update-release:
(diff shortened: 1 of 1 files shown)


0marperez (approved):

.github/workflows/update-release-branch.yml
```diff
@@ -13,6 +13,8 @@ on:
         required: true
         type: boolean
         default: true
+  schedule:
+    - cron: "0 7 * * 1-5"  # At 07:00 UTC (00:00 PST, 03:00 EST), Monday through Friday
```
ianbotsf, line 17: **Question**: What parameters will this scheduled invocation run with? I see `description` and `dry_run` defined and set above but nothing equivalent here...

ianbotsf (approved):

.github/workflows/update-release-branch.yml
0marperez: If it runs with defaults we could make dry run default to false instead of true

lauzadis: Good point, it seems like the inputs will be null when a scheduled action is run.
https://github.com/orgs/community/discussions/74698

I'll update this to set the correct input values

Issue #314, opened 2021-09-08
Codegen from integrations generates misleading header comment


kggilmer:
## Describe the bug

The warning at the top of codegen files specify that all files were generated by smithy-kotlin-codegen.  However this is sometimes not the case, for instances when aws-kotlin-sdk, embedded integrations, and possibly in the future external integrations included at codegen time.  

## Expected Behavior
The codegen header should be accurate in all cases.

## Current Behavior
The codegen header specifies that the smithy-kotlin-codegen module generated the file.

## Possible Solution

[This logic](https://github.com/awslabs/smithy-kotlin/blob/main/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/core/KotlinWriter.kt#L238) should be overridable in some way so that it can be customized via integrations.  Possibly automatically.

PR #1357, opened 2024-07-10
initial implementation of codegen for low-level operations/types

ianbotsf:
## Issue \#

Part of https://github.com/awslabs/aws-sdk-kotlin/issues/76

## Description of changes

At long last, the wait for codegenned mapper operations and high-level types is ~done~ started! This change **removes** the handwritten toy implementation of a `getItem` operation and now generates the operation glue code for the following DDB ops:
* `deleteItem`
* `getItem`
* `putItem`
* `query`*
* `scan`*

\* These operations should be paginated but the codegen doesn't handle that yet. It's left as a `FIXME` in the code for follow-up.

### PR flight plan

⚠️ There's a lot to take in here, especially if you're unfamiliar with [KSP](https://kotlinlang.org/docs/ksp-overview.html). I recommend taking a short break to learn some basics about KSP before getting started with the review.

Recommended order of review:
* Start with the new **ddb-mapper-ops-codegen** module, in particular:
  * `HighLevelOpsProcessor` which is the main entry point for KSP. Everything else in this module is called somehow from this point.
  * The `model` package is for data types for describing the low- & high-level operations, structures, and types
  * The `core` package contains rudimentary code writing utilities such as indentation, blocks, imports, and template processing. These are good candidates for commonizing somewhere in the future, since we'll likely need similar functionality for DDB mapper annotation processing and possibly other HLLs in the future.
  * The `rendering` package contains structured codegen for rendering the operations, convenience functions, and types
* Next, hit up the **dynamodb-mapper** module, in particular:
  * **build.gradle.kts** has been beefed up with the new KSP application of **ddb-mapper-ops-codegen**. Some unpleasant (and hopefully _temporary_) hacks were necessary to get KSP to work in both JVM+Native builds _and_ JVM-only builds.
  * The new unit tests in the `operations` package which use [DynamoDB Local](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html) in lieu of mocking or making actual calls to DDB

### Sample codegen output

Understanding the codegen may be easier if you see the generated code. All code is generated into the **hll/ddb-mapper/dynamodb-mapper/build/generated/ksp/common/commonMain/kotlin** directory.

<details>
<summary>GetItem operation</summary>

```kotlin
// Code generated by ddb-mapper-ops-codegen. DO NOT EDIT!

package aws.sdk.kotlin.hll.dynamodbmapper.operations

import aws.sdk.kotlin.hll.dynamodbmapper.TableSpec
import aws.sdk.kotlin.hll.dynamodbmapper.items.ItemSchema
import aws.sdk.kotlin.hll.dynamodbmapper.model.toItem
import aws.sdk.kotlin.hll.dynamodbmapper.pipeline.internal.HReqContextImpl
import aws.sdk.kotlin.hll.dynamodbmapper.pipeline.internal.MapperContextImpl
import aws.sdk.kotlin.hll.dynamodbmapper.pipeline.internal.Operation
import aws.sdk.kotlin.services.dynamodb.model.ConsumedCapacity
import aws.sdk.kotlin.services.dynamodb.model.ReturnConsumedCapacity
import kotlin.Boolean
import kotlin.String
import aws.sdk.kotlin.services.dynamodb.model.GetItemRequest as LowLevelGetItemRequest
import aws.sdk.kotlin.services.dynamodb.model.GetItemResponse as LowLevelGetItemResponse

public interface GetItemRequest<T> {
    public companion object { }

    public val consistentRead: Boolean?
    public val key: T?
    public val returnConsumedCapacity: ReturnConsumedCapacity?
}

private data class GetItemRequestImpl<T>(
    override val consistentRead: Boolean?,
    override val key: T?,
    override val returnConsumedCapacity: ReturnConsumedCapacity?,
): GetItemRequest<T>

public fun <T> GetItemRequest(
    consistentRead: Boolean?,
    key: T?,
    returnConsumedCapacity: ReturnConsumedCapacity?,
): GetItemRequest<T> = GetItemRequestImpl(
    consistentRead,
    key,
    returnConsumedCapacity,
)

private fun <T> GetItemRequest<T>.convert(
    tableName: String?, 
    schema: ItemSchema<T>,
) = LowLevelGetItemRequest {
    consistentRead = this@convert.consistentRead
    returnConsumedCapacity = this@convert.returnConsumedCapacity
    this@convert.key?.let { key = schema.converter.toItem(it, schema.keyAttributeNames) }
    this.tableName = tableName
}

public interface GetItemResponse<T> {
    public companion object { }

    public val consumedCapacity: ConsumedCapacity?
    public val item: T?
}

private data class GetItemResponseImpl<T>(
    override val consumedCapacity: ConsumedCapacity?,
    override val item: T?,
): GetItemResponse<T>

public fun <T> GetItemResponse(
    consumedCapacity: ConsumedCapacity?,
    item: T?,
): GetItemResponse<T> = GetItemResponseImpl(
    consumedCapacity,
    item,
)

private fun <T> LowLevelGetItemResponse.convert(schema: ItemSchema<T>) = GetItemResponse<T>(
    consumedCapacity = this@convert.consumedCapacity,
    item = this@convert.item?.toItem()?.let(schema.converter::fromItem),
)

internal fun <T> getItemOperation(table: TableSpec<T>) = Operation(
    initialize = { hReq: GetItemRequest<T> -> HReqContextImpl(hReq, table.schema, MapperContextImpl(table, "GetItem")) },
    serialize = { hReq, schema -> hReq.convert(table.name, schema) },
    lowLevelInvoke = table.mapper.client::getItem,
    deserialize = LowLevelGetItemResponse::convert,
    interceptors = table.mapper.config.interceptors,
)
```
</details>

<details>
<summary>Table operations interface/implementation</summary>

```kotlin
// Code generated by ddb-mapper-ops-codegen. DO NOT EDIT!

package aws.sdk.kotlin.hll.dynamodbmapper.operations

import aws.sdk.kotlin.hll.dynamodbmapper.TableSpec

/**
 * Provides access to operations on a particular table, which will invoke low-level operations after
 * mapping objects to items and vice versa
 * @param T The type of objects which will be read from and/or written to this table
 */
public interface TableOperations<T> {
    public suspend fun deleteItem(request: DeleteItemRequest<T>): DeleteItemResponse<T>
    public suspend fun getItem(request: GetItemRequest<T>): GetItemResponse<T>
    public suspend fun putItem(request: PutItemRequest<T>): PutItemResponse<T>
    public suspend fun query(request: QueryRequest<T>): QueryResponse<T>
    public suspend fun scan(request: ScanRequest<T>): ScanResponse<T>
}

internal class TableOperationsImpl<T>(private val tableSpec: TableSpec<T>) : TableOperations<T> {
    override suspend fun deleteItem(request: DeleteItemRequest<T>) =
        deleteItemOperation(tableSpec).execute(request)
    
    override suspend fun getItem(request: GetItemRequest<T>) =
        getItemOperation(tableSpec).execute(request)
    
    override suspend fun putItem(request: PutItemRequest<T>) =
        putItemOperation(tableSpec).execute(request)
    
    override suspend fun query(request: QueryRequest<T>) =
        queryOperation(tableSpec).execute(request)
    
    override suspend fun scan(request: ScanRequest<T>) =
        scanOperation(tableSpec).execute(request)
    
}
```
</details>

### Known work remaining

Several `TODO`s and `FIXME`s are to be found right now. At the very least I know we'll need to:
* Replace function-style builders (e.g., for request/response creation) with DSL-style builders (similar to what we use in the low-level clients)
* Render DSL-style extension methods for operations (similar to what we do for low-level operations)
* Give KSP clearer indications of which output files depend on which input files. Right now, KSP can't detect which updates should trigger re-generating which code so it generates _all the code_ on _every build_.
* Generate paginated operations (e.g., `Query`, `Scan`, etc.) correctly

### DIFF
--- a/hll/build.gradle.kts
+++ b/hll/build.gradle.kts
@@ -24,7 +24,6 @@ val sdkVersion: String by project
 val libraries = libs
 
 subprojects {
-    println("Subproject $this needsKmpConfigured? $needsKmpConfigured")
     if (!needsKmpConfigured) return@subprojects
 
     group = "aws.sdk.kotlin"


hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/internal/TableImpl.kt
```diff
@@ -6,35 +6,48 @@ package aws.sdk.kotlin.hll.dynamodbmapper.internal
...
+    name: String,
+    schema: ItemSchema.CompositeKey<T, PK, SK>,
+): Table.CompositeKey<T, PK, SK> {
+    val specImpl = TableSpecCompositeKeyImpl(mapper, name, schema)
+    val opsImpl = TableOperationsImpl(specImpl)
+    return object :
+        Table.CompositeKey<T, PK, SK>,
+        TableSpec.CompositeKey<T, PK, SK> by specImpl,
+        TableOperations<T> by opsImpl {
+        override suspend fun getItem(partitionKey: PK, sortKey: SK) = TODO("Not yet implemented")
```
lauzadis, line 51: question: should these `override suspend fun getItem` be removed now that operations are codegenerated?

hll/ddb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/testutils/DdbLocalTest.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    private val ddbHolder = lazy {
+        DynamoDbClient {
+            endpointUrl = Url {
+                scheme = Scheme.HTTP
+                host = Host.Domain("localhost")
+                port = DDB_LOCAL_PORT
+            }
+
+            region = "us-west-2" // FIXME
```
lauzadis, line 42: question: What is the FIXME here? It seems like DDB Local requires a region:

> The AWS SDKs for DynamoDB require that your application configuration specify an access key value and an AWS Region value. Unless you're using the -sharedDb or the -inMemory option, DynamoDB uses these values to name the local database file. These values don't have to be valid AWS values to run locally. However, you might find it convenient to use valid values so that you can run your code in the cloud later by changing the endpoint you're using.
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.UsageNotes.html 

Maybe we can update it to `DUMMY` like the other unnecessary parameters

hll/ddb-mapper/dynamodb-mapper/jvm/test/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/OperationTest.kt
```diff
@@ -23,6 +23,8 @@ import kotlin.test.assertIs
 
 private const val TABLE_NAME = "foo-table"
 
+// FIXME Should be in commonTest but mockk is JVM-only and finding a good KMP mocking library is hard
```
lauzadis, line 26: Is it possible/useful to rewrite these tests and remove the mockk dependency? Do we need to use a mock to test this functionality?

hll/ddb-mapper/ddb-mapper-ops-codegen/build.gradle.kts
lauzadis, line 1: Is there a reason for the naming inconsistencies between modules like `ddb-mapper-*` and `dynamodb-mapper-*`?

hll/ddb-mapper/ddb-mapper-ops-codegen/jvm/src/aws/sdk/kotlin/hll/dynamodbmapper/codegen/core/CodeGenerator.kt
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    override fun openBlock(template: String, vararg args: Any) {
+        write(template, *args)
+        indent()
+    }
+
+    override fun persist() {
+        val content = buildString {
+            appendLine("// Code generated by ddb-mapper-ops-codegen. DO NOT EDIT!")
```
lauzadis, line 65: For a second I thought you fixed https://github.com/awslabs/aws-sdk-kotlin/issues/314 until I realized this was hand-written 😃

lauzadis (approved):

hll/ddb-mapper/ddb-mapper-ops-codegen/build.gradle.kts
ianbotsf: There's a reason, yes, but I'm not sure if it's a good one or not: modules named `dynamodb-*` are intended to be released publicly as Maven artifacts. Modules named `ddb-*` are intended to remain as build details and should not be released as Maven artifacts. It's a quick way to see/remember which modules our users will interact with vs not.

How does that sound to you? Should we make the naming more consistent?

hll/ddb-mapper/ddb-mapper-ops-codegen/jvm/src/aws/sdk/kotlin/hll/dynamodbmapper/codegen/core/CodeGenerator.kt
ianbotsf: Yep, totally separate codegen I'm afraid. #314 still stands.

hll/ddb-mapper/ddb-mapper-ops-codegen/jvm/src/aws/sdk/kotlin/hll/dynamodbmapper/codegen/model/MemberCodegenBehavior.kt
```diff
@@ -0,0 +1,66 @@
...
+            member.isTableName -> Hoist
+            else -> PassThrough
+        }.also { println("  ${member.name} is $it") }
+    }
+
+    data object PassThrough : MemberCodegenBehavior
+    data object MapAll : MemberCodegenBehavior
+    data object MapKeys : MemberCodegenBehavior
+    data object Drop : MemberCodegenBehavior
+    data object Hoist : MemberCodegenBehavior // FIXME Note sure this is useful...get rid of Hoist?
```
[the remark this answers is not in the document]
ianbotsf: `Hoist` indicates that the field shouldn't appear in the request/response structure itself but should be "hoisted" to a different location in the overall API. In the case of `tableName`, that's required before you can even invoke an operation:

```kotlin
val table = mapper.getTable("the-table-name", ...)
table.getItem(...)
```

As the `FIXME` notes, I'm not entirely sure this is a meaningful codegen behavior. Right now the effect is the same as `Drop` in that it's not included in the generated high-level structures. The [presence of `tableName` in other APIs](https://github.com/awslabs/aws-sdk-kotlin/blob/5f9d679434d8bbf0e514b8804916772801810def/hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/DynamoDbMapper.kt#L37-L60) is a result of hand-written code, not codegen. I've left it here because it _may_ be useful for non-table-centric operations such as [`BatchWriteItem`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html), in which multiple table names may be specified.

hll/ddb-mapper/dynamodb-mapper/build.gradle.kts
```diff
@@ -3,19 +3,139 @@
...
+        }
+    }
+} else {
+    // FIXME This is a dirty hack for JVM-only builds which KSP doesn't consider to be "multiplatform". Explanation of
+    //  hack follows in narrative, minimally-opinionated comments.
+
+    // Start by invoking the JVM-only KSP configuration
+    dependencies.kspJvm(project(":hll:ddb-mapper:ddb-mapper-ops-codegen"))
+
+    // Then we need to move the generated source from jvm to common. Gradle lacks a move task so that means a copy...
```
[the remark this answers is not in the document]
ianbotsf: 🤦‍♂️ No, that's precisely what I was looking for and failed to find. I'll try it out and see.

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/internal/TableImpl.kt
ianbotsf: No, they should stay here and get implemented because they're convenience overloads of the codegenned method. This will allow users to do something simple like `table.getItem(123)` instead of `table.getItem { key = SomeItem(id = 123) }`. We may add other hand-written convenience overloads for codegenned methods but I knew for sure at least this one would be useful.

hll/ddb-mapper/dynamodb-mapper/jvm/test/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/OperationTest.kt
ianbotsf: Technically we never _need_ a mocking framework to test functionality, they're merely convenient for stubbing out mock objects quickly and with minimal boilerplate. In this case, I'm using **mockk** to mock `DynamoDbMapper` (which will eventually have dozens of codegenned operations on it) and to [spy](https://mockk.io/#spy) on `Interceptor` (which _also_ has many methods) and verify invocation order. All that is achievable with hand-written mocks but would increase the size of the test code.

This test _shouldn't_ be platform-specific but also the _behavior_ isn't platform-specific either—it's all common code. Thus, the risk of it working differently on non-JVM platforms is pretty low. All things being equal, I'd definitely prefer it in **commonTest** but since the tradeoff is more boilerplate for marginal test utility, I think I prefer leaving it here until we can find a good KMP mocking library.

What do you think?

hll/ddb-mapper/ddb-mapper-ops-codegen/jvm/src/aws/sdk/kotlin/hll/dynamodbmapper/codegen/model/MemberCodegenBehavior.kt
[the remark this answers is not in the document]
ianbotsf: I suppose [the YAGNI principle](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) dictates that this code shouldn't exist _until_ there's a clear planned/actual use for it so perhaps I should just clean it up for now. 🤔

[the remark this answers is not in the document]
ianbotsf: Wait no, I did actually make use of it! 😅 It's used when rendering the `convert` operation that turns low-level requests into high-level requests:

```kotlin
// generated code
private fun <T> GetItemRequest<T>.convert(
    tableName: String?,
    schema: ItemSchema<T>,
) = LowLevelGetItemRequest {
    consistentRead = this@convert.consistentRead
    returnConsumedCapacity = this@convert.returnConsumedCapacity
    this@convert.key?.let { key = schema.converter.toItem(it, schema.keyAttributeNames) }
    this.tableName = tableName
}
```

In that generated code the name/type `tableName: String` are derived from the low-level structure, not hand-written. So I believe I'll keep it as-is for now and remove the `FIXME`.

hll/ddb-mapper/dynamodb-mapper/jvm/test/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/OperationTest.kt
lauzadis: I think it's fine to keep it JVM-only since the code is all common. We have other tests set up like this IIRC.

hll/ddb-mapper/ddb-mapper-ops-codegen/build.gradle.kts
lauzadis: Oh ok. I'm open to either way but I think keeping them consistent would be better. I'm guessing we'll have a 
clear mechanism for deciding which modules should be published (`ignoreProjects = listOf(...)`) or something like that?

hll/ddb-mapper/ddb-mapper-ops-codegen/jvm/src/aws/sdk/kotlin/hll/dynamodbmapper/codegen/rendering/OperationRenderer.kt
```diff
@@ -0,0 +1,113 @@
...
+        blankLine()
+        renderOperationFactory()
+    }
+
+    private fun renderOperationFactory() {
+        val factoryName = factoryFunctionName(operation)
+
+        withBlock("internal fun <T> #L(table: #T) = #T(", ")", factoryName, Types.tableSpec("T"), Types.Operation) {
+            write(
+                "initialize = { hReq: #T -> #T(hReq, table.schema, #T(table, #S)) },",
```
0marperez, line 36: style: `highLevelReq` over `hReq`/`hlReq` for readability

0marperez (approved):
style: k docs on some of the classes would be helpful

hll/build.gradle.kts
```diff
@@ -24,6 +24,7 @@ val sdkVersion: String by project
 val libraries = libs
 
 subprojects {
+    println("Subproject $this needsKmpConfigured? $needsKmpConfigured")
```
lauzadis, line 27: nit: stray println?

lauzadis (approved):
Nice, so many docs!

hll/build.gradle.kts
ianbotsf: Yep, should've removed that.

PR #1375, opened 2024-07-30
initial implementation of secondary index support

ianbotsf:
## Issue \#

(none)

## Description of changes

This change adds support for secondary indices by reworking some of the primitives for table representation in the runtime and slightly tweaking the codegen. Notable changes:
* `Table<T>` now derives from new parent interface `Queryable<T>`, which is also the parent of new interface `Index<T>` 
* Indices have their own schemas because not all base table attributes are necessarily projected into indices
* Moved `Table` into the **model** package alongside things like `Item`. It probably always belonged here but I _especially_ wanted it moved rather than create the new `Index` and `Queryable` interfaces in the **dynamodbmapper** base package.

### Things I don't love about this PR and for which I'm happy to receive suggestions

* The name `Queryable` doesn't sound quite right. It's technically true that tables and indices are "queryable" but they're also "scannable".
* The `PersistenceSpec` interface has an `indexName: String?` field but its inherited by both `Table` (which should never have an `indexName`) and `Index` (which should always have an `indexName`)
  * I also tried making individual `TableSpec` and `IndexSpec` interfaces but then operation factory methods need to be able to accept both (at least for `Query` and `Scan`). That complicates codegen but maybe it's worth it?—IDK, need more opinions.
* I had to introduce new unit tests for `Query` but neither `Query` or `Scan` are really _usable_ yet. They don't support pagination, object mapping, or condition expressions. So the unit test code which exercises the new index functionality is kinda gross. Maybe that's unavoidable at this point but now I wish I'd worked longer on getting `Query`/`Scan` ready before tackling indices.

### DIFF
--- a/hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/model/PersistenceSpec.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/model/PersistenceSpec.kt
@@ -10,19 +10,9 @@ import aws.sdk.kotlin.hll.dynamodbmapper.items.ItemSchema
 /**
  * Specifies how items can be read from and written to a specific DynamoDB location (such as a table or a secondary
  * index)
- * @param T The type of objects which will be read from and/or written to this table/index
+ * @param T The type of objects which will be read from and/or written to this item source
  */
 public interface PersistenceSpec<T> {
-    /**
-     * The name of the table
-     */
-    public val tableName: String
-
-    /**
-     * The name of the secondary index, if any
-     */
-    public val indexName: String?
-
     /**
      * The [DynamoDbMapper] which holds the underlying DynamoDB service client used to invoke operations
      */
--- a/hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/model/ItemSourceKind.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/model/ItemSourceKind.kt
@@ -4,24 +4,43 @@
  */
 package aws.sdk.kotlin.hll.dynamodbmapper.codegen.model
 
+import aws.sdk.kotlin.hll.dynamodbmapper.codegen.util.Pkg
+
 /**
  * Identifies a type in the `ItemSource<T>` hierarchy
+ * @param hoistedFields Which fields should be hoisted from the low-level request type for this item source kind (e.g.,
+ * for `TableSpec<T>` the `tableName` field should be hoisted)
+ * @param parent The parent type of this type (if any)
+ * @param isAbstract Indicates whether this item source kind is purely abstract and should not have an implementation
+ * class (e.g., `ItemSource<T>` should be abstract and non-instantiable)
  */
-enum class ItemSourceKind(val parent: ItemSourceKind? = null) {
+enum class ItemSourceKind(
+    val hoistedFields: List<String>,
+    val parent: ItemSourceKind? = null,
+    val isAbstract: Boolean = false,
+) {
     /**
      * Indicates the `ItemSource<T>` interface
      */
-    ItemSource,
+    ItemSource(listOf(), isAbstract = true),
 
     /**
      * Indicates the `Index<T>` interface
      */
-    Index(ItemSource),
+    Index(listOf("indexName", "tableName"), ItemSource),
 
     /**
      * Indicates the `Table<T>` interface
      */
-    Table(ItemSource),
+    Table(listOf("tableName"), ItemSource),
+
+    ;
+
+    /**
+     * Get the [TypeRef] for the `*Spec` type for this item source kind
+     * @param typeVar The type variable name to use for the generic type
+     */
+    fun getSpecType(typeVar: String): TypeRef = TypeRef(Pkg.Hl.Model, "${name}Spec", listOf(TypeVar(typeVar)))
 }
 
 /**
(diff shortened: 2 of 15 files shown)


hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/rendering/HighLevelRenderer.kt
```diff
@@ -14,7 +16,17 @@ import aws.sdk.kotlin.hll.dynamodbmapper.codegen.model.Operation
 class HighLevelRenderer(private val ctx: RenderContext, private val operations: List<Operation>) {
     fun render() {
         operations.forEach(::render)
-        TableOperationsRenderer(ctx, operations).render()
+
+        val operationsRenderers = mutableMapOf<QueryableKind, QueryableOperationsRenderer>()
+        QueryableKind.entries.forEach { qk ->
+            ctx.warn("About to generate hll operations for $qk")
```
lauzadis, line 22: question: should this be `info` or `debug`?

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/model/PersistenceSpec.kt
```diff
@@ -0,0 +1,49 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.sdk.kotlin.hll.dynamodbmapper.model
+
+import aws.sdk.kotlin.hll.dynamodbmapper.DynamoDbMapper
+import aws.sdk.kotlin.hll.dynamodbmapper.items.ItemSchema
+
+/**
+ * Represents a table (optionally specialized by a secondary index) in DynamoDB and an associated item schema
```
lauzadis, line 11: nit: This doc comment implies a table can only have one secondary index, but they can have up to 20 by default, could be worth clarifying (either here or in `Table`, `Index`)

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/model/Queryable.kt
```diff
@@ -0,0 +1,34 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.sdk.kotlin.hll.dynamodbmapper.model
+
+import aws.sdk.kotlin.hll.dynamodbmapper.operations.QueryableOperations
+
+/**
+ * Represents a source of data which may be queried, such as a table or secondary index
+ */
+public interface Queryable<T> :
```
lauzadis, line 12: naming: would `DataSource<T>` be a better name? I agree `Queryable` is not perfect because it can be queried or scanned, "query" has two meanings there.

lauzadis (commented):
Would splitting `PersistenceSpec` into `TableSpec` and `IndexSpec` just require code-generating two operation factory methods, or are there more complicated changes that get introduced?

After looking at the PR I think splitting would be the right thing to do if it doesn't complicate codegen too much.

hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/rendering/HighLevelRenderer.kt
ianbotsf: Ugh that was debug code and doesn't belong in this PR. Removing!

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/model/PersistenceSpec.kt
ianbotsf: You're right. What I wanted to say was something like "this represents a thing that can be queried, such as a table or a secondary index on a table". Maybe I'll go with that if I don't wind up refactoring it out of existence.

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/model/Queryable.kt
```diff
@@ -0,0 +1,34 @@
...
+    PersistenceSpec<T>,
```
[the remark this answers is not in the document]
ianbotsf: Yes I think `PersistenceSpec`'s KDocs can be cleaned up a bit. The intent, though, is that `PersistenceSpec` and therefore `Queryable` represent a source of items which may be queried/scanned. One such source is a table directly. Another such source is a secondary index—but in DDB secondary indices are scoped to tables and so the table itself is still relevant. That's what I (clumsily) attempted to convey in `PersistenceSpec`: that it represents a table _or_ a secondary index (on a table). I'll take a second crack at these docs and see if I can make it clearer.

ianbotsf:
> Would splitting `PersistenceSpec` into `TableSpec` and `IndexSpec` just require code-generating two operation factory methods, or are there more complicated changes that get introduced?
> 
> After looking at the PR I think splitting would be the right thing to do if it doesn't complicate codegen too much.

You might be right and perhaps a fresh attempt is in order. Let me code it up locally and see if it's worth committing.

ianbotsf:
OK https://github.com/awslabs/aws-sdk-kotlin/pull/1375/commits/eb418acb50ca6d62790947ece79ce6ff10f016b7 addresses the simple stuff like renames and docs and https://github.com/awslabs/aws-sdk-kotlin/pull/1375/commits/1d2ca60971641741c835582310739804e1f4d989 addresses the split of `PersistenceSpec` into `TableSpec` and `IndexSpec`. It makes the codegen a bit more complex but cleans up the oddities like having an always-`null` `indexName` on a `Table`.

0marperez (approved):

lauzadis (approved):

Issue #1368, opened 2024-07-24
HeadObject 403 throws "Failed to parse response as 'restXml'"


joeeinertson-gravie:
### Describe the bug

When making a HeadObject call, if the correct permissions are not granted, resulting in a 403 Forbidden response from AWS, the SDK fails with a deserialization error.

### Expected behavior

SDK should throw an error with a relevant error message

### Current behavior

SDK throws `aws.sdk.kotlin.services.s3.model.S3Exception: Failed to parse response as 'restXml'`

Full stacktrace:
```
aws.sdk.kotlin.services.s3.model.S3Exception: Failed to parse response as 'restXml' error, Request ID: 7NQ2X7PRGZFBB940, Extended request ID: OtgUPnKRDAVgcebkFyDl2QkKDRLjJaNVfv76BpKhFuax71AMAOE2Q85HiXqco5BJv4f9scsU9yc=
	at aws.sdk.kotlin.services.s3.serde.HeadObjectOperationDeserializerKt.throwHeadObjectError(HeadObjectOperationDeserializer.kt:101) ~[task/:?]
	at aws.sdk.kotlin.services.s3.serde.HeadObjectOperationDeserializerKt.access$throwHeadObjectError(HeadObjectOperationDeserializer.kt:1) ~[task/:?]
	at aws.sdk.kotlin.services.s3.serde.HeadObjectOperationDeserializer.deserialize(HeadObjectOperationDeserializer.kt:33) ~[task/:?]
	at aws.sdk.kotlin.services.s3.serde.HeadObjectOperationDeserializer.deserialize(HeadObjectOperationDeserializer.kt:28) ~[task/:?]
	at aws.smithy.kotlin.runtime.http.operation.DeserializeHandler.call(SdkOperationExecution.kt:341) ~[task/:?]
	at aws.smithy.kotlin.runtime.http.operation.DeserializeHandler$call$1.invokeSuspend(SdkOperationExecution.kt) ~[task/:?]
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33) ~[task/:?]
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104) ~[task/:?]
	at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.common.kt:277) ~[task/:?]
[...]

PR #1378, opened 2024-08-05
fix: correctly handle all empty HeadObject/HeadBucket responses

lauzadis:
Previously we only checked for empty bodies with HTTP 404, but it seems like S3 also returns empty bodies for HTTP 403. This PR expands the handler logic to catch all empty bodies rather than just 404's.

## Issue \#

Closes https://github.com/awslabs/aws-sdk-kotlin/issues/1368

## Description of changes

### DIFF
--- a/services/s3/e2eTest/src/S3IntegrationTest.kt
+++ b/services/s3/e2eTest/src/S3IntegrationTest.kt
@@ -373,6 +373,21 @@ class S3BucketOpsIntegrationTest {
             assertContains(ex.message!!, "$expectedHost")
         }
     }
+
+    @Test
+    fun testHeadObjectForbidden(): Unit = runBlocking {
+        val ex = assertFailsWith<S3Exception> {
+            client.withConfig {
+                region = "us-east-1"
+            }.headObject {
+                bucket = "bucket"
+                key = "any-key.txt"
+            }
+        }
+
+        assertContains(ex.message, "Service returned error code 403: Forbidden")
+        assertEquals("403: Forbidden", ex.sdkErrorMetadata.errorCode!!)
+    }
 }
 
 // generate sequence of "chunks" where each range defines the inclusive start and end bytes


codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3OperationErrorHandler.kt
```diff
@@ -66,28 +63,29 @@ class S3OperationErrorHandler : KotlinIntegration {
...
-            .write("")
+
+            withInlineBlock("else {", "}") {
+                write("checkNotNull(payload) { #S }", "unable to parse error from empty response")
+                write("#T(payload)", parseS3ErrorResponse)
+            }
+        }
+
+        writer.withBlock("catch (ex: Exception) {", "}") {
+            withBlock("throw #T(#S).also {", "}", exceptionBaseSymbol, "Failed to parse response as ${ctx.protocol.name}") {
```
0marperez, line 85: style: I would vote for keeping the mention that this is an error in the exception message. It's kind of useful for determining where the exception happened

0marperez (commented):
question: do S3s 403 responses frequently contain payloads? I think we would still fail with the same `Failed to parse response as ...` exception if that's the case

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3OperationErrorHandler.kt
0marperez: `Failed to parse response as '${ctx.protocol.name}' error`

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3OperationErrorHandler.kt
```diff
@@ -66,28 +63,29 @@ class S3OperationErrorHandler : KotlinIntegration {
...
+
+        writer.withInlineBlock("val errorDetails = try {", "} ") {
+            // customize error matching to handle HeadObject/HeadBucket error responses which have no payload
+            withInlineBlock("if (payload == null) {", "} ") {
+                withInlineBlock("if (call.response.status == #T.NotFound) {", "} ", RuntimeTypes.Http.StatusCode) {
+                    write("#T(code = #S)", s3ErrorDetails, "NotFound")
+                }
+                withInlineBlock("else {", "}") {
+                    write("#T(code = #L)", s3ErrorDetails, "call.response.status.toString()")
                 }
```
ianbotsf, line 75: **Question**: Isn't `call.response.status.toString()` also correct if the status code is `NotFound`? Do we _need_ the special case for `NotFound` in this `if` block?

lauzadis:
> question: do S3s 403 responses frequently contain payloads? I think we would still fail with the same `Failed to parse response as ...` exception if that's the case

HeadObject 403 never contains a payload. Here is a sample codegen:
```kt
private fun throwHeadObjectError(context: ExecutionContext, call: HttpCall, payload: ByteArray?): kotlin.Nothing {
    val wrappedResponse = call.response.withPayload(payload)
    val wrappedCall = call.copy(response = wrappedResponse)

    val errorDetails = try {
        if (payload == null) {
            if (call.response.status == HttpStatusCode.NotFound) {
                S3ErrorDetails(code = "NotFound")

            } else {
                S3ErrorDetails(code = call.response.status.toString())

            }
        } else {
            checkNotNull(payload) { "unable to parse error from empty response" }
            parseS3ErrorResponse(payload)

        }
    } catch (ex: Exception) {
        throw S3Exception("Failed to parse response as restXml").also {
            setS3ErrorMetadata(it, wrappedCall.response, null)
        }
    }
    val ex = when(errorDetails.code) {
        "NotFound" -> NotFoundDeserializer().deserialize(context, wrappedCall, payload)
        else -> S3Exception(errorDetails.message)
    }

    setS3ErrorMetadata(ex, wrappedResponse, errorDetails)
    throw ex
}
```

I'm not sure how a 403 with a non-null payload will get an `Failed to parse response as restXml error` exception thrown unless it's malformed.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3OperationErrorHandler.kt
```diff
@@ -66,28 +63,29 @@ class S3OperationErrorHandler : KotlinIntegration {
...
             }
-            .write("")
+
+            withInlineBlock("else {", "}") {
+                write("checkNotNull(payload) { #S }", "unable to parse error from empty response")
+                write("#T(payload)", parseS3ErrorResponse)
+            }
```
ianbotsf, line 81: **Question**: Why does this `else` block still have a `checkNotNull(payload)` in it? The `if` condition was that `payload == null` so it should be impossible for `payload` to be `null` in this `else` block.

ianbotsf (changes requested):
**Correctness**: Add E2E test(s) please.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3OperationErrorHandler.kt
lauzadis: Yeah, it's probably not necessary, this was carried over from the previous codegen

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/S3OperationErrorHandler.kt
lauzadis: Actually it would be `Not Found` for a 404. We map it to "NotFound" which is how S3 models it.

ianbotsf: Does `code` have to match an S3 error code? Because `call.response.status.toString()` won't necessarily.

ianbotsf (approved):

0marperez (approved):

PR #1388, opened 2024-08-23
feat: smoke tests

0marperez:
## Description of changes

Adds gradle support for smoke tests

Companion PR: https://github.com/smithy-lang/smithy-kotlin/pull/1141

### DIFF
--- a/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/GradleGenerator.kt
+++ b/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/GradleGenerator.kt
@@ -4,8 +4,8 @@
  */
 package aws.sdk.kotlin.codegen
 
-import aws.sdk.kotlin.codegen.model.traits.FailedResponseTrait
-import aws.sdk.kotlin.codegen.model.traits.SuccessResponseTrait
+import aws.sdk.kotlin.codegen.model.traits.testing.TestFailedResponseTrait
+import aws.sdk.kotlin.codegen.model.traits.testing.TestSuccessResponseTrait
 import aws.sdk.kotlin.codegen.smoketests.smokeTestDenyList
 import software.amazon.smithy.kotlin.codegen.core.*
 import software.amazon.smithy.kotlin.codegen.integration.KotlinIntegration
@@ -119,8 +119,8 @@ class GradleGenerator : KotlinIntegration {
      * Generates a gradle task to run smoke tests
      */
     private fun generateSmokeTestTask(writer: GradleWriter, ctx: CodegenContext) {
-        val hasSuccessResponseTrait = ctx.model.expectShape<ServiceShape>(ctx.settings.service).hasTrait(SuccessResponseTrait.ID)
-        val hasFailedResponseTrait = ctx.model.expectShape<ServiceShape>(ctx.settings.service).hasTrait(FailedResponseTrait.ID)
+        val hasSuccessResponseTrait = ctx.model.expectShape<ServiceShape>(ctx.settings.service).hasTrait(TestSuccessResponseTrait.ID)
+        val hasFailedResponseTrait = ctx.model.expectShape<ServiceShape>(ctx.settings.service).hasTrait(TestFailedResponseTrait.ID)
         val inTestingEnvironment = hasFailedResponseTrait || hasSuccessResponseTrait
 
         /**
--- a/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/model/traits/testing/SmokeTestTraits.kt
+++ b/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/model/traits/testing/SmokeTestTraits.kt
@@ -1,4 +1,4 @@
-package aws.sdk.kotlin.codegen.model.traits
+package aws.sdk.kotlin.codegen.model.traits.testing
 
 import software.amazon.smithy.model.node.ObjectNode
 import software.amazon.smithy.model.shapes.ShapeId
@@ -6,8 +6,10 @@ import software.amazon.smithy.model.traits.AnnotationTrait
 
 /**
  * Indicates the annotated service should always return a failed response.
+ * IMPORTANT: This trait is intended for use in integration or E2E tests only, not in real-life smoke tests that run
+ * against a service endpoint.
  */
-class FailedResponseTrait(node: ObjectNode) : AnnotationTrait(ID, node) {
+class TestFailedResponseTrait(node: ObjectNode) : AnnotationTrait(ID, node) {
     companion object {
         val ID: ShapeId = ShapeId.from("smithy.kotlin.traits#failedResponseTrait")
     }
@@ -15,8 +17,10 @@ class FailedResponseTrait(node: ObjectNode) : AnnotationTrait(ID, node) {
 
 /**
  * Indicates the annotated service should always return a success response.
+ * IMPORTANT: This trait is intended for use in integration or E2E tests only, not in real-life smoke tests that run
+ * against a service endpoint.
  */
-class SuccessResponseTrait(node: ObjectNode) : AnnotationTrait(ID, node) {
+class TestSuccessResponseTrait(node: ObjectNode) : AnnotationTrait(ID, node) {
     companion object {
         val ID: ShapeId = ShapeId.from("smithy.kotlin.traits#successResponseTrait")
     }
--- a/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/smoketests/SmokeTestsCodegenRegionIntegration.kt
+++ b/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/smoketests/SmokeTestsCodegenRegionIntegration.kt
@@ -18,7 +18,7 @@ import software.amazon.smithy.smoketests.traits.SmokeTestsTrait
  */
 class SmokeTestsCodegenRegionIntegration : KotlinIntegration {
     override fun enabledForService(model: Model, settings: KotlinSettings): Boolean =
-        model.topDownOperations(settings.service).any { it.hasTrait<SmokeTestsTrait>() } && settings.sdkId !in smokeTestDenyList
+        model.topDownOperations(settings.service).any { it.hasTrait<SmokeTestsTrait>() }
 
     override val sectionWriters: List<SectionWriterBinding>
         get() = listOf(
@@ -29,8 +29,8 @@ class SmokeTestsCodegenRegionIntegration : KotlinIntegration {
 
     private val envVars = SectionWriter { writer, _ ->
         writer.write(
-            "private val regionOverride = #T(#S)",
-            RuntimeTypes.Core.SmokeTests.getEnv,
+            "private val regionOverride = #T.System.getenv(#S)",
+            RuntimeTypes.Core.Utils.PlatformProvider,
             "AWS_SMOKE_TEST_REGION",
         )
     }
--- a/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/smoketests/testing/SmokeTestFailHttpEngineIntegration.kt
+++ b/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/smoketests/testing/SmokeTestFailHttpEngineIntegration.kt
@@ -1,7 +1,7 @@
-package aws.sdk.kotlin.codegen.smoketests
+package aws.sdk.kotlin.codegen.smoketests.testing
 
-import aws.sdk.kotlin.codegen.model.traits.FailedResponseTrait
-import aws.sdk.kotlin.codegen.model.traits.SuccessResponseTrait
+import aws.sdk.kotlin.codegen.model.traits.testing.TestFailedResponseTrait
+import aws.sdk.kotlin.codegen.model.traits.testing.TestSuccessResponseTrait
 import software.amazon.smithy.kotlin.codegen.KotlinSettings
 import software.amazon.smithy.kotlin.codegen.core.RuntimeTypes
 import software.amazon.smithy.kotlin.codegen.core.withBlock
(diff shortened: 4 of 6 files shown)


codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/GradleGenerator.kt
```diff
@@ -16,6 +23,14 @@ import software.amazon.smithy.kotlin.codegen.rendering.GradleWriter
...
 class GradleGenerator : KotlinIntegration {
 
+    private var hasSmokeTests = AtomicBoolean(false)
+
+    // Only used to access settings - will always be true
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean {
+        hasSmokeTests.set(model.operations(settings.service).any { it.hasTrait<SmokeTestsTrait>() })
+        return super.enabledForService(model, settings)
+    }
+
```
lauzadis, line 33: You can access `ctx.settings` in `generateGradleBuild`, do you need this atomic boolean / `enabledForService`?

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/GradleGenerator.kt
```diff
@@ -69,4 +88,41 @@ class GradleGenerator : KotlinIntegration {
...
         delegator.fileManifest.writeFile("build.gradle.kts", contents)
     }
+
+    private fun generateSmokeTestJarTask(writer: GradleWriter) {
+        val moneySign = "$"
+        writer.withBlock("jvm {", "}") {
+            withBlock("compilations {", "}") {
+                write("val main = getByName(\"main\")")
+                withBlock("tasks {", "}") {
+                    withBlock("register<Jar>(\"smokeTestJar\") {", "}") {
```
lauzadis, line 98: Use #S instead of managing the quotes yourself

settings.gradle.kts
```diff
@@ -57,6 +59,13 @@ file("services").listFiles().forEach {
     }
 }
 
+// generated services by smoke tests - test suite
+file("tests/codegen/smoke-tests/services").listFiles().forEach {
+    if (it.isServiceDir) {
+        include(":tests:codegen:smoke-tests:services:${it.name}")
+    }
+}
+
```
ianbotsf, line 68: **Correctness**: The **tests/** modules are intended to test our own SDK-developed code. These smoke tests will actually be verifying model correctness, which is a slightly different scope and intent. Moreover, I believe we'll want to package these up and distribute them to test runners at some point, which is more reason for them not to live here.

Have we explored codegenning into the **services/** modules directly? We don't have to include them in the main publication (so they're _not_ included in our Maven Central artifacts) but we can create a new publication that would include them separately.

tests/codegen/smoke-tests/src/test/kotlin/SmokeTestE2ETest.kt
```diff
@@ -0,0 +1,63 @@
...
+import kotlin.test.*
+
+class SmokeTestE2ETest {
+    @Test
+    fun successService() {
+        val smokeTestRunnerOutput = runSmokeTests("successService")
+
+        assertContains(smokeTestRunnerOutput, "ok SuccessService SuccessTest - no error expected from service")
+        assertContains(smokeTestRunnerOutput, "ok SuccessService SuccessTestWithTags - no error expected from service")
+    }
```
ianbotsf, line 13: **Question**: Are any of these tests SDK specific? Could they live in **smithy-kotlin** instead?

tests/codegen/smoke-tests/src/test/resources/smoke-tests-failure.smithy
```diff
@@ -0,0 +1,62 @@
...
+            "endpoint": {
+                "url": "https://static.endpoint"
+            }
+        }
+    ]
+)
+service FailureService {
+    version: "1.0.0",
+    operations: [ TestOperation ],
+}
```
ianbotsf, line 31: **Question**: Were these model files copied or adapted from somewhere? If so, let's include a comment with a link back to them in case we ever need to refresh them. (If you wrote these from scratch then never mind!)

ianbotsf (changes requested):

tests/codegen/smoke-tests/build.gradle.kts
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    tasks.build {
+        dependsOn(tasks.getByName("stageServices"))
+    }
+
+    tasks.register<Exec>("unstageServices") {
+        commandLine("git", "clean", "-fdx", "services")
+    }
```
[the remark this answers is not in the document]
0marperez: Oh I see, this task only cleans `<sdk root dir>/tests/codegen/smoke-tests/services`. That's the directory where the code-generated services for the E2E tests go. It doesn't clean the `<sdk root dir>/services` folder. I don't see a case where we should be writing any code there but I can see how it's unsafe. I added it as a QOL utility I guess. It's not necessary though so I can get rid of it.

tests/codegen/smoke-tests/build.gradle.kts
```diff
@@ -0,0 +1,113 @@
...
+    smithyBuild {
+        val pathToSmithyModels = "src/test/resources/"
+
+        this@Build_gradle.projections.forEach { projection ->
+            projections.register(projection.name) {
+                imports = listOf(layout.projectDirectory.file(pathToSmithyModels + projection.modelFile).asFile.absolutePath)
+                smithyKotlinPlugin {
+                    serviceShapeId = projection.serviceShapeId
+                    packageName = "aws.sdk.kotlin.test.smoketests"
+                    packageVersion = "1.0"
```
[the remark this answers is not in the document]
0marperez: With the SDK version? Is there a reason why? These projections are just for testing

settings.gradle.kts
0marperez: The actual smoke tests runners are being code-generated directly into the `services` module. The path is something like: `<sdk root dir>/services/<relevant service>/generated-src-jvm/test/java/aws/sdk/kotlin/services/<relevant service>/smoketest/SmokeTests.kt`. This block of code is for including the code-generated services that are meant for E2E testing into the project.

tests/codegen/smoke-tests/src/test/kotlin/SmokeTestE2ETest.kt
0marperez: They're sort of SDK specific. The tests and infra around them mock the setup we have for copying directories from the smithy projections into the dedicated services folders. So mocking `<sdk root dir>/services/<relevant service>` but in `<sdk root dir>/tests/codegen/smoke-tests/services/<relevant service>` for our tests. This includes the `gg smokeTest` task that I created for the SDK repo and that is not available in smithy kotlin.

They also test the actual runner which doesn't _need_ to be tested in the SDK but it's a lot easier to test here rather than in smithy kotlin. If we move the tests to smithy kotlin we would lose a lot of test coverage around all this SDK specific stuff I mentioned.

tests/codegen/smoke-tests/src/test/resources/smoke-tests-failure.smithy
0marperez: These were very loosely based on [this example](https://smithy.io/2.0/additional-specs/smoke-tests.html#smoke-tests-example). I don't know if we need to keep up to date with it though, they're only very loosely based on the example. If something changes in the smithy syntax, then the E2E tests would probably break anyway. Do you feel like we should link that example? And do you have a reason in mind for why we should?

settings.gradle.kts
ianbotsf: Oh I think I must've gotten confused by the word "services". So **/tests/codegen/smoke-tests/services** contains generated clients for the services defined in **[smoke-tests-failure.smithy](https://github.com/awslabs/aws-sdk-kotlin/pull/1388/files/d7a4877515ccb0aad6fa5405d49b39ae01d6fa87#diff-515b98ec7be69d02b783d537019d7eaf5f6bbc07706c7e467a9469ca945e4616)** and **[smoke-tests-success.smithy](https://github.com/awslabs/aws-sdk-kotlin/pull/1388/files/d7a4877515ccb0aad6fa5405d49b39ae01d6fa87#diff-5f83884c293fdcd749bb400ed751339c5d831107d77e6e104dddb9f3c77c0b02)**? Our other **tests/** projects generate those into a **build** directory so they get cleaned automatically, don't get checked-in, etc. It also means that the set of loaded Gradle subprojects isn't changing dynamically.

For example, our [event streams test project](https://github.com/awslabs/aws-sdk-kotlin/tree/86c002df182c4d864ad74b931ce9be45faa77f72/tests/codegen/event-stream) codegens the client, model, and serde files into **tests/codegen/event-stream/build/smithyprojections/event-stream/restJson1/kotlin-codegen**.

Can we do something similar here?

tests/codegen/smoke-tests/build.gradle.kts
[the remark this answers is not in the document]
ianbotsf: If config this is just for generated service clients _from the smoke tests test suite_ then it doesn't matter. Disregard!

tests/codegen/smoke-tests/src/test/resources/smoke-tests-failure.smithy
ianbotsf: No, we only need a link if it's something we think should be kept up-to-date with some source. Seems like that's not necessary here so never mind.

tests/codegen/smoke-tests/src/test/kotlin/SmokeTestE2ETest.kt
ianbotsf: > the setup we have for copying directories from the smithy projections into the dedicated services folders
Can you elaborate more? Is this setup defined in our smoke test generators or our manually-maintained build files?

0marperez: The smoke tests integration from smithy kotlin [writes](https://github.com/smithy-lang/smithy-kotlin/blob/4e58bbd672dff777e19b31fe9d0559182c180d03/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/smoketests/SmokeTestsIntegration.kt#L23) the smoke test runner file to `<relevant service smithy projection dir>/jvm-src/test/java/` and then the `stageSdks` task that we created [moves](https://github.com/awslabs/aws-sdk-kotlin/blob/d7a4877515ccb0aad6fa5405d49b39ae01d6fa87/codegen/sdk/build.gradle.kts#L179-L182) the whole `jvm-src` directory into the `<sdk root dir>/services/<relevant service>/generated-src-jvm` directory. 

This is also done for the regular service code generated code. I'm not sure where `<relevant service smithy projection dir>/src` is created but you can see where it is [copied](https://github.com/awslabs/aws-sdk-kotlin/blob/d7a4877515ccb0aad6fa5405d49b39ae01d6fa87/codegen/sdk/build.gradle.kts#L175-L178) over to `<sdk root dir>/services/<relevant service>/generated-src` in the `stageSdks` task.

settings.gradle.kts
0marperez: > So /tests/codegen/smoke-tests/services contains generated clients for the services defined in [smoke-tests-failure.smithy](https://github.com/awslabs/aws-sdk-kotlin/pull/1388/files/d7a4877515ccb0aad6fa5405d49b39ae01d6fa87#diff-515b98ec7be69d02b783d537019d7eaf5f6bbc07706c7e467a9469ca945e4616) and [smoke-tests-success.smithy](https://github.com/awslabs/aws-sdk-kotlin/pull/1388/files/d7a4877515ccb0aad6fa5405d49b39ae01d6fa87#diff-5f83884c293fdcd749bb400ed751339c5d831107d77e6e104dddb9f3c77c0b02)?

Yeah, that's right

> For example, our [event streams test project](https://github.com/awslabs/aws-sdk-kotlin/tree/86c002df182c4d864ad74b931ce9be45faa77f72/tests/codegen/event-stream) codegens the client, model, and serde files into tests/codegen/event-stream/build/smithyprojections/event-stream/restJson1/kotlin-codegen.

> Can we do something similar here?

I don't think we can because the generated test services need to be inside a gradle project so we can run the `gg smokeTest` task. I'm not sure that we can run a gradle task from the `build` directory.

ianbotsf: Our other codegen tests _also_ need to be run after generating them. The event streams tests [generate sources via our Smithy build plugin](https://github.com/awslabs/aws-sdk-kotlin/blob/6f4673f4fd087faf7c1fab6ea9db2091dbc6707d/tests/codegen/event-stream/build.gradle.kts#L46-L77) (which go into the **build** dir) and then [add each of the generated client roots to the Kotlin source](https://github.com/awslabs/aws-sdk-kotlin/blob/6f4673f4fd087faf7c1fab6ea9db2091dbc6707d/tests/codegen/event-stream/build.gradle.kts#L109-L113). Then, statically written tests can [invoke the generated code](https://github.com/awslabs/aws-sdk-kotlin/blob/6f4673f4fd087faf7c1fab6ea9db2091dbc6707d/tests/codegen/event-stream/src/test/kotlin/RpcEventStreamTests.kt#L100).

It seems to me like a similar setup could work for the smoke tests: use Smithy build plugin (may require new options/tweaks to support smoke test codegen) to generate tests into the **build** dir and then invoke those tests in statically-written code.

One of my concerns here is repo hygene and consistency. We had to add new **.gitignore** entries for this which _seems_ wrong. Maybe it's not and this is the best option but I'd like to at least verify this can't be done in a manner more similar to other codegenned tests.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/GradleGenerator.kt
```diff
@@ -97,17 +95,18 @@ class GradleGenerator : KotlinIntegration {
...
             withBlock("compilations {", "}") {
-                write("val runtimePath = configurations.getByName(#S).map { if (it.isDirectory) it else zipTree(it) }", "jvmRuntimeClasspath")
                 write("val mainPath = getByName(#S).output.classesDirs", "main")
                 write("val testPath = getByName(#S).output.classesDirs", "test")
                 withBlock("tasks {", "}") {
                     withBlock("register<Jar>(#S) {", "}", "smokeTestJar") {
                         write("description = #S", "Creates smoke tests jar")
                         write("group = #S", "application")
                         write("dependsOn(build)")
+                        write("mustRunAfter(build)")
```
lauzadis, line 105: `dependsOn` and `mustRunAfter` seem redundant, why do we need both?

lauzadis (approved):

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/GradleGenerator.kt
ianbotsf: > And when running a task with the parallel flag , dependsOn only guarantees that the task that is depended on starts first but not that it ends first.

That's kind of surprising. Is that documented somewhere? Or were you seeing errors in the build without this ordering constraint?

ianbotsf (approved):

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/GradleGenerator.kt
0marperez: I don't think it's documented anywhere but I was seeing errors in the build without the ordering constraint.

ianbotsf (approved):

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/smoketests/SmokeTestFailHttpEngineIntegration.kt
```diff
@@ -0,0 +1,49 @@
...
+import software.amazon.smithy.model.shapes.ServiceShape
+import software.amazon.smithy.smoketests.traits.SmokeTestsTrait
+
+/**
+ * Adds [FailedResponseTrait] support to smoke tests
+ */
+class SmokeTestFailHttpEngineIntegration : KotlinIntegration {
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean =
+        model.topDownOperations(settings.service).any { it.hasTrait<SmokeTestsTrait>() } &&
+            settings.sdkId !in smokeTestDenyList &&
```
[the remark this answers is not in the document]
0marperez: It's not required, I ended up removing this check

PR #1392, opened 2024-08-27
feat: annotation processor codegen configuration

lauzadis:
- Adds a skeleton for supporting codegen configuration of the annotation processor. One config field `GenerateBuilderClasses` is added with options of `WHEN_REQUIRED` (default) or `ALWAYS`. When set to `WHEN_REQUIRED`, the annotation processor will not generate builder classes if they are not necessary.  
- Adds support for generating schemas for classes annotated with `@DynamoDbSortKey` 
- Disable the plugin test package until the plugin can be applied correctly

## Description of changes

### DIFF
--- a/hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/AnnotationsProcessorOptions.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/AnnotationsProcessorOptions.kt
@@ -50,12 +50,12 @@ public sealed class DestinationPackage {
     /**
      * The package where code-generated classes / objects will be placed.
      */
-    internal abstract val pkg: String
+    public abstract val pkg: String
 
     /**
      * Constructs should be code-generated into a package relative to the class being processed. Defaults to `aws.sdk.kotlin.hll.dynamodbmapper.generatedschemas`.
      */
-    public class Relative(override val pkg: String = "aws.sdk.kotlin.hll.dynamodbmapper.generatedschemas") : DestinationPackage()
+    public class Relative(override val pkg: String = "dynamodbmapper.generatedschemas") : DestinationPackage()
 
     /**
      * Constructs should be code-generated into an absolute package. Defaults to `aws.sdk.kotlin.hll.dynamodbmapper.generatedschemas`.
--- a/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/build.gradle.kts
+++ b/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/build.gradle.kts
@@ -14,6 +14,10 @@ plugins {
     alias(libs.plugins.gradle.plugin.publish)
 }
 
+kotlin {
+    explicitApi()
+}
+
 dependencies {
     implementation(kotlin("gradle-plugin"))
     implementation(libs.ksp.gradle.plugin)
@@ -64,6 +68,7 @@ tasks.test {
     }
 }
 
+// FIXME Commonize the following functions into the aws-kotlin-repo-tools build-support
 /**
  * Create a file containing the sdkVersion to use as a resource
  * This saves us from having to manually change version numbers in multiple places
@@ -87,8 +92,8 @@ val generateSdkVersionFile by tasks.registering {
 val generateKotlinVersionFile by tasks.registering {
     val resourcesDir = layout.buildDirectory.dir("resources/main/aws/sdk/kotlin/hll/dynamodbmapper/plugins").get()
     val versionFile = file("$resourcesDir/kotlin-version.txt")
-    val gradlePropertiesFile = rootProject.file("gradle.properties")
-    inputs.file(gradlePropertiesFile)
+    val versionCatalogFile = rootProject.file("gradle/libs.versions.toml")
+    inputs.file(versionCatalogFile)
     outputs.file(versionFile)
     sourceSets.main.get().output.dir(resourcesDir)
     doLast {
--- a/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginExtension.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginExtension.kt
@@ -5,29 +5,29 @@ import aws.sdk.kotlin.hll.codegen.rendering.Visibility.PUBLIC
 import aws.sdk.kotlin.hll.dynamodbmapper.codegen.annotations.DestinationPackage
 import aws.sdk.kotlin.hll.dynamodbmapper.codegen.annotations.GenerateBuilderClasses
 
-const val SCHEMA_GENERATOR_PLUGIN_EXTENSION = "dynamoDbMapper"
+internal const val SCHEMA_GENERATOR_PLUGIN_EXTENSION = "dynamoDbMapper"
 
-open class SchemaGeneratorPluginExtension {
+public open class SchemaGeneratorPluginExtension {
     /**
      * Determines when a builder class should be generated for user classes. Defaults to "WHEN_REQUIRED".
      * With this setting, builder classes will not be generated for user classes which consist of only public mutable members
      * and have a zero-arg constructor.
      */
-    var generateBuilderClasses: GenerateBuilderClasses = GenerateBuilderClasses.WHEN_REQUIRED
+    public var generateBuilderClasses: GenerateBuilderClasses = GenerateBuilderClasses.WHEN_REQUIRED
 
     /**
      * Determines the visibility of code-generated classes / objects. Defaults to [Visibility.PUBLIC].
      */
-    var visibility: Visibility = Visibility.PUBLIC
+    public var visibility: Visibility = Visibility.PUBLIC
 
     /**
      * Determines the package where code-generated classes / objects will be placed.
      * Defaults to [DestinationPackage.Relative] from the package of the class being processed, suffixed with `aws.sdk.kotlin.hll.dynamodbmapper.generatedschemas`.
      */
-    var destinationPackage: DestinationPackage = DestinationPackage.Relative()
+    public var destinationPackage: DestinationPackage = DestinationPackage.Relative()
 
     /**
      * Determines whether a `DynamoDbMapper.get<CLASS>Table` convenience extension function will be generated. Defaults to true.
      */
-    var generateGetTableExtension: Boolean = true
+    public var generateGetTableExtension: Boolean = true
 }
--- a/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
@@ -80,7 +80,7 @@ class SchemaGeneratorPluginTest {
         val result = runner.build()
         assertContains(setOf(TaskOutcome.SUCCESS, TaskOutcome.UP_TO_DATE), result.task(":build")?.outcome)
 
-        val schemaFile = File(testProjectDir, "build/generated/ksp/main/kotlin/org/example/aws/sdk/kotlin/hll/dynamodbmapper/generatedschemas/UserSchema.kt")
+        val schemaFile = File(testProjectDir, "build/generated/ksp/main/kotlin/org/example/dynamodbmapper/generatedschemas/UserSchema.kt")
         assertTrue(schemaFile.exists())
 
         val schemaContents = schemaFile.readText()
@@ -152,7 +152,7 @@ class SchemaGeneratorPluginTest {
         val result = runner.build()
         assertContains(setOf(TaskOutcome.SUCCESS, TaskOutcome.UP_TO_DATE), result.task(":build")?.outcome)
 
-        val schemaFile = File(testProjectDir, "build/generated/ksp/main/kotlin/org/example/aws/sdk/kotlin/hll/dynamodbmapper/generatedschemas/BuilderNotRequiredSchema.kt")
+        val schemaFile = File(testProjectDir, "build/generated/ksp/main/kotlin/org/example/dynamodbmapper/generatedschemas/BuilderNotRequiredSchema.kt")
         assertTrue(schemaFile.exists())
 
         val schemaContents = schemaFile.readText()
@@ -214,7 +214,7 @@ class SchemaGeneratorPluginTest {
         val result = runner.build()
         assertContains(setOf(TaskOutcome.SUCCESS, TaskOutcome.UP_TO_DATE), result.task(":build")?.outcome)
 
-        val schemaFile = File(testProjectDir, "build/generated/ksp/main/kotlin/org/example/aws/sdk/kotlin/hll/dynamodbmapper/generatedschemas/BuilderNotRequiredSchema.kt")
+        val schemaFile = File(testProjectDir, "build/generated/ksp/main/kotlin/org/example/dynamodbmapper/generatedschemas/BuilderNotRequiredSchema.kt")
         assertTrue(schemaFile.exists())
 
         val schemaContents = schemaFile.readText()
@@ -244,7 +244,7 @@ class SchemaGeneratorPluginTest {
         val result = runner.build()
         assertContains(setOf(TaskOutcome.SUCCESS, TaskOutcome.UP_TO_DATE), result.task(":build")?.outcome)
 
-        val schemaFile = File(testProjectDir, "build/generated/ksp/main/kotlin/org/example/aws/sdk/kotlin/hll/dynamodbmapper/generatedschemas/UserSchema.kt")
+        val schemaFile = File(testProjectDir, "build/generated/ksp/main/kotlin/org/example/dynamodbmapper/generatedschemas/UserSchema.kt")
         assertTrue(schemaFile.exists())
 
         val schemaContents = schemaFile.readText()
@@ -271,7 +271,7 @@ class SchemaGeneratorPluginTest {
         val result = runner.build()
         assertContains(setOf(TaskOutcome.SUCCESS, TaskOutcome.UP_TO_DATE), result.task(":build")?.outcome)
 
-        val schemaFile = File(testProjectDir, "build/generated/ksp/main/kotlin/org/example/aws/sdk/kotlin/hll/dynamodbmapper/generatedschemas/UserSchema.kt")
+        val schemaFile = File(testProjectDir, "build/generated/ksp/main/kotlin/org/example/dynamodbmapper/generatedschemas/UserSchema.kt")
         assertTrue(schemaFile.exists())
 
         val schemaContents = schemaFile.readText()
@@ -351,7 +351,7 @@ class SchemaGeneratorPluginTest {
 
         val buildResult = runner.build()
         assertContains(setOf(TaskOutcome.SUCCESS, TaskOutcome.UP_TO_DATE), buildResult.task(":build")?.outcome)
-        val schemaFile = File(testProjectDir, "build/generated/ksp/main/kotlin/org/example/aws/sdk/kotlin/hll/dynamodbmapper/generatedschemas/UserSchema.kt")
+        val schemaFile = File(testProjectDir, "build/generated/ksp/main/kotlin/org/example/dynamodbmapper/generatedschemas/UserSchema.kt")
         assertTrue(schemaFile.exists())
 
         val testResult = runner.withArguments("test").build()
--- a/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/api/dynamodb-mapper-schema-generator-plugin.api
+++ b/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/api/dynamodb-mapper-schema-generator-plugin.api
@@ -26,7 +26,3 @@ public class aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginExte
 	public final fun setVisibility (Laws/sdk/kotlin/hll/codegen/rendering/Visibility;)V
 }
 
-public final class aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginExtensionKt {
-	public static final field SCHEMA_GENERATOR_PLUGIN_EXTENSION Ljava/lang/String;
-}
-
--- a/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/resources/tests/UserTest.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/resources/tests/UserTest.kt
@@ -6,7 +6,7 @@ package org.example
 
 import aws.sdk.kotlin.hll.dynamodbmapper.model.itemOf
 import aws.sdk.kotlin.services.dynamodb.model.AttributeValue
-import org.example.aws.sdk.kotlin.hll.dynamodbmapper.generatedschemas.UserConverter
+import org.example.dynamodbmapper.generatedschemas.UserConverter
 import kotlin.test.Test
 import kotlin.test.assertEquals
 


hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
```diff
@@ -40,7 +40,38 @@ class SchemaGeneratorPluginTest {
...
+    }
+
+    // TODO Parameterize the test across multiple versions of Kotlin and Gradle
+    @Test
+    fun `configures the plugin`() {
+        val buildFileContent = """
+        import aws.sdk.kotlin.hll.dynamodbmapper.plugins.GenerateBuilderClasses
+
+         plugins {
+             id("org.jetbrains.kotlin.jvm") version "2.0.0"
```
ianbotsf, line 68: **Nit**: We're on Kotlin version 2.0.10 now.

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
```diff
@@ -40,7 +40,38 @@ class SchemaGeneratorPluginTest {
...
+             id("aws.sdk.kotlin.hll.dynamodbmapper.schema.generator")
+         }
+         configure<aws.sdk.kotlin.hll.dynamodbmapper.plugins.SchemaGeneratorPluginExtension>{
+             generateBuilderClasses = GenerateBuilderClasses.WHEN_REQUIRED
+             generateBuilderClasses = GenerateBuilderClasses.ALWAYS
+         }
```
ianbotsf, line 74: **Question**: What exactly are we testing here?

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
```diff
@@ -40,7 +40,38 @@ class SchemaGeneratorPluginTest {
...
+            .withGradleVersion("8.5")
+            .forwardOutput()
+            .build()
+
+        assertContains(setOf(TaskOutcome.SUCCESS, TaskOutcome.UP_TO_DATE), result.task(":build")?.outcome)
+    }
+
+    // TODO Parameterize the test across multiple versions of Kotlin and Gradle
+    @Test
+    fun `configures the plugin`() {
```
ianbotsf, line 63: **Correctness**: There's a bit of duplication in this test with the existing test. Let's extract test environment/config to a helper method so we can reuse it in these tests. That'll also help with parameterizing the build later. I suspect we're going to want a suite of tests to run across a matrix of configurations, sort of like we do for our HTTP engine tests or our service benchmarks.

hll/dynamodb-mapper/tests/dynamodb-mapper-schema-generator-plugin-test/common/test/aws/sdk/kotlin/hll/dynamodbmapper/tests/plugins/UserTest.kt
```diff
@@ -4,30 +4,26 @@
...
-        )
-
-        val unconverted = UserConverter.fromItem(converted)
-
-        assertEquals(user, unconverted)
-    }
+//    @Test
+//    fun testConversion() {
+//        val user = User(123, "Steve", "Rogers", 84)
+//        val converted = UserConverter.toItem(user)
```
ianbotsf, line 13: **Nit**: Add `// FIXME`

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginExtension.kt
```diff
@@ -0,0 +1,17 @@
+package aws.sdk.kotlin.hll.dynamodbmapper.plugins
+
+const val SCHEMA_GENERATOR_PLUGIN_EXTENSION = "schemaGeneratorPluginExtension"
```
ianbotsf, line 3: **Style**: This is the name that gets provided as a convenience DSL method in users' build files right? If so, we should select a name that's more ergonomic like `"dynamoDbMapper"` or something.

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPlugin.kt
```diff
@@ -4,22 +4,23 @@
...
     override fun apply(project: Project): Unit = project.run {
-        createExtension()
+        val extension = createExtension()
         configureDependencies()
+
+        project.afterEvaluate {
+            extensions.configure<KspExtension> {
+                arg(CodegenAttributes.AlwaysGenerateBuilders.name, extension.generateBuilderClasses.name)
+            }
+        }
```
ianbotsf, line 23: **Question**: Why does this need to be applied in `project.afterEvaluate`?

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
ianbotsf, line 74: **Nit**: We should test (either here or in a new test) using the DSL method for this:

```kotlin
dynamoDbMapper {
    generateBuilderClasses = ...
}
```

ianbotsf (changes requested):

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/AnnotationsProcessor.kt
```diff
@@ -30,14 +33,20 @@ class AnnotationsProcessor(private val environment: SymbolProcessorEnvironment)
...
-        HighLevelRenderer(annotatedClasses, logger, codeGeneratorFactory).render()
+        HighLevelRenderer(annotatedClasses, logger, codeGeneratorFactory, codegenAttributes).render()
 
         return invalid
     }
 }
+
+object CodegenAttributes {
+    val AlwaysGenerateBuilders: AttributeKey<Boolean> = AttributeKey("AlwaysGenerateBuilders")
+}
```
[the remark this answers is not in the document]
lauzadis: No problem, I wasn't sure exactly what we wanted to offer for configurability, this PR sets up the framework so adding those additional config options should be easy

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -120,14 +168,18 @@ class SchemaRenderer(
             "public fun #T.#L(name: String): #T = #L(name, #L)",
             MapperTypes.DynamoDbMapper,
             fnName,
-            MapperTypes.Model.tablePartitionKey(classType, keyProperty.typeRef),
+            if (sortKeyProp != null) {
+                MapperTypes.Model.tableCompositeKey(classType, partitionKeyProp.typeRef, sortKeyProp.typeRef)
+            } else {
+                MapperTypes.Model.tablePartitionKey(classType, partitionKeyProp.typeRef)
+            },
```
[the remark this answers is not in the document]
lauzadis: I don't see the duplication besides checking `sortKeyProp != null`, one is ItemSchema and one is Table

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/model/Member.kt
```diff
@@ -11,14 +11,15 @@ import com.google.devtools.ksp.symbol.KSPropertyDeclaration
...
         /**
          * Derive a [Member] from a [KSPropertyDeclaration]
          */
         fun from(prop: KSPropertyDeclaration) = Member(
             name = prop.simpleName.getShortName(),
             type = Type.from(prop.type),
+            mutable = prop.isMutable,
         )
     }
 }
```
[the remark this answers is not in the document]
lauzadis: Yes, we could. I thought it would be safer to require providing a value, but assuming immutability seems safe too

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -38,27 +41,57 @@ class SchemaRenderer(
...
+    private val sortKeyProp = properties.singleOrNull { it.isSk }
+
+    /**
+     * We skip rendering a class builder if:
+     *   - the user has configured GenerateBuilders to WHEN_REQUIRED (default value) AND
+     *   - the class has all mutable members AND
+     *   - the class has a zero-arg constructor
+     */
+    private val shouldRenderBuilder: Boolean = run {
+        val alwaysGenerateBuilders = ctx.attributes.getOrNull(CodegenAttributes.AlwaysGenerateBuilders) ?: true
```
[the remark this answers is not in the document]
lauzadis: When using the plugin, yes, these will always be set. If somebody uses the annotation processor standalone (such as in our dynamodb-mapper-annotation-processor-test module), the attributes won't be set. 

Should we still try to support that use-case or will we assume people only use the annotation processor through the plugin?

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPlugin.kt
lauzadis: Without this, the default values of the args were passed to KSP without reading the user's extension configuration. I think it's because the plugin's `apply(...)` runs during the configuration stage which is before the configured extension is available.

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
lauzadis: Just checking that the config options exist / can be used

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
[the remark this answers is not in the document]
ianbotsf: Good point. We should probably leave the door open for using the annotation processor standalone, which means we should probably get-or-default the values into attributes at some common level like `AnnotationsProcessor`.

0marperez (approved):

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -38,27 +44,65 @@ class SchemaRenderer(
...
+    private val shouldRenderBuilder: Boolean = run {
+        val alwaysGenerateBuilders = ctx.attributes[AnnotationsProcessorOptions.GenerateBuilderClassesAttribute] == GenerateBuilderClasses.ALWAYS
+        val hasAllMutableMembers = classDeclaration.getAllProperties().all { it.isMutable }
+        val hasZeroArgConstructor = classDeclaration.getConstructors().any { constructor -> constructor.parameters.all { it.hasDefault } }
+
+        !(!alwaysGenerateBuilders && hasAllMutableMembers && hasZeroArgConstructor)
+    }
+
+    private val visibility: String = when (ctx.attributes[AnnotationsProcessorOptions.VisibilityAttribute]) {
+        Visibility.IMPLICIT -> "public"
```
lauzadis, line 75: I'm not sure what we should be doing with `Visibility.IMPLICIT`. I'm guessing it will change / get more refined in the future, but for now, `PUBLIC` and `IMPLICIT` are exactly the same.

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/build.gradle.kts
```diff
@@ -79,3 +82,54 @@ val generateSdkRuntimeVersion by tasks.registering {
...
     dependsOn(generateSdkRuntimeVersion)
 }
+
+tasks.withType<Test> {
+    dependsOn(generateSdkRuntimeVersion)
+}
+
+/**
+ * Set up Maven Local dependencies to be used in the Gradle TestKit
+ */
```
lauzadis, line 92: This is annoying but required. Without this, the Gradle TestKit execution does not have access to any dependencies 
from smithy-kotlin / aws-sdk-kotlin

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPlugin.kt
```diff
@@ -4,22 +4,32 @@
...
 public class SchemaGeneratorPlugin : Plugin<Project> {
     override fun apply(project: Project): Unit = project.run {
-        createExtension()
+        val extension = createExtension()
         configureDependencies()
+
+        project.afterEvaluate {
+            val dstPkgSerialized = when (val dstPkg = extension.destinationPackage) {
+                is DestinationPackage.RELATIVE -> "relative=${dstPkg.pkg}"
+                is DestinationPackage.ABSOLUTE -> "absolute=${dstPkg.pkg}"
```
lauzadis, line 23: Any thoughts on this serialization method? KSP can only pass strings as args

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
ianbotsf: `IMPLICIT` should be unspecified:

```kotlin
Visibility.IMPLICIT -> ""
```

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/build.gradle.kts
ianbotsf: Ooh, what about [this](https://discuss.gradle.org/t/testkit-for-plugin-that-depends-on-another-module-in-a-multi-module-build/42887/6)? Might something along those lines work?

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPlugin.kt
ianbotsf: The serialization format looks fine to me but I don't love the implementation details being spread across two modules. I suggest overriding `DestinationPackage.toString()` to generate the serialized form and adding a `DestinationPackage.fromString(...)` method to handle deserializing.

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPlugin.kt
```diff
@@ -30,17 +40,16 @@ public class SchemaGeneratorPlugin : Plugin<Project> {
...
 
-    // Reads sdk-version.txt for the SDK version to add dependencies on. The file is created in this module's build.gradle.kts
-    private fun getSdkVersion(): String = try {
-        this.javaClass.getResource("sdk-version.txt")?.readText() ?: throw IllegalStateException("sdk-version.txt does not exist")
-    } catch (ex: Exception) {
-        throw IllegalStateException("Failed to load sdk-version.txt which sets the SDK version", ex)
-    }
+// Reads sdk-version.txt for the SDK version to add dependencies on. The file is created in this module's build.gradle.kts
+internal fun getSdkVersion(): String = try {
+    SchemaGeneratorPlugin::class.java.getResource("sdk-version.txt")?.readText() ?: throw IllegalStateException("sdk-version.txt does not exist")
```
ianbotsf, line 52: **Style**: `?: throw IllegalStateException` can be replaced with [`checkNotNull`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/check-not-null.html):

```kotlin
checkNotNull(SchemaGeneratorPlugin::class.java.getResource("sdk-version.txt")?.readText()) { "sdk-version.txt does not exist" }
```

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
```diff
@@ -2,57 +2,415 @@ package aws.sdk.kotlin.hll.dynamodbmapper.plugins
...
         val buildFileContent = """
-         plugins {
-            id("org.jetbrains.kotlin.jvm") version "2.0.0"
+        repositories {
+            mavenCentral()
+            mavenLocal()
+        }
+        
+        plugins {
+            id("org.jetbrains.kotlin.jvm") version "2.0.10"
```
ianbotsf, line 39: **Nit**: Let's add a FIXME for un-hardcoding (softcoding?) the Kotlin version

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
```diff
@@ -2,57 +2,415 @@ package aws.sdk.kotlin.hll.dynamodbmapper.plugins
...
+        val existingContent = readText()
+        writeText(text)
+        appendText(existingContent)
+    }
+
+    private fun getResource(resourceName: String): String = this::class.java
+        .getResourceAsStream(resourceName)
+        ?.bufferedReader().use {
+            it?.readText()
+        } ?: error("Could not read $resourceName")
```
ianbotsf, line 71: **Question**: You used a shortened syntax elsewhere for reading a resource into a string...is there any reason that won't work here?

```kotlin
checkNotNull(this::class.java.getResource("sdk-version.txt")?.readText()) { "Could not read $resourceName" }
```

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
```diff
@@ -2,57 +2,415 @@ package aws.sdk.kotlin.hll.dynamodbmapper.plugins
...
+                    return User(
+                        id,
+                        givenName,
+                        surname,
+                        age,
+                    )
+                }
+            }
+            """.trimIndent(),
+        )
```
ianbotsf, line 120: **Question**: I'm a little iffy on asserting big codegen blocks in an integration-style test like this. Would it be feasible/better to add unit tests to the test projects and assert that Gradle ran them successfully (i.e., make the tests more _functional_ and less _literal_)?

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/rendering/BuilderRenderer.kt
```diff
@@ -14,19 +15,20 @@ class BuilderRenderer(
...
+        withBlock("#L class #L {", "}", visibility, "${className}Builder") {
             members.forEach {
-                write("public var #L: #T? = null", it.name, it.type)
+                write("#L var #L: #T? = null", visibility, it.name, it.type)
             }
             blankLine()
 
-            withBlock("public fun build(): #T {", "}", classType) {
+            withBlock("#L fun build(): #T {", "}", visibility, classType) {
                 members.forEach {
```
ianbotsf, line 32: **Ponderance**: It's slightly irksome to me how we have to either accept unnecessary whitespace from our codegen (i.e., when `visibility == ""`, then we'll generate ` fun build():…`) or jump through extra hoops to pad non-empty/null placeholder values (i.e., `val visibilityPrefix = if (visibility.isNullOrEmpty) "" else "$visibility "`). It's the same in Smithy codegen but I wonder if our HLL codegen could simplify this somehow. 🤔 I don't have a concrete suggestion nor am I requesting you come up with something in the scope of this review—just wondering how we could streamline this in the future.

ianbotsf (changes requested):

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/AnnotationsProcessor.kt
```diff
@@ -36,8 +41,30 @@ class AnnotationsProcessor(private val environment: SymbolProcessorEnvironment)
...
+
+        val generateGetTableMethod = environment.options.getOrDefault(GenerateGetTableMethodAttribute.name, "true")
+        check(generateGetTableMethod.equals("true", ignoreCase = true) || generateGetTableMethod.equals("false", ignoreCase = true)) { "Unsupported value for ${GenerateGetTableMethodAttribute.name}, expected \"true\" or \"false\", got $generateGetTableMethod" }
+
+        return attributesOf {
+            GenerateBuilderClassesAttribute to GenerateBuilderClasses.valueOf(environment.options[GenerateBuilderClassesAttribute.name] ?: GenerateBuilderClasses.WHEN_REQUIRED.name)
+            VisibilityAttribute to Visibility.valueOf(environment.options.getOrDefault(VisibilityAttribute.name, Visibility.IMPLICIT.name))
+            DestinationPackageAttribute.name to dstPkg
+            GenerateGetTableMethodAttribute.name to generateGetTableMethod.equals("true", ignoreCase = true)
+        }
```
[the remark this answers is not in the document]
lauzadis: Probably because I had to use `.name` when passing the KSP args and it was carried over from a copy-paste

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/AnnotationsProcessorOptions.kt
```diff
@@ -0,0 +1,87 @@
...
+}
+
+/**
+ * Determines the visibility of code-generated classes / objects. Defaults to [IMPLICIT].
+ */
+enum class Visibility {
+    /**
+     * An implicit visibility will be used, which is recommended for most use-cases.
+     */
+    IMPLICIT,
```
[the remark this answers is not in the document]
lauzadis: I think `Visibility.DEFAULT` sounds good

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/AnnotationsProcessorOptions.kt
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Determines the package where code-generated classes / objects will be placed.
+ * Defaults to [DestinationPackage.RELATIVE] from the package of the class being processed, suffixed with "mapper.schemas".
+ */
+sealed class DestinationPackage {
+    abstract val pkg: String
+
+    /**
+     * Constructs should be code-generated into a package RELATIVE to the class being processed. Defaults to mapper.schemas.
```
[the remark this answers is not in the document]
lauzadis: I think at one point these had link symbols `[RELATIVE]` but were lost during a refactor. I'll update to something like what you recommended

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/AnnotationsProcessorOptions.kt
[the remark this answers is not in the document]
lauzadis: Looking at the Kotlin docs, I'm now wondering why we have separate options for `Visibility.DEFAULT` and `Visibility.PUBLIC` when they accomplish the same thing (`PUBLIC` does it explicitly). What do you think about defaulting to `PUBLIC` and removing `DEFAULT`?

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/AnnotationsProcessorOptions.kt
```diff
@@ -0,0 +1,87 @@
...
+     */
+    class RELATIVE(override val pkg: String = "mapper.schemas") : DestinationPackage()
+
+    /**
+     * Constructs should be code-generated into an ABSOLUTE package. Defaults to mapper.schemas.
+     */
+    class ABSOLUTE(override val pkg: String = "mapper.schemas") : DestinationPackage()
```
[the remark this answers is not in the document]
lauzadis: Yeah I think that's a good idea

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -38,27 +44,65 @@ class SchemaRenderer(
...
+        if (ctx.attributes[AnnotationsProcessorOptions.GenerateGetTableMethodAttribute]) {
+            renderGetTable()
+        }
     }
 
     private fun renderBuilder() {
-        // TODO Not all classes need builders generated (i.e. the class consists of all public mutable members), add configurability here
         val members = classDeclaration.getAllProperties().map(Member.Companion::from).toSet()
-        BuilderRenderer(this, classType, members).render()
+        BuilderRenderer(this, classType, members, visibility).render()
```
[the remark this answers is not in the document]
lauzadis: Yeah since the BuilderRenderer is in **hll-codegen** and it uses visibility, it makes sense to lift `Visibility` to that level too

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
lauzadis: I ended up just fixing it by creating another resource file called `kotlin-version.txt`

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
lauzadis: It's feasible, and probably better. I was modeling these tests off of our other codegen tests which assert on the output. Those are a pain to update when you make a big change to the code generation so I'm in favor of replacing this with unit tests

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/rendering/BuilderRenderer.kt
lauzadis: One option (and what I did here in the latest revision) is add an extension method `Attributes.visibility` which handles the whitespace padding, then you just need to write `#Lfun build(...), ctx.attributes.visibility` and it will render all visibilities with the correct amount of whitespace.

Maybe we could make a new template token `#V` that renders the visibility and also consumes the following whitespace. So you could write `#V fun build(...)` and it will render `public ` and consume the whitespace following the `#V`

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/AnnotationsProcessorOptions.kt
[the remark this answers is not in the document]
lauzadis: We spoke offline and agreed to reduce the options to `Visibility.PUBLIC` and `Visibility.INTERNAL` for now.

A few reasons to keep `Visibility.DEFAULT` were discussed:
1. IntelliJ displays a message when the `public` modifier is used unnecessarily (in non-explicit API mode)
1. Kotlin / JetBrains may change the meaning of an implicit/default/unspecified modifier in the future
	- We can address this change when/if it happens

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/build.gradle.kts
```diff
@@ -79,3 +82,54 @@ val generateSdkRuntimeVersion by tasks.registering {
...
+tasks.register("publishSmithyKotlinToMavenLocal") {
+    if (gradle.includedBuilds.none { it.name == "smithy-kotlin" }) {
+        return@register
+    }
+
+    val smithyKotlin = gradle.includedBuild("smithy-kotlin")
+    dependsOn(smithyKotlin.task(":runtime:auth:aws-credentials:publishToMavenLocal"))
+    dependsOn(smithyKotlin.task(":runtime:auth:aws-signing-common:publishToMavenLocal"))
+    dependsOn(smithyKotlin.task(":runtime:auth:aws-signing-default:publishToMavenLocal"))
```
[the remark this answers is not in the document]
lauzadis: CI passes so I think it's safe to use Maven Local: https://github.com/awslabs/aws-sdk-kotlin/actions/runs/10686643498/job/29622369526?pr=1392

We can refactor in the future if we see issues with using it

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -38,27 +44,65 @@ class SchemaRenderer(
...
+
+    /**
+     * We skip rendering a class builder if:
+     *   - the user has configured GenerateBuilders to WHEN_REQUIRED (default value) AND
+     *   - the class has all mutable members AND
+     *   - the class has a zero-arg constructor
+     */
+    private val shouldRenderBuilder: Boolean = run {
+        val alwaysGenerateBuilders = ctx.attributes[AnnotationsProcessorOptions.GenerateBuilderClassesAttribute] == GenerateBuilderClasses.ALWAYS
+        val hasAllMutableMembers = classDeclaration.getAllProperties().all { it.isMutable }
```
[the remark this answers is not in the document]
lauzadis: `hasAllMutableMembers` _would_ be **false** in that example. Should annotation processing be ignoring private members entirely? Right now we'll still generate a builder entry / attribute descriptor for them.

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
lauzadis: I _thought_ it would be better to have unit test resources, but in practice these are pretty difficult to create and I envision will be difficult to maintain. There is no IDE support (autocompletion, error detection) when writing them because the schemas don't exist until the unit tests are run. So, I think asserting on the raw codegen output is better than having to maintain essentially plain-text unit tests.

We _can_ still have unit tests (see the last test in this file) but I think the majority of this plugin's tests can be successfully accomplished by asserting on the codegen output (i.e asserting a builder was / was not generated, asserting visibility of codegenerated classes / objects).

Do you have any oppositions to this approach? Is the issue that we're asserting on **large** blocks of text, would you rather see them made smaller?

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/build.gradle.kts
lauzadis: `dependsOn(smithyKotlin.task("publishToMavenLocal"))` doesn't work because it doesn't start with `:`. For some reason [Gradle requires included build tasks to have this prefix](https://github.com/gradle/gradle/blob/master/subprojects/composite-builds/src/main/java/org/gradle/composite/internal/DefaultIncludedBuild.java#L164). I found a seemingly related GitHub issue: https://github.com/gradle/gradle/issues/22335 

I tried reading through that link you shared but didn't see anything immediately useful. I think the solution here is fine besides having to be exceedingly verbose, which should be fixed by that GH issue. 🤔  

Or maybe we can create a new task like `:publishSubprojectsToMavenLocal` in smithy-kotlin which `dependsOn` all the subprojects' publishToMavenLocal tasks, then we can just depend on that single task.

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/rendering/BuilderRenderer.kt
ianbotsf: Yeah extension methods like you made work, they just feel hyper-specific and will lead to boilerplate. A new token type is a good idea, although handling the space after something is not the only use case: we've got code that conditionally [prepends spaces](https://github.com/smithy-lang/smithy-kotlin/blob/b3d817c14bb5a31602ee38bb7189967a4b95875f/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpBindingProtocolGenerator.kt#L183-L190), [appends non-spaces](https://github.com/smithy-lang/smithy-kotlin/blob/b3d817c14bb5a31602ee38bb7189967a4b95875f/codegen/smithy-kotlin-codegen/src/main/kotlin/software/amazon/smithy/kotlin/codegen/rendering/protocol/HttpStringValuesMapSerializer.kt#L85-L86), etc. It'd be nice to have some token or other facility to handle all these weird cases.

Again, nothing needed for now. Just noodling.

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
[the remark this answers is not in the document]
ianbotsf: Yes, I _think_ it should ignore private members altogether. 🤔 I wonder what kind of problems that could get us into...

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
ianbotsf: I think asserting on codegen blocks for unit tests is "fine"—not really great but kinda necessary in some situations. (And in those situations yes, smaller blocks = better blocks.)

My issue is that this _isn't_ strictly a unit test—it's much more of an integration test because we're testing the Gradle plugin, our usage of KSP, and our codegen stack. We're operating at the scope where individual bytes of code output don't matter as much as functionality. That said, if the alternative (e.g., maintaining unit tests as text files) is so difficult to work with we may be stuck with this for now.

So if you're convinced codegen assertions are the more maintainable approach then let's got with that for now. And take whatever reasonable steps you can to trim down assertion blocks. Thanks for looking into it!

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/build.gradle.kts
ianbotsf: OK then we can live with the verbosity for now. Maybe leave a FIXME or TODO or something related to the Gradle issue you found.

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/AnnotationsProcessorOptions.kt
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Determines the package where code-generated classes / objects will be placed.
+ * Defaults to [DestinationPackage.Relative] from the package of the class being processed, suffixed with `aws.sdk.kotlin.hll.dynamodbmapper.generatedschemas`.
+ */
+public sealed class DestinationPackage {
+    /**
+     * The package where code-generated classes / objects will be placed.
+     */
+    internal abstract val pkg: String
```
ianbotsf, line 53: **Nit**: Should be `public`

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/AnnotationsProcessorOptions.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Constructs should be code-generated into a package relative to the class being processed. Defaults to `aws.sdk.kotlin.hll.dynamodbmapper.generatedschemas`.
+     */
+    public class Relative(override val pkg: String = "aws.sdk.kotlin.hll.dynamodbmapper.generatedschemas") : DestinationPackage()
```
ianbotsf, line 58: **Correctness**: The default for relative package destinations shouldn't be something AWS-specific or very long because it's getting nested inside of a user's existing namespace. Something like `dynamodbmapper.generatedschemas` should suffice.

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/build.gradle.kts
```diff
@@ -76,6 +80,79 @@ val generateSdkRuntimeVersion by tasks.registering {
...
 
+/**
+ * Create a file containing the Kotlin version to use as a resource
+ * This saves us from having to manually change version numbers in multiple places
+ */
+val generateKotlinVersionFile by tasks.registering {
+    val resourcesDir = layout.buildDirectory.dir("resources/main/aws/sdk/kotlin/hll/dynamodbmapper/plugins").get()
+    val versionFile = file("$resourcesDir/kotlin-version.txt")
+    val gradlePropertiesFile = rootProject.file("gradle.properties")
+    inputs.file(gradlePropertiesFile)
```
ianbotsf, line 91: **Correctness**: The Kotlin version isn't specified in **gradle.properties** anymore and so setting that as the input to this task means we may not run this task at the right times. We should be binding to [**gradle/libs.versions.toml**](https://github.com/awslabs/aws-sdk-kotlin/blob/main/gradle/libs.versions.toml).

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginExtension.kt
```diff
@@ -0,0 +1,33 @@
+package aws.sdk.kotlin.hll.dynamodbmapper.plugins
+
+import aws.sdk.kotlin.hll.codegen.rendering.Visibility
+import aws.sdk.kotlin.hll.codegen.rendering.Visibility.PUBLIC
+import aws.sdk.kotlin.hll.dynamodbmapper.codegen.annotations.DestinationPackage
+import aws.sdk.kotlin.hll.dynamodbmapper.codegen.annotations.GenerateBuilderClasses
+
+const val SCHEMA_GENERATOR_PLUGIN_EXTENSION = "dynamoDbMapper"
+
+open class SchemaGeneratorPluginExtension {
```
ianbotsf, line 10: **Nit**: This `SCHEMA_GENERATOR_PLUGIN_EXTENSION` constant shouldn't be `public` (probably `internal` at best).

**Correctness**: This `SchemaGeneratorPluginExtension` should definitely be (explicitly) `public`. Seeing that it's not made me realize we're not correctly enforcing API compatibility in this module either. 🥲 We should turn on the binary compat checker and then adjust all the implicit visibilities that result.

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/build.gradle.kts
```diff
@@ -76,6 +80,79 @@ val generateSdkRuntimeVersion by tasks.registering {
...
+    outputs.file(versionFile)
+    sourceSets.main.get().output.dir(resourcesDir)
+    doLast {
+        versionFile.writeText(kotlin.coreLibrariesVersion)
+    }
+}
```
ianbotsf, line 97: **Nit**: Now that we're doing this in multiple places (including deeply-nested subprojects like this one) we should probably extract it into something simple/reusable in [**build-support**](https://github.com/awslabs/aws-kotlin-repo-tools/tree/main/build-plugins/build-support). That's well outside the scope of this PR so let's just add a FIXME here for now.

ianbotsf (changes requested):
Getting pretty close!

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginExtension.kt
lauzadis: We are enforcing API compatibility in this module (there is a .api file), but `explicitApi()` was not enabled. I'll enable it and add visibilities

ianbotsf (approved):

Issue #472, opened 2021-12-13
Provide a high-level DynamoDB client


michaelbrewer:
### Describe the feature

Provide a high-level DynamoDB client which performs object-data mapping, similar to [the DynamoDBMapper for the AWS SDK for Java](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.html).

### Is your Feature Request related to a problem?

Loading and inserting values using the attribute values apis is tedious

### Proposed Solution

see the existing aws java sdk v1 (or v2) DynamoDbMapper

### Acknowledge

- [ ] I may be able to implement this feature request

### AWS Kotlin SDK version used

all

### Platform (JVM/JS/Native)

all

### Operating System and version

all

ianbotsf:
Hi @michaelbrewer, thank you for the feature request. We'd definitely love to provide the equivalent of DynamoDbMapper in the future. I'll leave this request open to allow others to 👍 upvote and that may help us with prioritization.

ianbotsf:
Additional context from #76:
> We might be able to do this as a compiler plugin and generate rather optimized code so that we don't have to rely on annotation processing (which I believe is JVM only at the moment) OR on reflection capabilities.

> See a good high level overview of considerations from rust SDK: awslabs/aws-sdk-rust#70

popematt:
Another possibility might be to leverage [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization). I haven't looked into it too thoroughly yet, but it seems like it would be straightforward to convert classes to and from DynamoDB Json.

randyhbh:
Any update on this issue?

ianbotsf:
We have no new updates to share at this time. Providing a high-level DynamoDB client remains on our roadmap but will likely not be prioritized until the underlying SDK itself (including the low-level DynamoDB client) is stable and released as 1.0.

scmorse:
@ianbotsf Any updates on the priority / timeline for implementing an 'Enhanced' dynamodb client? 🙏

ianbotsf:
Yes, we are currently putting the finishing touches on a Dev Preview version of the high-level DynamoDB client, which will be called "DynamoDB Mapper". This version will have a subset of the full features expected in the final version and will not be recommended for use in production. But it should provide users an opportunity to test out the mapper implementation so far and give feedback on what works and what doesn't.

[...]

PR #1409, opened 2024-09-17
add DSL overloads, paginators, and better builder integration for DDB Mapper ops codegen

ianbotsf:
## Issue \#

Part of #472 

## Description of changes

This change implements the TODOed and FIXMEd features of DSL overloads, pagination, and builder support for code-generated operations. Those changes required a fair amount of refactoring so you may see some classes moved/renamed without much functional difference. The key changes are in `BuilderRenderer`, `DataTypeGenerator`, `OperationRenderer`, and `OperationsTypeRenderer`.

In flagrant violation of the one-change-per-PR principle, I also turned on explicit API mode for **hll-codegen** since I noticed it was off when I moved some classes up from **dynamodb-mapper-codegen**. That mostly results in a bunch of new explicit visibility modifiers, return types, and `@InternalSdkApi`s all over the place.


hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/ModelAttributes.kt
```diff
@@ -19,4 +21,6 @@ internal object ModelAttributes {
      * For a given high-level [Structure], this attribute key identifies the associated low-level [Structure]
      */
     val LowLevelStructure: AttributeKey<Structure> = AttributeKey("aws.sdk.kotlin.ddbmapper#LowLevelStructure")
+
+    val PaginationInfo: AttributeKey<PaginationMembers> = AttributeKey("aws.sdk.kotlin.ddbmapper#PaginationInfo")
```
lauzadis, line 25: nit: missing kdocs

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/model/Pkg.kt
```diff
@@ -0,0 +1,50 @@
...
+        public val Base: String = "aws.sdk.kotlin.hll.dynamodbmapper"
+        public val Annotations: String = "$Base.annotations"
+        public val Items: String = "$Base.items"
+        public val Model: String = "$Base.model"
+        public val Ops: String = "$Base.operations"
+        public val PipelineImpl: String = "$Base.pipeline.internal"
+        public val Values: String = "$Base.values"
+        public val ScalarValues: String = "$Values.scalars"
+        public val SmithyTypeValues: String = "$Values.smithytypes"
+    }
```
lauzadis, line 25: Should we be keeping DDB mapper packages in the `dynamodbmapper` package?

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/model/Operation.kt
```diff
@@ -0,0 +1,49 @@
...
+         * Derive an [Operation] from a [KSFunctionDeclaration]
+         */
+        public fun from(declaration: KSFunctionDeclaration): Operation {
+            val op = Operation(
+                methodName = declaration.simpleName.getShortName(),
+                request = Structure.from(declaration.parameters.single().type),
+                response = Structure.from(declaration.returnType!!),
+            )
+
+            return ModelParsingPlugin.transform(op, ModelParsingPlugin::postProcessOperation)
```
lauzadis, line 46: Why does only this `from` function mean that an operation has been processed? What if you create an `Operation` with the primary constructor, should the postProcessOperation run in that case too?

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
```diff
@@ -37,13 +37,43 @@ internal class OperationsTypeRenderer(
...
+    }
+
+    private fun renderDslOps() = operations
+        .filterNot { it.appliesToAncestorKind() }
+        .forEach(::renderDslOp)
+
+    private fun renderDslOp(op: Operation) {
+        val builderType = BuilderRenderer.builderType(op.request.type)
+        val generics = op.request.genericVars().asParamsList(" ").also {
+            println("Generics for ${op.request} = $it")
```
lauzadis, line 59: nit: stray println?

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/HighLevelOpsProcessor.kt
```diff
@@ -33,6 +35,9 @@ internal class HighLevelOpsProcessor(environment: SymbolProcessorEnvironment) :
         if (!invoked) {
             invoked = true
 
+            val loggerOutStream = KspLoggerOutputStream(logger)
+            System.setOut(loggerOutStream.asPrintStream())
```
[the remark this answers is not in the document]
ianbotsf: This was really useful in debugging deep codegen stacks where the KSP logger wasn't passed through. Regular `println` to stdout appear to be swallowed by KSP (or possibly Gradle?) so I needed a way to access the logger despite not having it in scope. I thought it could be useful in future debug scenarios but if we feel this doesn't merit permanent inclusion (or perhaps warrants a better _solution_) I can remove/update as requested.

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/model/Pkg.kt
ianbotsf: Yes, moving them here was a mistake. Will fix!

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/model/Operation.kt
ianbotsf: Well we cannot modify the return value from a constructor—it _always_ return the newly-initialized instance.

More broadly, the intent of the plugin was to modify model shapes after they've been parsed. Since that parsing happens in the `from` methods it seemed reasonable to limit plugin application to that spot. If we _could_ move it to the constructor (or more likely a factory function) then we'd still have a case where plugins would execute any time we constructed the object in unit tests or other non-model cases.

hll/dynamodb-mapper/dynamodb-mapper-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/HighLevelOpsProcessor.kt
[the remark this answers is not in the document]
lauzadis: Ok that's what I thought. I've also ran into this issue in the past, so I'm happy to have this feature, as long as we remove the println's before merging. Could you add the same thing in the high level class of the annotations processor?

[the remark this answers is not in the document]
ianbotsf: How about a base class from which all of our symbol processors inherit? 😃

lauzadis (approved):

0marperez (approved):

PR #1401, opened 2024-09-09
DDB Mapper filter expressions (runtime components)

ianbotsf:
## Issue \#

Related to #76

## Description of changes

This change adds low-level filter expressions support to the runtime, although nothing uses them yet. The filter DSL is defined in the `Filter` interface and will _eventually_ (in the next PR) be part of a Lambda signature in a codegenned member for operation requests.

Something like this:

```kotlin
// Generated code
interface QueryRequest<T> {
    // Property syntax
    val filter: Filter.() -> BooleanExpr
    
    // DSL method syntax
    fun filter(block: Filter.() -> BooleanExpr)

    // …all other Query members
}
```

Callers will then use this when building their query structures to specify their filter expression (if desired):

```kotlin
// User code
table.query {
    filter {
        and(
            attr("productName") eq "Nimbus 2000",
            attr("condition") isIn setOf("New", "Refurbished"),
            attr("price") lt 500.00,
        )
    }

    // …set other Query members
}
```

In addition to the normal things, I could _especially_ use feedback on the DSL—names, conventions, types, etc. If it's incomprehensible to you then it'll probably be even worse for users!

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1424, opened 2024-09-29
DDB Mapper filter expressions (codegen components)

ianbotsf:
## Issue \#

Related to #472 

## Description of changes

The other half of https://github.com/awslabs/aws-sdk-kotlin/pull/1401, this change adds codegen support for Filter expressions and runtime/codegen support for Key Condition expressions required by Query operations.

**Key areas of change**:
* `MemberCodegenBehavior` has been expanded with new behaviors for expressions and behavior derivation has been spruced up to a new rules-based system
* `OperationRenderer` now generates the logic for instantiating an expression visitor, applying it to the various operation expressions, and setting the expression literal and attributes on the low-level request
* `BuilderRenderer` now generates DSL methods for annotated types
* `QueryTest` is now fixed up to remove the last of the hacks necessary to run queries

**Key additions**:
* `KeyFilter` and `SortKeyFilter` are new types to support Key Condition expressions in Query operations. `KeyFilter` supports partition key equality comparisons for _literal values_ only. `SortKeyFilter` contains a subset of `Filter` methods suitable for use on sort keys (again, for _literal values_ only).
* `ScanTest` now exists for the Scan operation

Aside from that, a lot of small moves, renames, and refactors to deal with the new changes.


hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/MemberCodegenBehavior.kt
```diff
@@ -60,56 +57,105 @@ internal sealed interface MemberCodegenBehavior {
...
+
+    // Hoisted members
+    Rule("tableName", Types.Kotlin.String, Hoist),
+    Rule("indexName", Types.Kotlin.String, Hoist),
+
+    // Expression literals
+    Rule("keyConditionExpression", Types.Kotlin.String, ExpressionLiteral(KeyCondition)),
+    Rule("filterExpression", Types.Kotlin.String, ExpressionLiteral(Filter)),
+
+    // TODO add support for remaining expression types
```
0marperez, line 148: Nit: 
``` kotlin
// TODO add support for remaining expression types
// Expression types
```

0marperez (approved):

hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationRenderer.kt
```diff
@@ -119,33 +158,54 @@ internal class OperationRenderer(
...
+}
+
+private inline operator fun Map<MemberCodegenBehavior, List<Member>>.invoke(
+    predicate: (MemberCodegenBehavior) -> Boolean,
+    block: Member.() -> Unit,
+) {
+    entries.forEach { (behavior, members) ->
+        if (predicate(behavior)) members.forEach(block)
     }
 }
```
lauzadis, line 210: I don't think this function is used anywhere

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/Expressions.kt
```diff
@@ -27,4 +27,6 @@ public sealed interface Expression {
  * items. This is a [marker interface](https://en.wikipedia.org/wiki/Marker_interface_pattern) which adds no additional
  * declarations.
  */
-public interface BooleanExpr : Expression
+public sealed interface BooleanExpr : Expression
+
+public sealed interface SortKeyExpr : Expression
```
lauzadis, line 32: nit: missing kdocs

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/Filter.kt
```diff
@@ -660,6 +660,14 @@ public interface Filter {
      */
     public fun AttributePath.isBetween(min: Expression, max: Expression): BooleanExpr
 
+    /**
+     * Creates a range expression for verifying the sort key is between two other expressions
```
lauzadis, line 664: nit: KDocs reference a sort key but we only use `AttributePath` here. Do only sort keys use this function in practice?

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/internal/KeyFilterImpl.kt
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.sdk.kotlin.hll.dynamodbmapper.expressions.internal
+
+import aws.sdk.kotlin.hll.dynamodbmapper.expressions.*
+import aws.sdk.kotlin.hll.dynamodbmapper.items.ItemSchema
+import aws.sdk.kotlin.hll.dynamodbmapper.util.dynamicAttr
+import aws.sdk.kotlin.hll.dynamodbmapper.util.requireNull
+
+internal data class KeyFilterImpl(override val partitionKey: Any, override val sortKey: SortKeyExpr?) : KeyFilter
```
lauzadis, line 12: should there be a runtime check ensuring `partitionKey` is a byte array, string, or number?

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/SortKeyFilter.kt
```diff
@@ -0,0 +1,449 @@
...
+ * provide methods and properties which create sort key expressions to narrow results in Query operations. Expressions
+ * are formed by referencing [sortKey] and then exercising some function upon it.
+ *
+ * For example:
+ *
+ * ```kotlin
+ * { sortKey eq 42 }
+ * ```
+ *
+ * This example creates an expression which checks whether an attribute named `foo` is equal to the value `42`.
```
lauzadis, line 23: > checks whether an attribute named `foo`

`foo` is not defined in this example, I'm assuming it's the name of the sortKey but can it be made clearer?

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/SortKeyFilter.kt
```diff
@@ -0,0 +1,449 @@
...
+ *
+ * # Ranges and sets
+ *
+ * Expressions can check whether the value of the sort key is in a given range possible values. These checks are
+ * available via the [isBetween] or [isIn] functions:
+ *
+ * ```kotlin
+ * // Checks whether the value of the sort key is between 40 and 60 (inclusive)
+ * sortKey isIn 40..60
+ * ```
```
lauzadis, line 59: nit: provide an example of `isBetween` too

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/SortKeyFilter.kt
```diff
@@ -0,0 +1,449 @@
...
+     */
+    public fun SortKey.isBetween(min: LiteralExpr, max: LiteralExpr): SortKeyExpr
+
+    /**
+     * Creates a range expression for verifying the sort key is between two other expressions
+     * @param min The lower bound value
+     * @param max The upper bound value (inclusive)
+     */
+    public fun SortKey.isBetween(min: ByteArray, max: ByteArray): SortKeyExpr =
+        isBetween(LiteralExpr(min), LiteralExpr(max))
```
lauzadis, line 376: question: this is missing support for `isBetween` with strings, is that intentional?

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/SortKeyFilter.kt
```diff
@@ -0,0 +1,449 @@
...
+        isBetween(LiteralExpr(range.start), LiteralExpr(range.endInclusive))
+
+    /**
+     * Creates a range expression for verifying the sort key is in the given range
+     * @param range The range to check
+     */
+    @Suppress("INAPPLICABLE_JVM_NAME")
+    @JvmName("isInRangeString")
+    public infix fun SortKey.isIn(range: ClosedRange<String>): SortKeyExpr =
+        isBetween(LiteralExpr(range.start), LiteralExpr(range.endInclusive))
```
lauzadis, line 394: question: Do we want to support `isIn` for binary types?

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/model/Member.kt
```diff
@@ -30,3 +40,10 @@ public data class Member(val name: String, val type: Type, val mutable: Boolean
         }
     }
 }
+
+/**
+ * Gets the low-level [Structure] equivalent for this high-level structure
```
lauzadis, line 45: nit: `Gets the low-level [Member] equivalent for this high-level Member`

hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/MemberCodegenBehavior.kt
ianbotsf: The expression types which follow the `TODO` are still part of the logical group `// Expression literals` started above.

hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/Structure.kt
```diff
@@ -24,20 +18,54 @@ internal fun Structure.toHighLevel(pkg: String): Structure {
...
+            MemberCodegenBehavior.PassThrough, is MemberCodegenBehavior.ExpressionArguments -> llMember
 
             MemberCodegenBehavior.MapAll, MemberCodegenBehavior.MapKeys ->
-                llMember.copy(type = TypeVar("T", llMember.type.nullable))
+                llMember.copy(type = TypeVar("T", nullable))
 
             MemberCodegenBehavior.ListMapAll -> {
                 val llListType = llMember.type as? TypeRef ?: error("`ListMapAll` member is required to be a TypeRef")
-                val hlListType = llListType.copy(genericArgs = listOf(TypeVar("T")), nullable = llListType.nullable)
+                val hlListType = llListType.copy(genericArgs = listOf(TypeVar("T")))
```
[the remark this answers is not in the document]
ianbotsf: It never did require setting `nullable`. This is a `copy` call which carries forward all existing values unless overridden. The previous code setting `nullable = llListType.nullable` is redundant because that's _already the value_ of `nullable`. Not sure why I wrote it that way originally but I noticed it was unnecessary when extracting `nullable` out to a separate `val` on L21.

hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationRenderer.kt
```diff
@@ -24,9 +25,19 @@ internal class OperationRenderer(
     private val ctx: RenderContext,
     private val operation: Operation,
 ) : RendererBase(ctx, operation.name) {
-    private val members = operation.request.lowLevel.members.groupBy { m ->
-        m.codegenBehavior.also { ctx.info("  ${m.name} → $it") }
-    }
+    private val requestMembers = operation
+        .request
+        .also { ctx.info("For type ${it.lowLevelName}:") }
```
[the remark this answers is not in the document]
ianbotsf: I omitted that because the type name already contains "request"/"response". (These are the low-level operation input/output types like `ScanRequest` and `QueryResponse`.)

hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationRenderer.kt
```diff
@@ -84,29 +95,57 @@ internal class OperationRenderer(
         imports += ImportDirective(operation.request.lowLevel.type, operation.request.lowLevelName)
 
         openBlock("private fun <T> #T.convert(", operation.request.type)
-        members(MemberCodegenBehavior.Hoist) { write("#L: #T, ", name, type) }
-        write("schema: #T,", MapperTypes.Items.itemSchema("T"))
+        requestMembers(MemberCodegenBehavior.Hoist) { write("#L: #T, ", name, type) }
+        write("schema: #T", MapperTypes.Items.itemSchema("T"))
```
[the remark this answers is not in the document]
ianbotsf: Good catch. No I shouldn't have dropped it since this parameter declaration is on its own line.

hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationRenderer.kt
ianbotsf: Good catch, this was left over from a previous version of the rule set and can be removed.

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/Expressions.kt
ianbotsf: Will add.

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/Filter.kt
ianbotsf: Nope, that's a copy-paste error. Will fix!

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/SortKeyFilter.kt
ianbotsf: Another copy-paste error. I'll rewrite the explanation and check the rest of the block for similar mistakes.

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/SortKeyFilter.kt
ianbotsf: Can do.

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/SortKeyFilter.kt
ianbotsf: Excellent question!

Under the hood, both `isBetween` and `isIn(ClosedRange<*>)` get converted into `BETWEEN` expressions in DynamoDB's expression syntax. (A DynamoDB `IN` operator also exists but it's only used for checking if a member is present in a set/list, which we make available via `isIn(Collection<*>)`.)

I made the decision both in `Filter` and `SortKeyFilter` to use different names for different argument types to closely mirror familiar Kotlin idioms. My reasoning was that, in idiomatic Kotlin, one would typically perform a range check with the `in` operator (something like `if (foo in 40..60)`). Unfortunately, Kotlin's ranges only work on `Comparable` types and `ByteArray` doesn't implement that interface. There's no way to use `in` to do something like `foo in someByteArray..someOtherByteArray`—that fails to compile. Neither would there be some way to reference attribute paths in an expression (e.g., `foo in attr("bar")..attr("baz")`) since those aren't comparable either.

So I needed a two-arg version of this check for `LiteralExpr` and `ByteArray` accepting a min and max. The name "in" no longer seemed appropriate since Kotlin's `in` is a binary operator and this is really a _trinary_ operation. Kotlin has no good equivalent so I decided that "between" made more sense for the two-arg methods.

Now, each of the `ClosedRange`/single-arg methods could be written as `isBetween`/min+max methods but it seemed unnecessary and I'd rather hear demand for that than add it preemptively. But the inverse is not true: none of the existing `isBetween`/min+max methods can be rewritten as a `ClosedRange`/single-arg method.

OK, now that the explanations and justifications are over, the question is "does this make sense?". Do these choices make the DSL easier or harder for users to understand/use?

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/SortKeyFilter.kt
ianbotsf: Longer answer above but [`ClosedRange`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-closed-range/) requires `Comparable` type arguments and `ByteArray` is not comparable.

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/SortKeyFilter.kt
```diff
@@ -0,0 +1,449 @@
...
+    @JvmName("isInRangeULong")
+    public infix fun SortKey.isIn(range: ClosedRange<ULong>): SortKeyExpr =
+        isBetween(LiteralExpr(range.start), LiteralExpr(range.endInclusive))
+
+    /**
+     * Creates a range expression for verifying the sort key is in the given range
+     * @param range The range to check
+     */
+    @Suppress("INAPPLICABLE_JVM_NAME")
+    @JvmName("isInRangeUShort")
```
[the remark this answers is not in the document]
ianbotsf: Unfortunately yes. Without the suppression, the compiler warns that `@JvmName` is inapplicable to this declaration. Without the `@JvmName`, the compiler warns that a platform type conflict will occur. This is a [known compiler bug](https://youtrack.jetbrains.com/issue/KT-68532) which should be fixed in an upcoming release.

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/internal/KeyFilterImpl.kt
ianbotsf: Yes, good idea. The error message users would otherwise see would not be very helpful. (Really wish Kotlin had unions. 🥲)

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/model/Member.kt
ianbotsf: Yep, copy-paste error.

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/rendering/BuilderRenderer.kt
```diff
@@ -61,8 +61,30 @@ public class BuilderRenderer(
...
-    private fun renderProperty(member: Member) = generator.apply {
+    private fun renderProperty(member: Member) {
+        val dslInfo = member.dslInfo
+
+        if (dslInfo != null) blankLine()
+
         write("#Lvar #L: #T = null", ctx.attributes.visibility, member.name, member.type.nullable())
-        // TODO add DSL methods for structure members
+
+        ctx.info("For member $builderName.${member.name} dslInfo = $dslInfo")
```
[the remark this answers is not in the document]
ianbotsf: It was helpful for debugging but can probably be removed now.

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/SortKeyFilter.kt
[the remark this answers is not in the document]
lauzadis: It looks like that was fixed in 2.0.20 which is now released, we should just have to upgrade

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/SortKeyFilter.kt
ianbotsf: Discussed offline and decided the current naming schema makes sense for now.

lauzadis (approved):

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/SortKeyFilter.kt
ianbotsf: Actually, I think it is there in the latest revision. (**SortKeyFilter.kt**:L60-63)

I initially couldn't find it either when I searched the files on GH with my browser. But I forgot that GH hides large diffs and I had to load **SortKeyFilter.kt** before I found it.

lauzadis: Oh yeah I see it now. I did explicitly load the file but maybe was looking at a previous commit. Sorry about that!

PR #1434, opened 2024-10-07
misc: credentials search precedence change

0marperez:
## Description of changes

-Changes to the credential provider chain and profile credentials provider to abide to new specification.
-Old tests were modified to accommodate the new order and precedence specific tests were added.
-Draft of breaking change announcement


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -237,11 +248,12 @@ private fun AwsProfile.ssoSessionCreds(config: AwsSharedConfig): LeafProviderRes
...
 /**
- * Attempt to load [LeafProvider.Process] from the current profile or `null` if the current profile does not contain
+ * Attempt to load [LeafProvider.Process] from the current profile or exception if the current profile does not contain
  * a credentials process command to execute
  */
-private fun AwsProfile.processCreds(): LeafProviderResult? {
-    if (!contains(CREDENTIAL_PROCESS)) return null
+private fun AwsProfile.processCreds(): LeafProviderResult {
+    // Process is last in precedence - credentials not found means no credentials in profile
+    if (!contains(CREDENTIAL_PROCESS)) return LeafProviderResult.Err("profile ($name) did not contain credential information")
```
ianbotsf, line 256: **Nit**: Throwing the chain-level no-creds-in-profile exception feels like a concern which should live _outside_ this method (whose job it is to return _process_ credentials).

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -257,7 +269,7 @@ private fun AwsProfile.staticCreds(): LeafProviderResult {
     val secretKey = getOrNull(AWS_SECRET_ACCESS_KEY)
     val accountId = getOrNull(AWS_ACCOUNT_ID)
     return when {
-        accessKeyId == null && secretKey == null -> LeafProviderResult.Err("profile ($name) did not contain credential information")
+        accessKeyId == null && secretKey == null -> LeafProviderResult.Err("profile ($name) missing `aws_access_key_id` & `aws_secret_access_key`")
         accessKeyId == null -> LeafProviderResult.Err("profile ($name) missing `aws_access_key_id`")
         secretKey == null -> LeafProviderResult.Err("profile ($name) missing `aws_secret_access_key`")
         else -> {
```
ianbotsf, line 275: **Comment**: Nice improvement to this error message. 👍

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
ianbotsf, line 1: **Question**: It feels like the precedence logic is a bit spread out in this file (some of which predates your changes). How easy/possible would it be to centralize the logic into a single place, preferably as a data structure like a list, so that we had a single place to look for the ordering specifically? (Not saying to move the actual resolution logic into a single place, just the precedence order.)

ianbotsf (changes requested):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
0marperez: It's possible, and I think it would make this file more readable. I think it would take maybe a day or two of just working on this to refactor it to use a list and have the assume role logic fit into that.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
0marperez: Thanks!

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -53,11 +63,9 @@ internal data class ProfileChain(
...
                 }
 
-                // when chaining assume role profiles, SDKs MUST terminate the chain as soon as they hit a profile with static credentials
-                if (visited.size > 1) {
-                    leaf = profile.staticCredsOrNull()
-                    if (leaf != null) break@loop
-                }
+                // static credentials have the highest precedence
+                leaf = profile.staticCredsOrNull()
+                if (leaf != null) break@loop
```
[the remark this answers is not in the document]
0marperez: The change in order makes it unnecessary. The check was there because static credentials were of highest precedence in the assume role profile only. Now that static credentials are always highest precedence there's no need to check if we're in the `1 + n` visited  profile.

.changes/announcement.md
```diff
@@ -0,0 +1,44 @@
...
+
+The SDK will also be changing the order in which credentials are resolved from
+in the shared credentials and config files. The new order will be:
+
+1. Static credentials
+2. Assume role with source profile OR assume role with named provider (mutually exclusive)
+3. Web identity token
+4. SSO session
+5. Legacy SSO
+6. Process
```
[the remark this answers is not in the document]
0marperez: I didn't combine them because of concerns I had about how users would tell apart profile only credentials locations. I think maybe using `4.x` for the profile chain would work or indentation.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
ianbotsf: OK then unless you've already started on it then let's skip the refactor for now. Let's just address the other comments and get this shipped.

.changes/0b5b53ab-70c0-4c1b-a445-8663ae86d6d1.json
lauzadis, line 1: correctness: Add requiresMinorVersionBump

.changes/announcement.md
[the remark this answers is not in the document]
lauzadis: It can also be made clearer if you separate these into two separate sections (`## Default credentials provider chain`, `## Profile credentials provider`)

lauzadis (commented):
correctness: this PR is missing the emission of business metrics

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/DefaultChainCredentialsProvider.kt
```diff
@@ -85,10 +86,10 @@ public class DefaultChainCredentialsProvider constructor(
...
-    val platformProvider: PlatformProvider = PlatformProvider.System,
-    val httpClient: HttpClientEngine? = null,
-    val region: String? = null,
+public class StsWebIdentityProvider(
+    private val platformProvider: PlatformProvider = PlatformProvider.System,
+    private val httpClient: HttpClientEngine? = null,
+    private val region: String? = null,
 ) : CloseableCredentialsProvider {
     override suspend fun resolve(attributes: Attributes): Credentials {
         val wrapped = StsWebIdentityCredentialsProvider.fromEnvironment(platformProvider = platformProvider, httpClient = httpClient, region = region)
```
lauzadis, line 95: `StsWebIdentityCredentialsProvider` is already public, why do we need to make the wrapper class public too?

0marperez: The wrapper delays any exceptions until `resolve()` is called so it can be part of a credentials provider chain. Without the wrapper some exceptions are thrown outside of the `resolve()` call and not caught by the credentials provider chain.

For example using the client below, if there's no valid env config for `StsWebIdentityCredentialsProvider` an exception is thrown early.
```kotlin
S3Client{
    region = "us-west-2"
    credentialsProvider = CredentialsProviderChain(
        SystemPropertyCredentialsProvider(),
        EnvironmentCredentialsProvider(),
        StsWebIdentityCredentialsProvider.fromEnvironment(),
        ProfileCredentialsProvider(),
        EcsCredentialsProvider(),
        ImdsCredentialsProvider(),
    )
}.use { client ->
    client.listBuckets(ListBucketsRequest{})
}
```
```
Exception in thread "main" aws.sdk.kotlin.runtime.auth.credentials.ProviderConfigurationException: Required field `roleArn` could not be automatically inferred for StsWebIdentityCredentialsProvider. Either explicitly pass a value, set the environment variable `AWS_ROLE_ARN`, or set the JVM system property `aws.roleArn`
```
Instead of what we want:
```
Exception in thread "main" aws.smithy.kotlin.runtime.identity.IdentityProviderException: No identity could be resolved from the chain: CredentialsProviderChain -> SystemPropertyCredentialsProvider -> EnvironmentCredentialsProvider -> StsWebIdentityProvider -> ProfileCredentialsProvider -> EcsCredentialsProvider -> ImdsCredentialsProvider .......
```

0marperez: Tbh I think the name doesn't help

lauzadis: I think it's fine for the chain to fail early when users are manually configuring a custom chain. We just don't want it to fail early in the default chain because not every user will have a valid environment config

0marperez: I think we should avoid failing early in any credentials provider chain not just the default one. IMO a credentials provider chain is supposed to keep looking for credentials and only fail once the options are exhausted. If we fail early it could be seen as a bug.

ianbotsf (approved):

PR #1441, opened 2024-10-15
misc: upgrade dependencies

lauzadis:
## Description of changes

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -9,8 +9,8 @@ coroutines-version = "1.9.0"
 atomicfu-version = "0.25.0"
 
 # smithy-kotlin codegen and runtime are versioned separately
-smithy-kotlin-runtime-version = "1.3.16"
-smithy-kotlin-codegen-version = "0.33.16"
+smithy-kotlin-runtime-version = "1.3.17"
+smithy-kotlin-codegen-version = "0.33.17"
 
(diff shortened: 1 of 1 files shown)


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/ConfigSection.kt
```diff
@@ -8,6 +8,7 @@ package aws.sdk.kotlin.runtime.config.profile
 import aws.sdk.kotlin.runtime.InternalSdkApi
 
 @InternalSdkApi
+@ConsistentCopyVisibility // Opt-in to future internal copy() generation https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-consistent-copy-visibility/
```
ianbotsf, line 11: **Correctness**: As shown in the API dump, this is a breaking change because it hides previously-`public` `copy` methods. What modules outside of **aws-config** use this class?

lauzadis: Its use is limited to **aws-config**. The other option is to apply `@ExposedCopyVisibility` to keep the current behavior, do you prefer that?

0marperez (approved):

PR #1442, opened 2024-10-16
misc: credentials business metrics

0marperez:
## Description of changes

-Emitting credentials related business metrics
-Tests for credentials related business metrics


aws-runtime/aws-config/common/test-resources/default-provider-chain/imds_assume_role/test-case.json
```diff
@@ -7,7 +7,8 @@
       "secret_access_key": "secretkeycorrect",
       "session_token": "tokencorrect",
       "expiry": 1632249686,
-      "accountId": "130633740322"
+      "accountId": "130633740322",
+      "business_metrics": ["p","0", "i"]
```
lauzadis, line 11: If this is just meant to resolve credentials from IMDS, why are there three credentials provider metrics emitted?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/profile/ProfileChain.kt
```diff
@@ -108,6 +108,12 @@ internal data class RoleArn(
      * ARN of role to assume
      */
     val roleArn: String,
+
+    /**
+     * The source used to create the [RoleArn]
+     */
+    val roleArnSource: RoleArnSource? = null,
```
lauzadis, line 115: naming: `source` makes this more readable, `roleArn.source` vs. `roleArn.roleArnSource`

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProviderTest.kt
```diff
@@ -318,4 +326,286 @@ class ProfileCredentialsProviderTest {
         val expected = credentials("AKID-Default", "Default-Secret", accountId = "12345")
         assertEquals(expected, actual)
     }
+
+    @Test
+    fun staticCredentialsBusinessMetrics() = runTest {
```
lauzadis, line 331: naming: static -> profile

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/BusinessMetricsIntegration.kt
```diff
@@ -65,4 +66,28 @@ class BusinessMetricsIntegration : KotlinIntegration {
             )
         }
     }
+
+    private val credentialsBusinessMetricsMiddleware = object : ProtocolMiddleware {
+        override val name: String = "credentialsOverrideBusinessMetricsMiddleware"
+        override fun render(ctx: ProtocolGenerator.GenerationContext, op: OperationShape, writer: KotlinWriter) {
+            writer.withBlock(
+                "if (config.credentialsProvider.#T != #S ) {",
```
lauzadis, line 74: Compare classes directly rather than their simple names, `if (config.credentialsProvider !is ManagedCredentialsProvider)`

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/BusinessMetricsIntegration.kt
```diff
@@ -65,4 +66,28 @@ class BusinessMetricsIntegration : KotlinIntegration {
...
+                /**
+                 * If a [CredentialsProvider] is not wrapped by [aws.sdk.kotlin.runtime.auth.credentials.internal.ManagedCredentialsProvider]
+                 * it means that it was not created by the SDK, suggesting that the user has supplied their own [CredentialsProvider].
+                 *
+                 * See:[AwsServiceConfigIntegration.CredentialsProviderProp]
+                 */
+                "ManagedCredentialsProvider",
+            ) {
+                write(
+                    "op.context.#T(#T.Credentials.CREDENTIALS_CODE)",
```
lauzadis, line 86: This middleware will emit `CREDENTIALS_CODE` any time a custom credentials provider is configured and used. I thought that this metric referred to using the `StaticCredentialsProvider`. 

If someone configures a custom `ProfileCredentialsProvider`, should we really not be emitting the `CREDENTIALS_PROFILE` metric?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/StsWebIdentityCredentialsProvider.kt
```diff
@@ -42,6 +44,10 @@ public class StsWebIdentityCredentialsProvider(
     public val platformProvider: PlatformProvider = PlatformProvider.System,
     public val httpClient: HttpClientEngine? = null,
 ) : CredentialsProvider {
+    /**
+     * Indicates if the class was created using [fromEnvironment]
+     */
+    private var createdFromEnvironment: Boolean = false
```
lauzadis, line 50: I understand why you needed to add this but it seems a little hacky. 

We only call `StsWebIdentityCredentialsProvider.fromEnvironment` in one spot, the default chain, can the metrics emission for that be lifted to that point and then we can remove this `createdFromEnvironment` here?

aws-runtime/aws-config/common/test-resources/default-provider-chain/imds_assume_role/test-case.json
0marperez: The test actually uses [this](https://github.com/awslabs/aws-sdk-kotlin/blob/main/aws-runtime/aws-config/common/test-resources/default-provider-chain/imds_assume_role/fs/home/.aws/config) config file. So `p` is for the named provider, `0` is for IMDS, and `i` is for the assume role call with the credentials resolved from IMDS. I think it might be a good idea to change the docs for this test because I made the same assumption when I first added the business metrics.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/BusinessMetricsIntegration.kt
0marperez: `CREDENTIALS_CODE` is meant for when there's any code to setup credentials, so configuring a custom / non-default credentials provider, the `StaticCredentialsProvider`, session objects for boto, etc. We don't have a business metric for `StaticCredentialsProvider` but it falls under `CREDENTIALS_CODE`. We could get more specific in the future though, we would just have to discuss adding a metric for something like the `StaticCredentialsProvider` that all SDKs share.

`CREDENTIALS_PROFILE` is confusing, it's meant for static credentials in a profile. It was called something else but I think the name changed during review. Same as `StaticCredentialsProvider`, we don't have a metric for `ProfileCredentialsProvider`. The focus for the initial metrics was more on the default credentials provider chain related metrics not all credentials providers but we can go back and more metrics soon because I think they would be very useful.

We still capture _most_ of what's going on, e.g. If a user sets the `ProfileCredentialsProvider` we would see that they set a non-default credentials provider, and then we could see where in the profile the credentials are resolved from because every profile option is captured in the metrics.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/StsWebIdentityCredentialsProvider.kt
0marperez: I originally had it in the default credentials provider chain but it wasn't working correctly so I moved it here. The default credentials provider chain doesn't have attributes so I was creating some and then merging them with the ones from the resolve call, but something about that wasn't working correctly. I can try again for a bit and see if I can make it work.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/BusinessMetricsIntegration.kt
0marperez: `CREDENTIALS_CODE` is now mapped to `StaticCredentialsProvider`

aws-runtime/aws-http/api/aws-http.api
ianbotsf, line 1: **Correctness**: More backward-incompatible changes.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/BusinessMetricsIntegration.kt
ianbotsf: > `CREDENTIALS_CODE` is now mapped to `StaticCredentialsProvider`. It's not in the provider itself because it's an implementation detail for other providers.

I don't think I understand the point being made here. Why can't we move this specialized logic into `StaticCredentialsProvider`? This custom middleware and the symbol references seem way more complex than just tracking the metric in static code the same way we do for all other providers.

ianbotsf (changes requested):

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/interceptors/businessmetrics/AwsBusinessMetricsUtils.kt
```diff
@@ -0,0 +1,118 @@
...
+        CREDENTIALS_IMDS("0"),
+    }
+}
+
+/**
+ * Emits a business metric into [Credentials.attributes]
+ * @param identifier The identifier of the [BusinessMetric] to be emitted.
+ */
+@InternalApi
+public fun Credentials.emitBusinessMetric(identifier: String): Credentials =
```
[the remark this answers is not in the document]
0marperez: Same comment would go for every other use/variation of `emitBusinessMetric `. It's what we've been using so far.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/DefaultChainCredentialsProvider.kt
```diff
@@ -52,7 +53,10 @@ public class DefaultChainCredentialsProvider constructor(
     private val chain = CredentialsProviderChain(
         SystemPropertyCredentialsProvider(platformProvider::getProperty),
         EnvironmentCredentialsProvider(platformProvider::getenv),
-        LazilyInitializedCredentialsProvider("EnvironmentStsWebIdentityCredentialsProvider") {
+        LazilyInitializedCredentialsProvider(
+            "EnvironmentStsWebIdentityCredentialsProvider",
+            AwsBusinessMetric.Credentials.CREDENTIALS_ENV_VARS_STS_WEB_ID_TOKEN,
+        ) {
```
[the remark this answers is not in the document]
0marperez: That was originally what I had, you can see it [here](https://github.com/awslabs/aws-sdk-kotlin/pull/1442/commits/18c80e780be28756376944d12f0b78c0122e9d3f#r1804773271). There's some conversation you can follow

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/BusinessMetricsIntegration.kt
0marperez: The issue is that it's used in the [ProfileCredentialsProvider](https://github.com/awslabs/aws-sdk-kotlin/blob/290c04810ea57d17eb430733b3121bcd55a3472e/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt#L189). When testing I saw the `StaticCredentialsProvider` business metrics being emitted when using `ProfileCredentialsProvider`.

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/interceptors/businessmetrics/AwsBusinessMetricsUtils.kt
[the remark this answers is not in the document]
lauzadis: I agree with Ian, it's a "nit" comment but would improve readability

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/DefaultChainCredentialsProvider.kt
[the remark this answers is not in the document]
lauzadis: This doesn't really address my initial concern of the business metrics seeming to be hacked into the credentials provider implementation. I agree with Ian, `LazilyInitializedCredentialsProvider` should not have to know about business metrics

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
```diff
@@ -86,6 +88,7 @@ public class ProfileCredentialsProvider @InternalSdkApi constructor(
     public val httpClient: HttpClientEngine? = null,
     public val configurationSource: AwsConfigurationSource? = null,
 ) : CloseableCredentialsProvider {
+    private val credentialsBusinessMetrics: MutableSet<BusinessMetric> = mutableSetOf()
```
lauzadis, line 91: correctness: `MutableSet` does not guarantee order of elements

> A generic unordered collection of elements that does not support duplicate elements, and supports adding and removing elements.
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-mutable-set/

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/interceptors/businessmetrics/AwsBusinessMetricsUtils.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+/**
+ * Emits business metrics into [Credentials.attributes]
+ * @param metrics The [BusinessMetric]s to be emitted.
+ */
+@InternalApi
+public fun Credentials.emitBusinessMetrics(metrics: Set<BusinessMetric>): Credentials {
+    var credentials = this
+    metrics.forEach { identifier ->
```
lauzadis, line 94: nit: this is no longer an identifier but the metric itself

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
```diff
@@ -131,12 +134,21 @@ public class ProfileCredentialsProvider @InternalSdkApi constructor(
...
+            creds.attributes.getOrNull(BusinessMetrics)?.forEach { metric ->
+                credentialsBusinessMetrics.add(metric)
+            }
+
             creds = assumeProvider.resolve(attributes)
         }
 
         logger.debug { "Obtained credentials from profile; expiration=${creds.expiration?.format(TimestampFormat.ISO_8601)}" }
-        return creds
+        return creds.emitBusinessMetrics(credentialsBusinessMetrics)
```
lauzadis, line 151: note: We're calling `emitBusinessMetrics` on a `creds` which might already have business metrics applied, which would result in duplication if you weren't using a `MutableSet` (which you might not want to be, based on my other comment above)

lauzadis (changes requested):
Overall it looks good and getting close to merge. My last concern is around `LazilyInitializedCredentialsProvider` being too closely coupled with business metrics

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/ProfileCredentialsProvider.kt
0marperez: It actually does

From `MutableSet`:
> [LinkedHashSet](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-linked-hash-set/index.html)
The implementation of the [MutableSet](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-mutable-set/index.html#kotlin.collections.MutableSet) interface, backed by a [InternalMap](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-mutable-set/#) implementation.

From `LinkedHashSet`:

> The implementation of the [MutableSet](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-mutable-set/index.html#kotlin.collections.MutableSet) interface, backed by a [InternalMap](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-linked-hash-set/#) implementation.

> The insertion order is preserved by the corresponding [InternalMap](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-linked-hash-set/#) implementation.

You can double check we're using `LinkedHashSet` by getting the `::class.simpleName` of the business metrics

lauzadis: That's the current implementation, but it's not guaranteed and could change in a future version of Kotlin. It's unlikely, and I'm open to keeping this `MutableSet` but just wanted to call it out 🤷

lauzadis (approved):

PR #1450, opened 2024-10-24
fix: various issues found during testing

lauzadis:
## Description of changes


hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/resources/standard-item-converters/src/NullableItem.kt
```diff
@@ -0,0 +1,35 @@
...
+        if (other !is NullableItem) return false
+
+        if (id != other.id) return false
+        if (string != other.string) return false
+        if (byte != other.byte) return false
+        if (int != other.int) return false
+        if (instant?.epochSeconds != other.instant?.epochSeconds) return false
+
+        return true
+    }
```
ianbotsf, line 34: **Question**: Why is a custom `equals` method necessary in this data class?

ianbotsf (approved):

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/resources/standard-item-converters/src/NullableItem.kt
lauzadis: Actually I don't think it is necessary, I copied it from a different test class which had an `override equals` for byte arrays / other collection types.

Can remove in a follow-on PR

PR #1474, opened 2024-11-26
misc: increase business metrics checks

0marperez:
These changes are related to malformed business metrics we are seeing in the `User-Agent` header.

Also see: https://github.com/smithy-lang/smithy-kotlin/pull/1188

## Description of changes

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -11,8 +11,8 @@ coroutines-version = "1.9.0"
 atomicfu-version = "0.25.0"
 
 # smithy-kotlin codegen and runtime are versioned separately
-smithy-kotlin-runtime-version = "1.3.26"
-smithy-kotlin-codegen-version = "0.33.26"
+smithy-kotlin-runtime-version = "1.3.27"
+smithy-kotlin-codegen-version = "0.33.27"
 
 # codegen
 smithy-version = "1.51.0"


ianbotsf (commented):
These changes look good but let's also add a check that all strings we're concatenating into the `m/` section are _actually_ 2 characters or less. If they are, great. If they're not, we should probably log an error/warning and _skip_ sending the business metrics.

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/interceptors/BusinessMetricsInterceptor.kt
```diff
@@ -34,10 +39,19 @@ public class BusinessMetricsInterceptor : HttpInterceptor {
...
-    val metricsString = metrics.joinToString(",", "m/") { it.identifier }
+private fun formatMetrics(metrics: MutableSet<BusinessMetric>, logger: Logger): String {
+    val allowedMetrics = metrics.filter {
+        if (it.identifier.length > 2) {
+            logger.warn { "Business metric '${it.identifier}' will be skipped due to length being > 2" }
+            false
+        } else {
+            true
+        }
+    }
```
ianbotsf, line 52: **Suggestion**: I really hope we can discover where these are coming from. Let's add to the log message something like "This is likely a bug. Please raise an issue at \<url>".

ianbotsf (approved):

PR #1475, opened 2024-11-27
feat: support default checksums

0marperez:
## Description of changes


0marperez:
Note to reviewers: I'll address the failing CI checks and clean up the PR soon. In the meantime, I’d appreciate a review of the checksum-related changes.

tests/codegen/checksums/src/commonTest/kotlin/utils/ChecksumTestUtils.kt
```diff
@@ -0,0 +1,69 @@
+package utils
```
lauzadis, line 1: package seems incorrect

tests/codegen/checksums/src/commonTest/kotlin/ChecksumResponseTest.kt
lauzadis, line 1: Missing package / copyright header. Applies to other test files

buildSrc/src/main/kotlin/shared/CodegenTest.kt
```diff
@@ -0,0 +1,19 @@
+package shared
```
lauzadis, line 1: Package is incorrect

tests/codegen/event-stream/src/commonTest/resources/event-stream-model-template.smithy
```diff
@@ -1,15 +1,15 @@
...
 use aws.api#service
 use aws.auth#sigv4
 
-@{protocol-name}
+@restJson1
 @sigv4(name: "event-stream-test")
 @service(sdkId: "EventStreamTest")
 service TestService { version: "123", operations: [TestStreamOp] }
 
-{op-traits}
```
lauzadis, line 12: How do these `{protocol-name}` and `{op-traits}` templates work now?

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/checksums/ResolveRequestChecksumCalculation.kt
```diff
@@ -0,0 +1,29 @@
...
+ */
+@InternalSdkApi
+public suspend fun resolveRequestChecksumCalculation(platform: PlatformProvider = PlatformProvider.System, profile: LazyAsyncValue<AwsProfile>): HttpChecksumConfigOption {
+    val unparsedString = AwsSdkSetting.AwsRequestChecksumCalculation.resolve(platform) ?: profile.get().requestChecksumCalculation
+    return unparsedString?.let {
+        when (unparsedString.uppercase()) {
+            "WHEN_SUPPORTED" -> HttpChecksumConfigOption.WHEN_SUPPORTED
+            "WHEN_REQUIRED" -> HttpChecksumConfigOption.WHEN_REQUIRED
+            else -> throw ConfigurationException(
+                "'$it' is not a valid value for request checksum calculation. Valid values are: ${HttpChecksumConfigOption.entries.toTypedArray()}",
```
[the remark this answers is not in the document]
0marperez: You're right, thanks

buildSrc/src/main/kotlin/shared/CodegenTest.kt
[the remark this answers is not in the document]
0marperez: This is an abstraction we can use for codegen tests (smoke tests, event streams, checksums, + any new ones we add)

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsRequest.kt
```diff
@@ -26,8 +29,44 @@ class FlexibleChecksumsRequest : KotlinIntegration {
...
+        override fun isEnabledFor(ctx: ProtocolGenerator.GenerationContext, op: OperationShape): Boolean =
+            op.hasTrait<HttpChecksumTrait>()
+
+        override fun render(ctx: ProtocolGenerator.GenerationContext, op: OperationShape, writer: KotlinWriter) {
+            writer.withBlock("when(config.requestChecksumCalculation) {", "}") {
+                writer.write(
+                    "#T.WHEN_SUPPORTED -> op.context.#T(#T.FLEXIBLE_CHECKSUMS_REQ_WHEN_SUPPORTED)",
+                    RuntimeTypes.SmithyClient.Config.HttpChecksumConfigOption,
+                    RuntimeTypes.Core.BusinessMetrics.emitBusinessMetric,
+                    RuntimeTypes.Core.BusinessMetrics.SmithyBusinessMetric,
```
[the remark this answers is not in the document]
0marperez: These are two different runtime types, `emitBusinessMetric` & `SmithyBusinessMetric`. What do you mean?

services/s3/e2eTest/src/S3ChecksumTest.kt
```diff
@@ -0,0 +1,128 @@
...
+import org.junit.jupiter.api.*
+import java.io.File
+import java.io.FileInputStream
+import java.util.*
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@TestMethodOrder(MethodOrderer.OrderAnnotation::class)
+class S3ChecksumTest {
+    private val client = S3Client { region = "us-west-2" }
+    private val testBucket = "s3-test-bucket-ci-motorcade"
```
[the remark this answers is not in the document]
0marperez: We have to use the hardcoded bucket because it's the only one allowlisted. If we don't we'll get `S3Exception: Checksum Type mismatch occurred, expected checksum Type: null, actual checksum Type: crc32`

services/s3/e2eTest/src/S3ChecksumTest.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    @AfterAll
+    private fun cleanUp(): Unit = runBlocking {
+        deleteMultiPartUploads(client, testBucket)
+        deleteBucketContents(client, testBucket)
+        client.close()
+    }
+
+    @Test
+    @Order(1)
```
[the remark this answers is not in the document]
0marperez: I had tests set up in a way that required them to run in a certain order before

tests/codegen/checksums/src/commonTest/kotlin/utils/ChecksumTestUtils.kt
0marperez: I noticed that none of the codegen tests were in packages. I put them all in packages, that should fix this

tests/codegen/event-stream/src/commonTest/resources/event-stream-model-template.smithy
0marperez: I turned them off for some reason. I added the templates back, they still work

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsRequest.kt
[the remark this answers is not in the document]
lauzadis: I was meaning your two separate calls to `emitBusinessMetric` but I'm failing to see/remember what style refactor I was trying to suggest. This should be fine as is

tests/codegen/checksums/src/commonTest/kotlin/ChecksumResponseTest.kt
lauzadis: Also are we running ktlint on these files? It should have caught the missing copyright header.

services/s3/e2eTest/src/S3ChecksumTest.kt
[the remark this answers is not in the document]
lauzadis: Should we add a TODO to move back to a random bucket once the allowlist is no longer necessary?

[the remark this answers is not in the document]
0marperez: Yes, good idea

tests/codegen/checksums/src/commonTest/kotlin/ChecksumResponseTest.kt
0marperez: I think ktlint didn't run on these files because they weren't in a package

services/s3/common/src/aws/sdk/kotlin/services/s3/express/S3ExpressDisableChecksumInterceptor.kt
```diff
@@ -22,14 +26,45 @@ internal class S3ExpressDisableChecksumInterceptor : HttpInterceptor {
         }
 
         val logger = coroutineContext.logger<S3ExpressDisableChecksumInterceptor>()
+        logger.warn { "Checksums must not be sent with S3 express upload part operation, removing checksum(s)" }
```
ianbotsf, line 29: **Suggestion**: We have a [dedicated `warn` extension method](https://github.com/smithy-lang/smithy-kotlin/blob/394eb3b2f975b111e234da89fd901a250bc6faf2/runtime/observability/telemetry-api/common/src/aws/smithy/kotlin/runtime/telemetry/logging/CoroutineContextLogExt.kt#L147-L156) (and `info`, `debug`, et al.) on `CoroutineContext` itself, which is useful when you only need to log a single message in a given scope:

```kotlin
coroutineContext.warn<S3ExpressDisableChecksumInterceptor> { "Checksums must not be sent with S3 Express UploadPart operation, removing checksum(s)" }
```

ianbotsf, line 29: **Nit**: Use proper feature name (S3 Express) and operation name (UploadPart) in message:

```kotlin
"Checksums must not be sent with S3 Express UploadPart operation, removing checksum(s)"
```

services/s3/common/src/aws/sdk/kotlin/services/s3/express/S3ExpressDisableChecksumInterceptor.kt
```diff
@@ -22,14 +26,45 @@ internal class S3ExpressDisableChecksumInterceptor : HttpInterceptor {
...
-        val configuredChecksumAlgorithm = context.executionContext.getOrNull(HttpOperationContext.ChecksumAlgorithm)
+/**
+ * Removes any checksums sent in the request's headers
+ */
+internal fun HeadersBuilder.removeChecksumHeaders(): Unit =
+    names().forEach { name ->
+        if (name.startsWith(CHECKSUM_HEADER_PREFIX)) {
+            remove(name)
+        }
+    }
```
ianbotsf, line 49: **Correctness**: Use `ignoreCase = true`. Applies to `removeChecksumTrailingHeaders` and `removeChecksumTrailingHeadersFromXAmzTrailer` methods too.

services/s3/common/test/aws/sdk/kotlin/services/s3/express/ChecksumRemovalTest.kt
```diff
@@ -0,0 +1,83 @@
+package aws.sdk.kotlin.services.s3.express
+
+import aws.smithy.kotlin.runtime.http.DeferredHeadersBuilder
+import aws.smithy.kotlin.runtime.http.HeadersBuilder
+import kotlin.test.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class ChecksumRemovalTest {
```
ianbotsf, line 9: **Nit**: Add tests for varying capitalizations of header names.

services/s3/e2eTest/src/PaginatorTest.kt
```diff
@@ -41,7 +41,11 @@ class PaginatorTest {
         S3TestUtils.deleteBucketAndAllContents(client, testBucket)
     }
 
+    // FIXME: Enable test when motorcade is ready
+    // Seeing: S3Exception: Checksum Type mismatch occurred, expected checksum Type: null, actual checksum Type: crc32
+    // Cause: "Post-motorcade SDK is expected not to work against Pre-Motorcade S3"
     // ListParts has a strange pagination termination condition via [IsTerminated]. Verify it actually works correctly.
+    @Ignore
     @Test
     fun testListPartsPagination() = runBlocking {
```
ianbotsf, line 50: **Question**: Do we _have_ to disable this test? Can't we just adjust the assertion to simplify verify the parts have the correct `partNumber`/`eTag` and ignore the checksum algorithm?

services/s3/e2eTest/src/S3ChecksumTest.kt
```diff
@@ -0,0 +1,124 @@
...
+            it.deleteOnExit()
+        }
+        val inputStream = FileInputStream(tempFile)
+
+        client.putObject {
+            bucket = testBucket
+            key = testObject + UUID.randomUUID()
+            body = ByteStream.fromInputStream(inputStream, testString.length.toLong())
+        }
+    }
```
ianbotsf, line 70: **Correctness**: We should be asserting something in these tests. Please verify the operation invocations had the desired effects by calling `GetObject` or similar and inspecting the resulting body and checksum.

services/s3/e2eTest/src/S3ChecksumTest.kt
```diff
@@ -0,0 +1,124 @@
...
+                )
+            }
+        }
+
+        client.getObject(
+            GetObjectRequest {
+                bucket = testBucket
+                key = testObject
+            },
+        ) {}
```
ianbotsf, line 122: **Correctness**: We should be fetching the object body and asserting something here.

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -98,6 +101,45 @@ object S3TestUtils {
...
 
+    suspend fun getBucketByName(
+        client: S3Client,
+        bucket: String,
+        region: String? = null,
+        accountId: String? = null,
+    ): String = withTimeout(60.seconds) {
+        val buckets = client.listBuckets()
+            .buckets
+            ?.mapNotNull { it.name }
```
ianbotsf, line 112: **Correctness**: This call does not paginate and may fail to find the desired bucket name if it's not in the first page of results. Fortunately, reading through the entire paginated list is not required. The `HeadBucket` operation provides all we need:

```kotlin
client
    .headBucket {
        this.bucket = bucket
        expectedBucketOwner = accountId
    }
    ?.takeIf { it.bucketRegion == region }
    ?: run {
        println("Creating S3 bucket: $bucket")

        client.createBucket {
            this.bucket = bucket
            createBucketConfiguration {
                locationConstraint = BucketLocationConstraint.fromValue(region ?: client.config.region!!)
            }
        }

        client.waitUntilBucketExists { this.bucket = bucket }
    }
```

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -98,6 +101,45 @@ object S3TestUtils {
...
+
+        var testBucket = buckets?.firstOrNull { bucketName ->
+            bucketName == bucket &&
+                region?.let {
+                    client.getBucketLocation {
+                        this.bucket = bucketName
+                        expectedBucketOwner = accountId
+                    }.locationConstraint?.value == region
+                } ?: true
+        }
```
ianbotsf, line 122: **Question**: This verifies an existing bucket is in the expected region but, if that isn't true, the following logic would attempt to _create_ a bucket with the same name in a different region. Won't this fail? What should we do in this case?

services/s3/e2eTest/src/S3TestUtils.kt
```diff
@@ -314,4 +364,16 @@ object S3TestUtils {
...
+        client.listMultipartUploads {
+            bucket = bucketName
+        }.uploads?.forEach { upload ->
+            client.abortMultipartUpload {
+                bucket = bucketName
+                key = upload.key
+                uploadId = upload.uploadId
+            }
+        }
+    }
```
ianbotsf, line 378: **Question**: How many multipart uploads do we expect to delete? Should we launching/joining these?

tests/codegen/build.gradle.kts
```diff
@@ -0,0 +1,96 @@
...
+
+subprojects {
+    apply(plugin = libraries.plugins.aws.kotlin.repo.tools.smithybuild.get().pluginId)
+    apply(plugin = libraries.plugins.kotlin.multiplatform.get().pluginId)
+
+    val optinAnnotations = listOf(
+        "kotlin.RequiresOptIn",
+        "aws.smithy.kotlin.runtime.InternalApi",
+        "aws.sdk.kotlin.runtime.InternalSdkApi",
+    )
```
ianbotsf, line 22: **Question**: Why do we need to opt into `RequiresOptIn`?

tests/codegen/checksums/src/commonTest/kotlin/aws/sdk/kotlin/tests/codegen/checksums/utils/ChecksumTestUtils.kt
```diff
@@ -0,0 +1,74 @@
...
+ * Checks if the specified business metrics are set in an HTTP request.
+ */
+internal class BusinessMetricsReader(
+    private val expectedBusinessMetrics: Set<BusinessMetric>,
+) : HttpInterceptor {
+    var containsExpectedBusinessMetrics = true
+
+    override fun readBeforeTransmit(context: ProtocolRequestInterceptorContext<Any, HttpRequest>) {
+        containsExpectedBusinessMetrics = context.executionContext[BusinessMetrics].containsAll(expectedBusinessMetrics)
+    }
```
ianbotsf, line 73: **Correctness**: Initializing `containsExpectedBusinessMetrics` to `true` will result in a false positive if the interceptor never runs.

ianbotsf (changes requested):

services/s3/e2eTest/src/PaginatorTest.kt
0marperez: I don't think so, this test fails because we're getting an exception from S3 for sending a request with a checksum. We're not failing the assertion.

services/s3/e2eTest/src/S3TestUtils.kt
0marperez: Currently just one per bucket, I don't think we need to be launching/joining these yet.

tests/codegen/build.gradle.kts
0marperez: I refactored a lot of things here and I don't remember. I think it's because of [this](https://github.com/awslabs/aws-sdk-kotlin/blob/fb94044e163e77cd973edb382b00f40721f54da0/aws-runtime/build.gradle.kts#L59-L61). I'll try removing it to see if anything breaks

services/s3/e2eTest/src/S3TestUtils.kt
0marperez: I think we can throw our own exception explaining the problem

services/s3/e2eTest/src/S3TestUtils.kt
0marperez: I like this solution but I think it won't work when the bucket doesn't exist. From the s3:headBucket docs:
>If the bucket does not exist or you do not have permission to access it, the HEAD request returns a generic 400 Bad Request, 403 Forbidden or 404 Not Found code. A message body is not included, so you cannot determine the exception beyond these HTTP response codes.

tests/codegen/build.gradle.kts
0marperez: We don't actually need it, I removed it

services/s3/e2eTest/src/S3TestUtils.kt
ianbotsf: Ah too bad. Well we should still use `HeadBucket` (perhaps wrapped in a helper method) or some other operation that gets us the bucket info in *O*(1) time instead of iterating through them when we already have the name.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/express/S3ExpressIntegration.kt
```diff
@@ -132,52 +129,35 @@ class S3ExpressIntegration : KotlinIntegration {
         }
     }
 
-    /**
-     * For any operations that require a checksum, set CRC32 if the user has not already configured a checksum.
-     */
-    private val useCrc32Checksum = object : ProtocolMiddleware {
```
lauzadis, line 138: question/correctness: S3 Express' spec says CRC32 is **required**. If we change our SDK's default algorithm to CRC64NVME in the future, how will we ensure CRC32 is still used for S3 Express? That's what this integration was responsible for.

Does S3 Express support CRC64NVME?

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/interceptors/S3CompositeChecksumsInterceptor.kt
```diff
@@ -0,0 +1,47 @@
...
+    }
+}
+
+/**
+ * Removes any checksum headers that contain composite checksums.
+ */
+internal fun HeadersBuilder.removeCompositeChecksums(logger: Logger): Unit =
+    names().forEach { header ->
+        if (header.startsWith(CHECKSUM_HEADER_PREFIX, ignoreCase = true) && get(header)!!.isCompositeChecksum()) {
+            logger.warn { "S3 returned a composite checksum, composite checksums are not supported, removing checksum" }
```
lauzadis, line 35: info: Log the composite checksum header name

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/s3/S3CompositeChecksumsIntegration.kt
```diff
@@ -0,0 +1,44 @@
...
+import software.amazon.smithy.kotlin.codegen.model.hasTrait
+import software.amazon.smithy.kotlin.codegen.rendering.protocol.ProtocolGenerator
+import software.amazon.smithy.kotlin.codegen.rendering.protocol.ProtocolMiddleware
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.model.shapes.ServiceShape
+
+/**
+ * Removes composite checksums returned by S3 so that flexible checksums won't validate them.
+ * Composite checksums are used for multipart uploads and end with "-#" where "#" is a number.
```
lauzadis, line 19: correctness/breaking change: I don't think we should be removing these headers, what if users need them for other purposes?

Isn't it easier to just ignore composite checksums in the flexible checksums implementation?

services/s3/common/src/aws/sdk/kotlin/services/s3/express/S3ExpressDisableChecksumInterceptor.kt
```diff
@@ -6,30 +6,72 @@ package aws.sdk.kotlin.services.s3.express
...
+import aws.smithy.kotlin.runtime.telemetry.logging.warn
 import kotlin.coroutines.coroutineContext
 
+private const val CHECKSUM_HEADER_PREFIX = "x-amz-checksum-"
+private const val S3_EXPRESS_ENDPOINT_PROPERTY_KEY = "backend"
+private const val S3_EXPRESS_ENDPOINT_PROPERTY_VALUE = "S3Express"
+
 /**
- * Disable checksums entirely for s3:UploadPart requests.
+ * Disables checksums for s3:UploadPart requests that use S3 express.
```
lauzadis, line 22: nit: S3 Express

tests/codegen/event-stream/src/commonTest/resources/event-stream-model-template.smithy
lauzadis: > I added the templates back

I still don't see them

buildSrc/build.gradle.kts
lauzadis, line 1: Does this need to exist? [buildSrc is treated as an included build](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources) which means every subproject will have these config options applied.

It seems odd to need this just for codegen tests

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/PresignerGenerator.kt
```diff
@@ -265,6 +277,45 @@ class PresignerGenerator : KotlinIntegration {
                 )
             }
 
+            checksumAlgorithmMember(op, ctx)?.let { checksumAlgorithmMember ->
+                withBlock("input.#L?.value?.let { checksumAlgorithmString ->", "}", checksumAlgorithmMember) {
```
lauzadis, line 281: nit/naming/style: avoid Hungarian notation, `checksumAlgorithmName`

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/PresignerGenerator.kt
```diff
@@ -265,6 +277,45 @@ class PresignerGenerator : KotlinIntegration {
...
+                                )
+                            }
+                            withBlock(" else {", "}") {
+                                withBlock("#T {", "}", runBlocking) {
+                                    write("class Presigner")
+                                    write(
+                                        "#T.#T<Presigner> { #S }",
+                                        coroutineContext,
+                                        warn,
+                                        "The requested checksum algorithm is not supported for pre-signed URL checksums, sending request without checksum.",
```
lauzadis, line 309: We should include which checksum algorithm the user chose and which are supported

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/PresignerGenerator.kt
```diff
@@ -287,4 +338,24 @@ class PresignerGenerator : KotlinIntegration {
      * > "my-object/example/photo.user". This is an incorrect path for that object.
      */
     private fun normalizeUriPath(service: ServiceShape) = service.sdkId != "S3"
+
+    /**
+     * Gets the checksum algorithm member if a user can configure request checksums otherwise null
```
lauzadis, line 343: > if a user can configure request checksums

style/docs: This seems awkward to me

"Gets the checksum algorithm member if configured on a request, otherwise null"

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsRequest.kt
```diff
@@ -26,8 +29,44 @@ class FlexibleChecksumsRequest : KotlinIntegration {
...
+                documentation = "Configures request checksum calculation"
+                propertyType = ConfigPropertyType.RequiredWithDefault("HttpChecksumConfigOption.WHEN_SUPPORTED")
+            },
+        )
+
     override fun customizeMiddleware(ctx: ProtocolGenerator.GenerationContext, resolved: List<ProtocolMiddleware>) =
-        resolved + flexibleChecksumsRequestMiddleware
+        resolved + flexibleChecksumsRequestMiddleware + configBusinessMetrics
+
+    private val configBusinessMetrics = object : ProtocolMiddleware {
```
lauzadis, line 47: nit: middleware name and value's name should match

buildSrc/build.gradle.kts
0marperez: It's the best way I could find to share a data class between subprojects, do you know of alternatives that would work better? I might've missed them

lauzadis: You can make a new project that both of those subprojects depend on

0marperez: This is still a work in progress, I don't want to block the review because of this. I think worst case scenario we can come back and refactor this without it being a breaking change

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/interceptors/S3FlexibleChecksumResponseInterceptor.kt
```diff
@@ -0,0 +1,27 @@
+package aws.sdk.kotlin.runtime.http.interceptors
+
+import aws.smithy.kotlin.runtime.client.config.HttpChecksumConfigOption
+import aws.smithy.kotlin.runtime.http.interceptors.FlexibleChecksumsResponseInterceptor
+
+/**
+ * S3 variant of the flexible checksum interceptor where composite checksums are not validated
```
lauzadis, line 7: nit/docs: `... of the [FlexibleChecksumsResponseInterceptor]...`

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/interceptors/S3FlexibleChecksumResponseInterceptor.kt
lauzadis, line 1: correctness: This file should go in `services/s3/common/src/aws/sdk/kotlin/services/s3/internal` with all the other S3-specific interceptors

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsRequest.kt
```diff
@@ -6,59 +6,132 @@ package aws.sdk.kotlin.codegen.customization.flexiblechecksums
...
+            writer.write("config.requestChecksumCalculation,")
+            writer.write("input.#L?.value,", requestAlgorithmMemberName)
         }
     }
 }
+
+/**
+ * Determines if an operation is set up to send flexible request checksums
+ */
+private fun requestChecksumsConfigured(ctx: ProtocolGenerator.GenerationContext, op: OperationShape): Boolean {
```
lauzadis, line 128: naming: "configured" implies that this is dependent on user behavior, but it's just based on the model. I suggest making this an extension function `OperationShape.hasRequestAlgorithmMember(...)`

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsResponse.kt
```diff
@@ -4,61 +4,108 @@
...
+        val requestValidationModeMember = ctx.model.expectShape<StructureShape>(op.input.get())
+            .members()
+            .first { it.memberName == httpChecksumTrait.requestValidationModeMember.get() }
+        val requestValidationModeMemberName = ctx.symbolProvider.toMemberName(requestValidationModeMember)
+
+        val interceptor = if (ctx.model.expectShape<ServiceShape>(ctx.settings.service).isS3) {
+            AwsRuntimeTypes.Http.Interceptors.S3FlexibleChecksumResponseInterceptor
+        } else {
+            RuntimeTypes.HttpClient.Interceptors.FlexibleChecksumsResponseInterceptor
+        }
```
lauzadis, line 100: docs: let's add a comment here about why S3 needs a custom interceptor

services/s3/common/src/aws/sdk/kotlin/services/s3/express/S3ExpressDefaultChecksumAlgorithm.kt
```diff
@@ -0,0 +1,37 @@
...
+            } else {
+                context.executionContext[HttpOperationContext.DefaultChecksumAlgorithm] = "CRC32"
+            }
+        }
+        return super.modifyBeforeSigning(context)
+    }
+}
+
+private fun usingS3Express(executionContext: ExecutionContext): Boolean =
+    executionContext.getOrNull(AttributeKey(S3_EXPRESS_ENDPOINT_PROPERTY_KEY)) != S3_EXPRESS_ENDPOINT_PROPERTY_VALUE
```
lauzadis, line 37: style: Could be an extension function off `ExecutionContext`

services/s3/common/src/aws/sdk/kotlin/services/s3/express/S3ExpressDefaultChecksumAlgorithm.kt
```diff
@@ -0,0 +1,37 @@
...
+internal const val S3_EXPRESS_ENDPOINT_PROPERTY_KEY = "backend"
+internal const val S3_EXPRESS_ENDPOINT_PROPERTY_VALUE = "S3Express"
+
+/**
+ * Re-configures the default checksum algorithm for S3 Express
+ * NOTE: Default checksums are disabled for s3:UploadPart.
+ */
+internal class S3ExpressDefaultChecksumAlgorithm(
+    private val isS3UploadPart: Boolean,
+) : HttpInterceptor {
```
lauzadis, line 23: style: This works but I think the previous implementation using two different interceptors was cleaner. It's not clear to me why this had to deviate so much from what we already had

.github/workflows/continuous-integration.yml
```diff
@@ -40,6 +40,12 @@ jobs:
           pwd
           ls -lsa
           ./gradlew -Ptest.java.version=${{ matrix.java-version }} jvmTest --stacktrace
+      - name: Save Test Reports
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: test-reports
+          path: '**/build/reports'
```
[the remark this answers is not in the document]
lauzadis: What are these test reports used for?

lauzadis (changes requested):

services/s3/common/src/aws/sdk/kotlin/services/s3/express/S3ExpressDefaultChecksumAlgorithm.kt
0marperez: Because `FlexibleChecksumsResponseInterceptor` is in smithy kotlin, which shouldn't know anything SDK related

0marperez: This was one of the suggestion we discussed during standup

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/flexiblechecksums/FlexibleChecksumsResponse.kt
0marperez: Now that I look at this again, we could replace `S3FlexibleChecksumResponseInterceptor` with something more generic that could be applied to other services. In case other services start sending composite checksums in the future. What does the team think?

tests/codegen/event-stream/src/commonTest/resources/event-stream-model-template.smithy
0marperez: I was having trouble "resetting" the model back to the original with templates. I was taking a closer look and realized that the templates weren't even required so I removed them

buildSrc/build.gradle.kts
lauzadis, line 1: I'm still a little concerned about this new `buildSrc` and the fact that it's treated as an include-build by every module. This means every module will have the Kotlin JVM plugin applied to it which shouldn't be necessary everywhere.

https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources

Can we accomplish the same thing without using buildSrc?

services/s3/e2eTest/src/PaginatorTest.kt
```diff
@@ -41,7 +41,10 @@ class PaginatorTest {
         S3TestUtils.deleteBucketAndAllContents(client, testBucket)
     }
 
+    // FIXME: Enable test
+    // Seeing: S3Exception: Checksum Type mismatch occurred, expected checksum Type: null, actual checksum Type: crc32
     // ListParts has a strange pagination termination condition via [IsTerminated]. Verify it actually works correctly.
+    @Ignore
```
lauzadis, line 47: When can this test be enabled? Is it still broken?

0marperez: Still broken, left a comment about why on slack

buildSrc/build.gradle.kts
0marperez: Discussed offline and we're keeping `buildSrc` for now

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/AwsSdkSetting.kt
```diff
@@ -208,6 +208,18 @@ public object AwsSdkSetting {
...
+     * Configures request checksum calculation
+     */
+    public val AwsRequestChecksumCalculation: EnvironmentSetting<String> =
+        strEnvSetting("aws.requestChecksumCalculation", "AWS_REQUEST_CHECKSUM_CALCULATION")
+
+    /**
+     * Configures response checksum validation
+     */
+    public val AwsResponseChecksumValidation: EnvironmentSetting<String> =
+        strEnvSetting("aws.responseChecksumValidation", "AWS_RESPONSE_CHECKSUM_VALIDATION")
```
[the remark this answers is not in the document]
0marperez: Well that's handy. I guess I forgot/couldn't find this type. Let's start using it

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
```diff
@@ -214,6 +216,32 @@ public fun AwsProfile.getLongOrNull(key: String, subKey: String? = null): Long?
...
+private fun String?.parseRequestHttpChecksumConfig(): RequestHttpChecksumConfig? =
+    when (this?.uppercase()) {
+        null -> null
+        "WHEN_SUPPORTED" -> RequestHttpChecksumConfig.WHEN_SUPPORTED
+        "WHEN_REQUIRED" -> RequestHttpChecksumConfig.WHEN_REQUIRED
+        else -> throw ConfigurationException(
+            "'$this' is not a valid value for 'request_checksum_calculation'. Valid values are: ${RequestHttpChecksumConfig.entries}",
+        )
+    }
+
```
ianbotsf, line 244: **Suggestion**: Oh, gross. I thought we had enum parsing support built for profiles but I guess not. This appears to work but relies on hardcoded name strings and wouldn't catch any new values we add to the enum. A safer way to parse would be what we do for [`retryMode`](https://github.com/awslabs/aws-sdk-kotlin/blob/757925d99d6703914ba5e33dd472e6602ee27345/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt#L79-L91) or [`accountIdEndpointMode`](https://github.com/awslabs/aws-sdk-kotlin/blob/757925d99d6703914ba5e33dd472e6602ee27345/aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt#L135-L147): iterate through the entries and pick the first matching name.

Even better would be making a generic `getEnumOrNull` function to support parsing to any enum type and then we could clean this up along with  `retryMode` and `accountIdEndpointMode`. Up to you.

ianbotsf (approved):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/profile/AwsProfile.kt
0marperez: I think `getEnumOrNull` is the right thing to do here. I can try to get this done before we have to merge the feature. Since it's not a breaking change, if I can't finish it now I'll just create a task to do it later

PR #1519, opened 2025-01-31
fix: enhance smoke test debuggability by echoing build output from inner Gradle runner

ianbotsf:
## Issue \#

(none)

## Description of changes

Similar to https://github.com/awslabs/aws-sdk-kotlin/pull/1515, enhancing smoke test debuggability by echoing build output from the inner Gradle runner to the outer Gradle process. Because the existing code waits for the inner process to exit before evaluating the output, I think we may be missing logs when the inner process exits abruptly.

### DIFF
--- a/tests/codegen/smoke-tests/src/test/kotlin/aws/sdk/kotlin/test/codegen/smoketest/SmokeTestE2ETest.kt
+++ b/tests/codegen/smoke-tests/src/test/kotlin/aws/sdk/kotlin/test/codegen/smoketest/SmokeTestE2ETest.kt
@@ -34,8 +34,6 @@ class SmokeTestE2ETest {
 
         assertContains(smokeTestRunnerOutput, "not ok ExceptionService ExceptionTest - no error expected from service")
         assertContains(smokeTestRunnerOutput, "#aws.smithy.kotlin.runtime.http.interceptors.SmokeTestsFailureException: Smoke test failed with HTTP status code: 400")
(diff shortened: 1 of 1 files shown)


tests/codegen/smoke-tests/src/test/kotlin/aws/sdk/kotlin/test/codegen/smoketest/SmokeTestE2ETest.kt
```diff
@@ -34,8 +34,6 @@ class SmokeTestE2ETest {
 
         assertContains(smokeTestRunnerOutput, "not ok ExceptionService ExceptionTest - no error expected from service")
         assertContains(smokeTestRunnerOutput, "#aws.smithy.kotlin.runtime.http.interceptors.SmokeTestsFailureException: Smoke test failed with HTTP status code: 400")
-        assertContains(smokeTestRunnerOutput, "#\tat aws.smithy.kotlin.runtime.http.interceptors.SmokeTestsInterceptor.readBeforeDeserialization(SmokeTestsInterceptor.kt:19)")
-        assertContains(smokeTestRunnerOutput, "#\tat aws.smithy.kotlin.runtime.http.interceptors.InterceptorExecutor.readBeforeDeserialization(InterceptorExecutor.kt:252)")
```
lauzadis, line 38: Did we lose the stack trace with this change?

ianbotsf: Yes, we did. Asserting on stack traces is brittle and adds no value to this test. I'd remove other parts of the exception message assertions as well but that would be a larger refactor so for now I'm just removing this.

lauzadis (approved):

PR #1526, opened 2025-02-13
misc: test union with a member that has the same name

0marperez:
## Issue \#

https://github.com/smithy-lang/smithy-kotlin/issues/1129

## Description of changes

Adds a test trying to reproduce the issue.
This issue is not reproducible anymore. It seems that an update to the Kotlin compiler has gotten rid of a bug causing it.


0marperez:
Refactored into:
https://github.com/smithy-lang/smithy-kotlin/pull/1241

PR #1535, opened 2025-02-20
fix: account ID based endpoints

0marperez:
## Issue \#

N/A

## Description of changes

Routes `accountIdEndpointMode ` to AWS built in endpoint parameters


codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/endpoints/BindAwsEndpointBuiltins.kt
```diff
@@ -106,6 +106,14 @@ fun renderBindAwsBuiltins(ctx: ProtocolGenerator.GenerationContext, writer: Kotl
                         AwsRuntimeTypes.Config.Endpoints.resolveAccountId,
                         AccountIdEndpointBuiltinCustomization.AccountIdEndpointModeProp.propertyName,
                     )
+
+                AwsBuiltins.ACCOUNT_ID_ENDPOINT_MODE -> {
```
lauzadis, line 110: The comment at the top of this function says "In practice, all of these values are sourced from the client config."

Is that still true? Isn't AccountId being sourced exclusively from endpoint rules in your manual test cases?

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/endpoints/BindAwsEndpointBuiltins.kt
```diff
@@ -106,6 +106,14 @@ fun renderBindAwsBuiltins(ctx: ProtocolGenerator.GenerationContext, writer: Kotl
...
+                    writer.write(
+                        "#L = config.#L.toString().lowercase()", // DDB endpoint rules assume lowercase value
```
lauzadis, line 112: Remove comment about DDB since this rule applies for all AWS endpoints

lauzadis (approved):

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/endpoints/BindAwsEndpointBuiltins.kt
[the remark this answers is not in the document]
0marperez: I think we can just remove the comment in that case

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/endpoints/BindAwsEndpointBuiltins.kt
0marperez: > Is that still true? Isn't AccountId being sourced exclusively from endpoint rules in your manual test cases?

Account ID isn't being sourced exclusively from endpoint rules. It's sourced from client config as a built in and from the operation context params using JMESPath. The endpoint rules use both values just for different conditions

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/endpoints/BindAwsEndpointBuiltins.kt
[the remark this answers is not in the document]
lauzadis: I think it's fine to mention that the enum values are expected to be lowercase

PR #1544, opened 2025-03-04
feat: sqs md5 checksum validation

xinsong-cui:
## Description of changes

### DIFF
--- a/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/sqs/SqsMd5ChecksumValidationIntegration.kt
+++ b/codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/sqs/SqsMd5ChecksumValidationIntegration.kt
@@ -113,7 +113,7 @@ class SqsMd5ChecksumValidationIntegration : KotlinIntegration {
             ),
         )
 
-    // add Sqs-specific config finalization
+    // add SQS-specific config finalization
     private val finalizeSqsConfigWriter = AppendingSectionWriter { writer ->
         val finalizeSqsConfig = buildSymbol {
             name = "finalizeSqsConfig"
--- a/services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
+++ b/services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
@@ -166,7 +166,7 @@ public class SqsMd5ChecksumValidationInterceptor(
 
                     validateMd5(clientSideBodyMd5, messageBodyMd5Returned)
 
-                    logger.debug { "Message body MD5 checksum for ReceiveMessage validated " }
+                    logger.debug { "Message body MD5 checksum for ReceiveMessage validated" }
                 }
             }
 
@@ -181,7 +181,7 @@ public class SqsMd5ChecksumValidationInterceptor(
 
                     validateMd5(clientSideAttrMd5, messageAttrMd5Returned)
 
-                    logger.debug { "Message attribute MD5 checksum for ReceiveMessage validated " }
+                    logger.debug { "Message attribute MD5 checksum for ReceiveMessage validated" }
                 }
             }
         }
--- a/services/sqs/e2eTest/src/SqsTestUtils.kt
+++ b/services/sqs/e2eTest/src/SqsTestUtils.kt
@@ -59,7 +59,7 @@ object SqsTestUtils {
 
             println("Queue purged successfully.")
 
-            println("Deleting Sqs queue: $queueUrl")
+            println("Deleting SQS queue: $queueUrl")
 
             client.deleteQueue(
                 DeleteQueueRequest {


services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,387 @@
...
+     */
+    private fun updateLengthAndBytes(messageDigest: Md5, binaryValue: ByteArray) {
+        println("updateLengthAndBytes")
+        val length = binaryValue.size
+        val lengthBytes = byteArrayOf(
+            (length shr 24).toByte(),
+            (length shr 16).toByte(),
+            (length shr 8).toByte(),
+            length.toByte(),
+        )
```
ianbotsf, line 368: **Comment**: Note that if you switch from directly using `Md5` to using `SdkBuffer` as recommended above, you may simply call `SdkBuffer.writeInt(binaryValue.size)` and it will handle the big-endian conversion when necessary.

services/sqs/common/src/aws.sdk.kotlin.services.sqs/internal/SQSSetting.kt
```diff
@@ -0,0 +1,49 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.sdk.kotlin.services.sqs.internal
+
+import aws.smithy.kotlin.runtime.config.*
+
+/**
+ * SQS specific system settings
+ */
+internal object SQSSetting {
```
ianbotsf, line 12: **Nit**: `SQSSetting` → `SqsSetting`

services/sqs/e2eTest/src/SqsMd5ChecksumValidationTest.kt
```diff
@@ -0,0 +1,241 @@
...
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class SqsMd5ChecksumValidationTest {
+    // An interceptor that set wrong md5 checksums in SQS response
+    private val wrongChecksumInterceptor = object : HttpInterceptor {
+        override suspend fun modifyBeforeCompletion(
+            context: ResponseInterceptorContext<Any, Any, HttpRequest?, HttpResponse?>,
+        ): Result<Any> {
+            val wrongMd5ofMessageBody = "wrong message md5".encodeToByteArray().md5().toString()
+            val wrongMd5ofMessageAttribute = "wrong attribute md5".encodeToByteArray().md5().toString()
+            val wrongMd5ofMessageSystemAttribute = "wrong system attribute md5".encodeToByteArray().md5().toString()
```
ianbotsf, line 42: **Nit**: Calling `toString()` on arrays does not produce a hex string of the contents but rather Java array notation (e.g., `[B@27bc2616`). For a true apples-to-apples comparison, you should use `.toHexString()` instead.

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,387 @@
...
+        val expectedMD5Hex = md5Digest.digest().toHexString()
+        return expectedMD5Hex
+    }
+
+    /**
+     * Update the digest using a sequence of bytes that consists of the length (in 4 bytes) of the
+     * input binaryValue and all the bytes it contains.
+     */
+    private fun updateLengthAndBytes(messageDigest: Md5, binaryValue: ByteArray) {
+        println("updateLengthAndBytes")
```
ianbotsf, line 361: **Correctness**: Remove `println`s before merging

services/sqs/e2eTest/src/SqsMd5ChecksumValidationTest.kt
```diff
@@ -0,0 +1,241 @@
...
+
+            when (val response = context.response.getOrNull()) {
+                is SendMessageResponse -> {
+                    val modifiedResponse = SendMessageResponse.invoke {
```
ianbotsf, line 46: **Nit**: `invoke` is unnecessary here. It's implied when passing a lambda argument:

```kotlin
val modifiedResponse = SendMessageResponse {
    ...
}
```

services/sqs/e2eTest/src/SqsMd5ChecksumValidationTest.kt
```diff
@@ -0,0 +1,241 @@
...
+                        messageId = response.messageId
+                        sequenceNumber = response.sequenceNumber
+                        md5OfMessageBody = wrongMd5ofMessageBody
+                        md5OfMessageAttributes = wrongMd5ofMessageAttribute
+                        md5OfMessageSystemAttributes = wrongMd5ofMessageSystemAttribute
+                    }
```
ianbotsf, line 52: **Nit**: Consider using `copy` instead to automatically inherit all the previous values and only update the ones that need changing:

```kotlin
val modifiedResponse = response.copy {
    md5OfMessageBody = wrongMd5ofMessageBody
    md5OfMessageAttributes = wrongMd5ofMessageAttribute
    md5OfMessageSystemAttributes = wrongMd5ofMessageSystemAttribute
}
```

services/sqs/e2eTest/src/SqsTestUtils.kt
```diff
@@ -0,0 +1,108 @@
...
+    ): String = getQueueUrlWithPrefix(client, prefix, region)
+
+    suspend fun getQueueUrlWithPrefix(
+        client: SqsClient,
+        prefix: String,
+        region: String? = null,
+    ): String = withTimeout(60.seconds) {
+        val queueUrls = client.listQueues().queueUrls
+
+        var matchingQueueUrl = queueUrls?.firstOrNull { url ->
```
ianbotsf, line 40: **Correctness**: This retrieves only [the first page of queues](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ListQueues.html):

> The response includes a maximum of 1,000 results.

While it's highly unlikely that any of our test accounts will have more than 1,000 queues in it, we should still use the paginated method for this:

```kotlin
val matchingQueueUrl = client
    .listQueuesPaginated()
    .queueUrls()
    .firstOrNull { ... }

```

ianbotsf, line 40: **Nit**: SQS provides a parameter to `listQueues` which filters based on a queue prefix, which reduces client-side filtering:

```kotlin
val matchingQueueUrl = client
    .listQueuesPaginated { queuePrefix = prefix }
    .queueUrls()
    .firstOrNull { ... }
```

services/sqs/e2eTest/src/SqsTestUtils.kt
```diff
@@ -0,0 +1,108 @@
...
+            val queueUrl = URI(url).toURL()
+            val hostParts = queueUrl.host.split(".")
+
+            val regionMatches = if (region != null) {
+                hostParts.getOrNull(1)?.equals(region, ignoreCase = true) ?: false
+            } else {
+                true
+            }
```
ianbotsf, line 48: **Question**: This region-matching behavior seems brittle (i.e., we don't control the naming convention for SQS's endpoints) and I wonder if it's even necessary. An SQS client initialized in a region should only return queues _in that region_ right?

services/sqs/e2eTest/src/SqsTestUtils.kt
```diff
@@ -0,0 +1,108 @@
...
+import kotlinx.coroutines.withTimeout
+import java.net.URI
+import java.util.*
+import kotlin.time.Duration.Companion.seconds
+
+object SqsTestUtils {
+    const val DEFAULT_REGION = "us-west-2"
+
+    const val TEST_QUEUE_WRONG_CHECKSUM_PREFIX = "sqs-test-queue-"
+    const val TEST_QUEUE_CORRECT_CHECKSUM_PREFIX = "sqs-test-queue-"
```
ianbotsf, line 20: **Question**: These two values are the same. Later, when you initialize the queue URLs, won't this just return the same queue twice (i.e., created by the first call, found by the second call)?

```kotlin
@BeforeAll
private fun setUp(): Unit = runBlocking {
    correctChecksumTestQueueUrl = getTestQueueUrl(correctChecksumClient, TEST_QUEUE_CORRECT_CHECKSUM_PREFIX, DEFAULT_REGION)
    wrongChecksumTestQueueUrl = getTestQueueUrl(wrongChecksumClient, TEST_QUEUE_WRONG_CHECKSUM_PREFIX, DEFAULT_REGION)
}
```

services/sqs/e2eTest/src/SqsTestUtils.kt
```diff
@@ -0,0 +1,108 @@
...
+                queueName = matchingQueueUrl
+            }
+        } else {
+            println("Using existing Sqs queue: $matchingQueueUrl")
+        }
+
+        matchingQueueUrl
+    }
+
+    suspend fun deleteQueueAndAllMessages(client: SqsClient, queueUrl: String): Unit = coroutineScope {
```
ianbotsf, line 70: **Question**: Why `coroutineScope`?

services/sqs/e2eTest/src/SqsTestUtils.kt
```diff
@@ -0,0 +1,108 @@
...
+                dataType = "String"
+                stringValue = TEST_MESSAGE_ATTRIBUTES_VALUE + batchId
+            },
+        )
+        messageSystemAttributes = hashMapOf(
+            MessageSystemAttributeNameForSends.AwsTraceHeader to MessageSystemAttributeValue {
+                dataType = "String"
+                stringValue = TEST_MESSAGE_SYSTEM_ATTRIBUTES_VALUE + batchId
+            },
+        )
```
ianbotsf, line 106: **Nit**: Generally, specifying `hashMapOf` is unnecessary. Prefer `mapOf` instead.

ianbotsf (changes requested):

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/sqs/SqsMd5ChecksumValidationIntegration.kt
```diff
@@ -0,0 +1,50 @@
...
+import software.amazon.smithy.kotlin.codegen.rendering.protocol.ProtocolGenerator
+import software.amazon.smithy.kotlin.codegen.rendering.protocol.ProtocolMiddleware
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.shapes.OperationShape
+import software.amazon.smithy.model.shapes.ServiceShape
+
+/**
+ * Register interceptor to handle SQS message MD5 checksum validation.
+ */
+class SqsMd5ChecksumValidationIntegration : KotlinIntegration {
```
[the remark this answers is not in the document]
xinsong-cui: I was trying to follow the pattern of our s3 integrations. Yeah, I agree with those two should be in a same integration

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,387 @@
...
+            if (!messageBodySent.isNullOrEmpty()) {
+                logger.debug { "Validating message body MD5 checksum for SendMessage" }
+
+                val bodyMD5Returned = sendMessageResponse.md5OfMessageBody
+                val clientSideBodyMd5 = calculateMessageBodyMd5(messageBodySent)
+                if (clientSideBodyMd5 != bodyMD5Returned) {
+                    throw ChecksumMismatchException("Checksum mismatch. Expected $clientSideBodyMd5 but was $bodyMD5Returned")
+                }
+            }
+        }
```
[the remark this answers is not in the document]
xinsong-cui: Yes, I see SonarCloud also complaining about this, I will try to do some deduplication

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,387 @@
...
+            if (!messageAttrSent.isNullOrEmpty()) {
+                logger.debug { "Validating message attribute MD5 checksum for SendMessage" }
+
+                val messageAttrMD5Returned = sendMessageResponse.md5OfMessageAttributes
+                val clientSideAttrMd5 = calculateMessageAttributesMd5(messageAttrSent)
+                if (clientSideAttrMd5 != messageAttrMD5Returned) {
+                    throw ChecksumMismatchException("Checksum mismatch. Expected $clientSideAttrMd5 but was $messageAttrMD5Returned")
+                }
+            }
+        }
```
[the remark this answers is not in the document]
xinsong-cui: That should be correct, I did some testing, Sqs will not return with attributes checksum if no message attribute was sent.

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,387 @@
...
+        val expectedMD5Hex = expectedMD5.toHexString()
+        return expectedMD5Hex
+    }
+
+    /**
+     * Calculates the MD5 digest for message attributes according to SQS specifications.
+     * https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html#sqs-attributes-md5-message-digest-calculation
+     */
+    @OptIn(InternalApi::class, ExperimentalStdlibApi::class)
+    private fun calculateMessageAttributesMd5(messageAttributes: Map<String, MessageAttributeValue>): String {
```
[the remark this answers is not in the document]
xinsong-cui: Yes `@InternalApi` seems redundant, we can get rid of that.
`toHexString()` is experimental.

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,387 @@
...
+        val md5Digest = Md5()
+
+        try {
+            for (attributeName in sortedAttributeNames) {
+                val attributeValue = messageAttributes[attributeName]
+                updateLengthAndBytes(md5Digest, attributeName.encodeToByteArray())
+
+                attributeValue?.dataType?.let { dataType ->
+                    updateLengthAndBytes(md5Digest, dataType.encodeToByteArray())
+                }
```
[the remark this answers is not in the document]
xinsong-cui: Datatype should be non-null, 
`public val dataType: kotlin.String = requireNotNull(builder.dataType) { "A non-null value must be provided for dataType" }`

My code is kinda confusing I can refactor that

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,387 @@
...
+                    "to false when instantiating the client." + e.message,
+            )
+        }
+        val expectedMD5Hex = md5Digest.digest().toHexString()
+        return expectedMD5Hex
+    }
+
+    private fun calculateMessageSystemAttributesMd5(
+        messageSysAttrs: Map<MessageSystemAttributeNameForSends, MessageSystemAttributeValue>,
+    ): String {
```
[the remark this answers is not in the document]
xinsong-cui: Yes, I see SonarCloud also complaining about this, I will try to do some deduplication

services/sqs/e2eTest/src/SqsTestUtils.kt
xinsong-cui: Yes you are right, I was referring to our s3 implementation and it seems like s3 works different.

services/sqs/e2eTest/src/SqsTestUtils.kt
xinsong-cui: Oh, I want to made the prefix different but forgot to change it, nice catch!

services/sqs/common/src/aws.sdk.kotlin.services.sqs/internal/FinalizeSqsConfig.kt
```diff
@@ -0,0 +1,35 @@
...
+    sharedConfig: LazyAsyncValue<AwsSharedConfig>,
+    provider: PlatformProvider = PlatformProvider.System,
+) {
+    val activeProfile = sharedConfig.get().activeProfile
+    builder.config.checksumValidationEnabled = builder.config.checksumValidationEnabled
+        ?: SQSSetting.checksumValidationEnabled.resolve(provider)
+        ?: activeProfile.checksumValidationEnabled
+        ?: ValidationEnabled.NEVER // TODO: MD5 checksum validation is temporarily disabled. Set default to ALWAYS in next minor version
+
+    builder.config.checksumValidationScopes = builder.config.checksumValidationScopes.ifEmpty {
```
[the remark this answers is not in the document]
xinsong-cui: Yeah you are right I had some thoughts here: https://github.com/awslabs/aws-sdk-kotlin/pull/1544/files#r1980231768

services/sqs/common/src/aws.sdk.kotlin.services.sqs/internal/SQSSetting.kt
lauzadis: I disagree with `SqsSettings`, we use [the singular form for S3](https://github.com/awslabs/aws-sdk-kotlin/blob/main/services/s3/common/src/aws/sdk/kotlin/services/s3/internal/S3Setting.kt) and it works better in my opinion

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/sqs/SqsModelUtils.kt
```diff
@@ -0,0 +1,14 @@
...
+package aws.sdk.kotlin.codegen.customization.sqs
+
+import aws.sdk.kotlin.codegen.sdkId
+import software.amazon.smithy.model.shapes.ServiceShape
+
+/**
+ * Returns true if the service is Sqs
+ */
+val ServiceShape.isSqs: Boolean
+    get() = sdkId.lowercase() == "sqs"
```
lauzadis, line 14: This is only used in one other place, I think it's not worth defining a function / new file for this

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,330 @@
...
+ * - [checksumValidationScopes] - Specifies which message components to validate
+ *
+ * Supported operations:
+ * - SendMessage
+ * - SendMessageBatch
+ * - ReceiveMessage
+ */
+@OptIn(ExperimentalStdlibApi::class)
+public class SqsMd5ChecksumValidationInterceptor(
+    private val validationEnabled: ValidationEnabled?,
```
lauzadis, line 40: Why is this nullable if the default value is `NEVER`?

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/sqs/SqsMd5ChecksumValidationIntegration.kt
```diff
@@ -0,0 +1,125 @@
...
+                name = "ValidationEnabled"
+                namespace = "aws.sdk.kotlin.services.sqs.internal"
+            }
+            documentation = """
+                Specifies when MD5 checksum validation should be performed for SQS messages. This controls the automatic 
+                calculation and validation of checksums during message operations.
+                
+                Valid values:
+                - `ALWAYS` (default) - Checksums are calculated and validated for both sending and receiving operations 
+                  (SendMessage, SendMessageBatch, and ReceiveMessage)
```
lauzadis, line 44: `ALWAYS` is not the current default, we should update the documentation once we do the minor version bump

codegen/aws-sdk-codegen/src/test/kotlin/aws/sdk/kotlin/codegen/customization/sqs/SqsMd5ChecksumValidationIntegrationTest.kt
```diff
@@ -0,0 +1,52 @@
...
+
+        assertEquals(listOf(FooMiddleware, SqsMd5ChecksumValidationMiddleware), actual)
+    }
+}
+
+object FooMiddleware : ProtocolMiddleware {
+    override val name: String = "FooMiddleware"
+    override fun render(ctx: ProtocolGenerator.GenerationContext, op: OperationShape, writer: KotlinWriter) =
+        fail("Unexpected call to `FooMiddleware.render`")
+}
```
lauzadis, line 52: nit/style: This is only used in one test (`testMiddlewareAddition`) and could be defined there rather than a top-level object

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,330 @@
...
+
+        val logger = context.executionContext.coroutineContext.logger<SqsMd5ChecksumValidationInterceptor>()
+
+        // Test MD5 availability
+        try {
+            "MD5".encodeToByteArray().md5()
+        } catch (e: Exception) {
+            logger.error { "MD5 checksums are not available (likely due to FIPS mode). Checksum validation will be disabled." }
+            return
+        }
```
lauzadis, line 61: How does FIPS mode work in practice? This is [using smithy-kotlin's Md5JVM implementation](https://github.com/smithy-lang/smithy-kotlin/blob/main/runtime/runtime-core/jvm/src/aws/smithy/kotlin/runtime/hashing/Md5JVM.kt), customers don't have the ability to swap implementations.

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,330 @@
...
+
+    private fun sendMessageOperationMd5Check(
+        sendMessageRequest: SendMessageRequest,
+        sendMessageResponse: SendMessageResponse,
+        logger: Logger,
+    ) {
+        if (validationScopes.contains(ValidationScope.MESSAGE_BODY)) {
+            val messageBodySent = sendMessageRequest.messageBody
+
+            if (!messageBodySent.isNullOrEmpty()) {
```
lauzadis, line 102: I think this condition should be `if (!sendMessageResponse.md5OfMessageBody.isNullOrEmpty())` rather than checking the request. Same for messageAttributes, messageSystemAttributes.

If SQS returns a non-null MD5, we should validate it

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,330 @@
...
+public class SqsMd5ChecksumValidationInterceptor(
+    private val validationEnabled: ValidationEnabled?,
+    private val validationScopes: Set<ValidationScope>,
+) : HttpInterceptor {
+    public companion object {
+        private const val STRING_TYPE_FIELD_INDEX: Byte = 1
+        private const val BINARY_TYPE_FIELD_INDEX: Byte = 2
+        private const val STRING_LIST_TYPE_FIELD_INDEX: Byte = 3
+        private const val BINARY_LIST_TYPE_FIELD_INDEX: Byte = 4
+    }
```
lauzadis, line 48: Do these need to be a companion object or can they be top-level `private const val`?

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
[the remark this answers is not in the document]
lauzadis: You marked this as resolved but I still feel like there's a lot of duplication. What did you change?

services/sqs/common/src/aws.sdk.kotlin.services.sqs/internal/ValidationConfig.kt
```diff
@@ -0,0 +1,46 @@
...
+ * Valid values:
+ * - ALWAYS - Validates checksums for both sending and receiving operations
+ *           (SendMessage, SendMessageBatch, and ReceiveMessage)
+ * - WHEN_SENDING - Validates checksums only when sending messages
+ *                 (SendMessage and SendMessageBatch)
+ * - WHEN_RECEIVING - Validates checksums only when receiving messages
+ *                   (ReceiveMessage)
+ * - NEVER - Disables checksum validation completely
+ *
+ * Default: ALWAYS
```
lauzadis, line 18: The default is `NEVER` until v1.5.x, so I think we should update the documentation at that point

services/sqs/e2eTest/src/SqsMd5ChecksumValidationTest.kt
```diff
@@ -0,0 +1,225 @@
...
+                }
+            }
+            return context.response
+        }
+    }
+
+    private val correctChecksumClient = SqsClient {
+        region = DEFAULT_REGION
+        checksumValidationEnabled = ValidationEnabled.ALWAYS
+        checksumValidationScopes = ValidationScope.entries.toSet()
```
lauzadis, line 90: Do you need to set this if the default is all validation scopes?

services/sqs/e2eTest/src/SqsMd5ChecksumValidationTest.kt
```diff
@@ -0,0 +1,225 @@
...
+    // used for wrong checksum tests
+    private val wrongChecksumClient = SqsClient {
+        region = DEFAULT_REGION
+        checksumValidationEnabled = ValidationEnabled.ALWAYS
+        checksumValidationScopes = ValidationScope.entries.toSet()
+        interceptors += wrongChecksumInterceptor
+    }
+
+    private lateinit var correctChecksumTestQueueUrl: String
+    private lateinit var wrongChecksumTestQueueUrl: String
```
lauzadis, line 102: We should be able to use the same queue for both of these tests

services/sqs/e2eTest/src/SqsMd5ChecksumValidationTest.kt
```diff
@@ -0,0 +1,225 @@
...
+    @AfterAll
+    private fun cleanUp(): Unit = runBlocking {
+        deleteQueueAndAllMessages(correctChecksumClient, correctChecksumTestQueueUrl)
+        deleteQueueAndAllMessages(wrongChecksumClient, wrongChecksumTestQueueUrl)
+        correctChecksumClient.close()
+        wrongChecksumClient.close()
+    }
+
+    @Test
+    fun testSendMessage(): Unit = runBlocking {
```
lauzadis, line 119: I don't see any assertions here, so what are we testing? Is there a way we can confirm that checksum validation occurred (similar to S3's flexible checksums)

services/sqs/e2eTest/src/SqsMd5ChecksumValidationTest.kt
```diff
@@ -0,0 +1,225 @@
...
+                        MessageSystemAttributeNameForSends.AwsTraceHeader to MessageSystemAttributeValue {
+                            dataType = "String"
+                            stringValue = TEST_MESSAGE_SYSTEM_ATTRIBUTES_VALUE
+                        },
+                    )
+                },
+            )
+        }
+
+        assert(exception.message!!.contains("Checksum mismatch"))
```
lauzadis, line 189: Asserting on exception messages is fragile, I think just `assertThrows<ChecksumMismatchException>` is enough

services/sqs/e2eTest/src/SqsTestUtils.kt
```diff
@@ -0,0 +1,100 @@
...
+    const val TEST_MESSAGE_BODY = "Hello World"
+    const val TEST_MESSAGE_ATTRIBUTES_NAME = "TestAttribute"
+    const val TEST_MESSAGE_ATTRIBUTES_VALUE = "TestAttributeValue"
+    const val TEST_MESSAGE_SYSTEM_ATTRIBUTES_VALUE = "TestSystemAttributeValue"
+
+    suspend fun getTestQueueUrl(
+        client: SqsClient,
+        prefix: String,
+        region: String? = null,
+    ): String = getQueueUrlWithPrefix(client, prefix, region)
```
lauzadis, line 32: question: did ktlint put all these parameters on new lines? I thought it could fit in one line...

services/sqs/e2eTest/src/SqsTestUtils.kt
```diff
@@ -0,0 +1,100 @@
...
+        region: String? = null,
+    ): String = withTimeout(60.seconds) {
+        var matchingQueueUrl = client
+            .listQueuesPaginated { queueNamePrefix = prefix }
+            .queueUrls()
+            .firstOrNull()
+
+        if (matchingQueueUrl == null) {
+            matchingQueueUrl = prefix + UUID.randomUUID()
+            println("Creating Sqs queue: $matchingQueueUrl")
```
lauzadis, line 46: capitalization / consistency: The service name is "SQS", not "Sqs"

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
xinsong-cui: I'm getting an error in the generated `DefaultSqsClient` (ine 123 of [SqsMd5ChecksumValidationIntegration.kt](https://github.com/awslabs/aws-sdk-kotlin/pull/1544/files#diff-4392ee759a506bee577009755cfaf986bd3ed0d91d512fcf2c6834661b8021da)) if I set it not nullable. 

Is there a workaround to handle this without making it nullable?

services/sqs/e2eTest/src/SqsTestUtils.kt
xinsong-cui: No it doesn't, I think that's because I refactored the parameters, should be able to fit in one line now.

lauzadis: Ok, don't worry about it too much, I was just curious how it was formatted this way

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
lauzadis: I think it's because the `ConfigProperty` that you created called `ValidationEnabledProp` is nullable. If you make it non-nullable with a default value of `false` then it will work

lauzadis:
```
            symbol = buildSymbol {
                name = "ValidationEnabled"
                namespace = "aws.sdk.kotlin.services.sqs.internal"
                nullable = false
                defaultValue = "false"
            }
```

xinsong-cui: Sorry I didn't include full context, `ValidationEnabled` should be non-nullable since it will go through reslution chain in [FinalizeSqsConfig](https://github.com/awslabs/aws-sdk-kotlin/pull/1544/files#diff-6997af09f81d7d78cd43a1b179d03116c5fb96592db62ae325d4fb265f774422)
But there's a type mismatch because`SqsMd5ChecksumValidationInterceptor` is using the the type defined in `ConfigProperty` 

If we setting a default value in the `ConfigProperty` will prevent further fallback resolution 

I was wondering if there is way to workaround this.

ianbotsf: OK I think we want to use a custom property type here:

```kotlin
val ValidationEnabledProp: ConfigProperty = ConfigProperty {
    name = "checksumValidationEnabled"
    symbol = buildSymbol {
        name = "ValidationEnabled"
        namespace = "aws.sdk.kotlin.services.sqs.internal"
        nullable = false
    }
    propertyType = ConfigPropertyType.Custom(
        render = { prop, writer ->
            writer.write("public val #1L: #2T = builder.#1L ?: #2T.NEVER", prop.propertyName, prop.symbol)
        },
        renderBuilder = { prop, writer ->
            writer.write("public var #L: #T? = null", prop.propertyName, prop.symbol)
        }
    )
    documentation = "..."
}
```

This allows us to have a nullable builder type and non-nullable built type which coalesces to the default value `NEVER`. Note that this coalescing runs _after_ `finalizeEnvironmentalConfig` (and therefore after `finalizeSqsConfig`) so the value will've been resolved from env config by then. This additional `?: NEVER` is only ever triggered when building a client _without_ using env config (e.g., `SqsClient { }`).

I thought we had an abstraction already for this but I was mistaken so this might be the best we can do for now.

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
ianbotsf, line 61: **Correctness**: This will log an error on every operation invocation when MD5 is not available, which is not what we want. Instead, we should log an error just once. One possible way to achieve this is to determine MD5 availability in the companion object:

```kotlin
companion object {
    private val isMd5Available = asyncLazy {
        try {
            "MD5".encodeToByteArray().md5()
            true
        } catch (e: NoSuchAlgorithmException) {
            coroutineContext.error(e) { "MD5 checksums are not available (likely due to FIPS mode). Checksum validation will be disabled." }
            false
        }
    }
}
```

And then to read the value in the hook:

```kotlin
override fun readAfterExecution(context: ResponseInterceptorContext<Any, Any, HttpRequest?, HttpResponse?>) {
    if (validationEnabled == ValidationEnabled.NEVER || !isMd5Available.get()) return
    ...
```

The `asyncLazy` will ensure the value is only ever calculated once, including the emission of the error message.

ianbotsf: Users can [swap out crypto implementations by configuring their JVM](https://docs.oracle.com/javase/8/docs/technotes/guides/security/crypto/CryptoSpec.html#ProviderInstalling). I haven't yet verified this but I believe that when the installed providers do not support MD5, then `MessageDigest.getInstance("MD5")` will throw a `NoSuchAlgorithmException`.

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,330 @@
...
+
+        val request = context.request
+        val response = context.response.getOrNull()
+
+        if (response != null) {
+            when (request) {
+                is SendMessageRequest -> {
+                    if (validationEnabled == ValidationEnabled.WHEN_RECEIVING) return
+
+                    val sendMessageRequest = request as SendMessageRequest
```
ianbotsf, line 71: **Nit**: Is this necessary? Doesn't the compiler smart cast `request` to `SendMessageRequest` since we're in the `is SendMessageRequest` branch?

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,330 @@
...
+            }
+        }
+    }
+
+    private fun calculateMessageBodyMd5(messageBody: String): String {
+        val expectedMD5 = messageBody.encodeToByteArray().md5()
+        val expectedMD5Hex = expectedMD5.toHexString()
+
+        return expectedMD5Hex
+    }
```
ianbotsf, line 245: **Nit**: Could be simplified:

```kotlin
private fun calculateMessageBodyMd5(messageBody: String) =
    messageBody.encodeToByteArray().md5().toHexString()
```

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,330 @@
...
+            .sortedBy { (name, _) -> name }
+            .forEach { (attributeName, attributeValue) ->
+                updateLengthAndBytes(buffer, attributeName)
+                updateLengthAndBytes(buffer, attributeValue.dataType)
+                when {
+                    attributeValue.stringValue != null -> updateForStringType(buffer, attributeValue.stringValue)
+                    attributeValue.binaryValue != null -> updateForBinaryType(buffer, attributeValue.binaryValue)
+                    !attributeValue.stringListValues.isNullOrEmpty() -> updateForStringListType(buffer, attributeValue.stringListValues)
+                    !attributeValue.binaryListValues.isNullOrEmpty() -> updateForBinaryListType(buffer, attributeValue.binaryListValues)
+                }
```
ianbotsf, line 265: **Nit**: We should add an `else` case here to handle the unlikely situation where none of the values are set:

```kotlin
else -> throw ClientException("No value type found for attribute $attributeName")
```

services/sqs/common/src/aws.sdk.kotlin.services.sqs/internal/ValidationConfig.kt
ianbotsf, line 18: **Nit**: Monospace the enum value names

services/sqs/common/src/aws.sdk.kotlin.services.sqs/internal/ValidationConfig.kt
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * This configuration determines which components of a message will be validated
+ * when checksum validation is enabled.
+ *
+ * Valid values:
+ * - MESSAGE_ATTRIBUTES - Validates checksums for message attributes
+ * - MESSAGE_SYSTEM_ATTRIBUTES - Validates checksums for message system attributes
+ *   (Note: Not available for ReceiveMessage operations as SQS does not calculate
+ *   checksums for system attributes during message receipt)
+ * - MESSAGE_BODY - Validates checksums for the message body
```
ianbotsf, line 38: **Nit**: Monospace the enum value names

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/sqs/SqsMd5ChecksumValidationIntegration.kt
```diff
@@ -0,0 +1,125 @@
...
+ * Register interceptor to handle SQS message MD5 checksum validation.
+ */
+class SqsMd5ChecksumValidationIntegration : KotlinIntegration {
+    override fun enabledForService(model: Model, settings: KotlinSettings): Boolean =
+        model.expectShape<ServiceShape>(settings.service).isSqs
+
+    companion object {
+        val ValidationEnabledProp: ConfigProperty = ConfigProperty {
+            name = "checksumValidationEnabled"
+            symbol = buildSymbol {
```
0marperez, line 34: Question: Are we setting the default value like in `ValidationScopeProp`?

services/sqs/e2eTest/src/SqsMd5ChecksumValidationTest.kt
0marperez: I agree with this, +1

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,330 @@
...
+            "MD5".encodeToByteArray().md5()
+        } catch (e: Exception) {
+            logger.error { "MD5 checksums are not available (likely due to FIPS mode). Checksum validation will be disabled." }
+            return
+        }
+
+        val request = context.request
+        val response = context.response.getOrNull()
+
+        if (response != null) {
```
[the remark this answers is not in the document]
xinsong-cui: I don't know why I cannot reply to your previous comment.. I will reply here instead: https://github.com/awslabs/aws-sdk-kotlin/pull/1544#discussion_r1987897949
Yeah it is possible to have empty `messageBodySent`, but I don't think it can be null

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/sqs/SqsMd5ChecksumValidationIntegration.kt
xinsong-cui: Yes, I have't update my code yet but it will be like Ian suggested: ttps://github.com/awslabs/aws-sdk-kotlin/pull/1544#discussion_r1983743776

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/sqs/SqsMd5ChecksumValidationIntegration.kt
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        private val validationScope = buildSymbol {
+            name = "ValidationScope"
+            namespace = "aws.sdk.kotlin.services.sqs.internal"
+        }
+
+        val ValidationScopeProp: ConfigProperty = ConfigProperty {
+            name = "checksumValidationScopes"
+            symbol = KotlinTypes.Collections.set(validationScope, default = "emptySet()")
```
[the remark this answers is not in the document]
xinsong-cui: Yes, default will be set to 3 scopes similar to `ValidationEnabled`
reference: https://github.com/awslabs/aws-sdk-kotlin/pull/1544#discussion_r1985744580

services/sqs/e2eTest/src/SqsTestUtils.kt
```diff
@@ -0,0 +1,92 @@
...
+
+            client.purgeQueue(
+                PurgeQueueRequest {
+                    this.queueUrl = queueUrl
+                },
+            )
+
+            println("Queue purged successfully.")
+
+            println("Deleting Sqs queue: $queueUrl")
```
lauzadis, line 62: capitalize SQS here and throughout the PR

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,342 @@
...
+                val messageBodyReturned = messageReceived.body
+
+                if (!messageBodyMd5Returned.isNullOrEmpty() && !messageBodyReturned.isNullOrEmpty()) {
+                    logger.debug { "Validating message body MD5 checksum for ReceiveMessage" }
+
+                    val clientSideBodyMd5 = calculateMessageBodyMd5(messageBodyReturned)
+
+                    validateMd5(clientSideBodyMd5, messageBodyMd5Returned)
+
+                    logger.debug { "Message body MD5 checksum for ReceiveMessage validated " }
```
lauzadis, line 169: nit: unnecessary trailing space here and below

services/sqs/common/src/aws.sdk.kotlin.services.sqs/SqsMd5ChecksumValidationInterceptor.kt
```diff
@@ -0,0 +1,342 @@
...
+                "MD5".encodeToByteArray().md5()
+                true
+            } catch (e: Exception) {
+                coroutineContext.error<SqsMd5ChecksumValidationInterceptor>(e) {
+                    "MD5 checksums are not available (likely due to FIPS mode). Checksum validation will be disabled."
+                }
+                false
+            }
+        }
+    }
```
lauzadis, line 66: Why does this need to be `async` or `lazy`? Can we just add an `init { ... }` block to this interceptor that checks if MD5 is enabled at initialization time?

lauzadis: Ah ok, we need to use `companion object` because that way it'll only ever be calculated once. `init { ... }` will calculated it every time the interceptor is created. I think this is fine

0marperez (approved):

PR #1554, opened 2025-03-28
feat: sync mirror

0marperez:
## Issue \#

N/A

## Description of changes

Syncs mirror's main branch on pushes to main

### DIFF
--- a/.github/workflows/sync-mirror.yml
+++ b/.github/workflows/sync-mirror.yml
@@ -7,13 +7,14 @@ on:
 
 jobs:
   git-sync:
+    # Only sync when pushing to source repo
     if: github.repository == 'awslabs/aws-sdk-kotlin'
     runs-on: ubuntu-latest
     steps:
       - name: git-sync
         uses: wei/git-sync@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-mirror.yml
```diff
@@ -0,0 +1,19 @@
+name: Sync Mirror
+
+on:
+  push:
+    branches: [ main ]
+  workflow_dispatch:
+
+jobs:
+  git-sync:
+    if: github.repository == 'awslabs/aws-sdk-kotlin'
```
lauzadis, line 10: In what situations would this check fail? Why is it necessary?

0marperez: The Swift team [has it](https://github.com/awslabs/aws-sdk-swift/blob/6769aaea818cd4e06cdd52887aa74804080213ff/.github/workflows/repo-sync.yml#L10) and I'm pretty sure it's so that a push to the destination repo doesn't trigger a sync. We want pushes to the source repo to trigger a sync but not pushes to the destination repo.

lauzadis (approved):

ianbotsf (approved):

PR #1560, opened 2025-04-03
feat: release readiness check

0marperez:
## Description of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## [1.4.56] - 04/04/2025
+
+### Features
+* (**eventbridge**) Amazon EventBridge adds support for customer-managed keys on Archives and validations for two fields: eventSourceArn and kmsKeyIdentifier.
+
+### Documentation
+* (**directoryservicedata**) Doc only update - fixed broken links.
+* (**ec2**) Doc-only updates for Amazon EC2
+* (**s3control**) Updated max size of Prefixes parameter of Scope data type.
+
 ## [1.4.55] - 04/03/2025
 
 ### Features
--- a/codegen/aws-sdk-codegen/src/main/resources/aws/sdk/kotlin/codegen/endpoints.json
+++ b/codegen/aws-sdk-codegen/src/main/resources/aws/sdk/kotlin/codegen/endpoints.json
@@ -15321,6 +15321,7 @@
           "ap-southeast-3" : { },
           "ap-southeast-4" : { },
           "ap-southeast-5" : { },
+          "ap-southeast-7" : { },
           "ca-central-1" : { },
           "ca-west-1" : { },
           "eu-central-1" : { },
@@ -15362,6 +15363,7 @@
           "il-central-1" : { },
           "me-central-1" : { },
           "me-south-1" : { },
+          "mx-central-1" : { },
           "sa-east-1" : { },
           "us-east-1" : {
             "variants" : [ {
--- a/codegen/sdk/aws-models/directory-service-data.json
+++ b/codegen/sdk/aws-models/directory-service-data.json
@@ -76,6 +76,20 @@
             ],
             "traits": {
                 "smithy.api#documentation": "<p>Adds an existing user, group, or computer as a group member.</p>",
+                "smithy.api#examples": [
+                    {
+                        "title": "To add a member to the Marketing group",
+                        "documentation": "The following command adds an existing user to the Marketing group in the europe.example.com domain.",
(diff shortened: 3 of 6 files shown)


.github/workflows/release-readiness.yml
```diff
@@ -0,0 +1,30 @@
+name: Release readiness (snapshot dependency)
+
+# Makes sure that we aren't relying on SNAPSHOT/dev versions of smithy-kotlin before merging
+# Otherwise we could forget because the rest of CI is masking it
+
+on:
+  pull_request:
+    branches: [ main ]
+  workflow_dispatch:
```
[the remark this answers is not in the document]
ianbotsf: It allows the workflow [to be triggered manually](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch). I don't know if we'll necessarily need to but seems fine to me.

.github/workflows/release-readiness.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Configure error message
+        run: echo "ERROR_MESSAGE=WARNING smithy-kotlin release and version bump might be required before merging!" >> $GITHUB_ENV
+
+      - name: Build SDK
+        run: ./gradlew test jvmTest || echo $ERROR_MESSAGE
+
+      - name: Build SDK client
+        run: |
+          ./gradlew -Paws.services=s3 -Paws.kotlin.native=false bootstrap || echo $ERROR_MESSAGE
+          ./gradlew build || echo $ERROR_MESSAGE
```
ianbotsf, line 30: **Question**: How does this work in the failure case?

.github/workflows/release-readiness.yml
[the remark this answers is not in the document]
lauzadis: Yeah, that's where I was leading with the question. I don't think we'll ever need to trigger this workflow manually, so we could simplify this a little by removing it?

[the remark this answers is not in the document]
0marperez: Sure, we can remove it and if we ever need to run it manually we can add the `workflow_dispatch` back

.github/workflows/release-readiness.yml
0marperez: If the command exits with a non-zero exit code then we'll echo the message at the end of the output

lauzadis: But echo returns a zero exit code, so does the CI properly report a failed status?

0marperez: Oh I see what you mean, it wouldn't. Fixed

.github/workflows/release-readiness.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Configure error message
+        run: echo "ERROR_MESSAGE=WARNING smithy-kotlin release and version bump might be required before merging" >> $GITHUB_ENV
+
+      - name: Build SDK
+        run: ./gradlew test jvmTest || { echo "$ERROR_MESSAGE"; exit 1; }
+
+      - name: Build SDK client
+        run: |
+          ./gradlew -Paws.services=s3 -Paws.kotlin.native=false bootstrap || { echo "$ERROR_MESSAGE"; exit 1; }
+          ./gradlew -Paws.kotlin.native=false build || { echo "$ERROR_MESSAGE"; exit 1; }
```
ianbotsf, line 29: **Nit**: I think this is still more complicated than it needs to be. A step can be [configured to run on failure](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif) so that a [GitHub-standard error message](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-error-message) can be emitted:

```yaml

jobs:
  release-readiness:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout sources
        uses: actions/checkout@v4

      - name: Configure Gradle
        uses: awslabs/aws-kotlin-repo-tools/.github/actions/configure-gradle@main

      - name: Build SDK
        run: ./gradlew test jvmTest

      - name: Build SDK client
        run: |
          ./gradlew -Paws.services=s3 -Paws.kotlin.native=false bootstrap
          ./gradlew -Paws.kotlin.native=false build

      - name: Emit error message
        if: ${{ failure() }}
        run: |
          echo "::error ::Build failed. Did you forget to release smithy-kotlin and bump the dependency version?"
          exit 1
```

PR #1582, opened 2025-04-22
feat: Add `authSchemePreference` configuration

lauzadis:
## Description of changes


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/auth/ResolveAuthSchemePreference.kt
```diff
@@ -0,0 +1,36 @@
...
+        ?.mapNotNull { AUTH_SCHEME_NAME_MAP[it.lowercase()] }
+        ?.toList()
+        ?: emptyList()
+}
+
+/**
+ * Map of lowercase auth scheme _name_ (not namespace) to auth scheme ID
+ */
+private val AUTH_SCHEME_NAME_MAP = mapOf<String, AuthSchemeId>(
+    "sigv4" to AuthSchemeId.AwsSigV4,
```
0marperez, line 29: Suggestion: I like the approach to see if an item is an auth scheme or not. Do you think we could populate the map automatically somehow? If a new auth scheme is added we might forget to include it in this map.

0marperez (approved):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/auth/ResolveAuthSchemePreference.kt
lauzadis: Do you have any suggestions? I agree with updating it automatically and tried a few things but wasn't happy with any solution. 
(1) Use reflection to generate the map at runtime 
(2) Add `AuthSchemeId.fromName` extension function, move this map to smithy-kotlin / `AuthSchemeId`. We'll still need to manually update the map but it'll be in the same file as the auth schemes. This doesn't work for AWS-specific auth schemes like SigV4S3Express

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/auth/ResolveAuthSchemePreference.kt
```diff
@@ -0,0 +1,36 @@
...
+    val content = AwsSdkSetting.AwsAuthSchemePreference.resolve(platform) ?: profile.get().authSchemePreference
+
+    return content
+        ?.split(",")
+        ?.map { it.trim() }
+        ?.filter { it.isNotEmpty() }
+        ?.mapNotNull { AUTH_SCHEME_NAME_MAP[it.lowercase()] }
+        ?.toList()
+        ?: emptyList()
+}
```
ianbotsf, line 23: **Correctness**: This cannot resolve any auth scheme IDs which come from plugins, integrations, or other sources, which seems incorrect. Moreover I'm not sure it's _necessary_ to reuse references to the existing instances of `AuthSchemeId` since it's a value class wrapper around the string ID.

I think this can be simplified to:

```kotlin
return content
    ?.split(",")
    ?.map { it.trim() }
    ?.filter { it.isNotEmpty() }
    ?.map(::AuthSchemeId)
    ?: emptyList()
```

ianbotsf (changes requested):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/auth/ResolveAuthSchemePreference.kt
lauzadis: `.map(::AuthSchemeId)` works but it's a lie, since we aren't providing the full ID (with namespace), just the name. It should be fine for this `authSchemePreference` usecase

ianbotsf (approved):

PR #1583, opened 2025-04-22
feat: add support for DDB Mapper `getItem` overloads that specify primary key(s)

### LINKED ISSUE #1577: DynamoDbMapper getItem not working by providing key as String
### Describe the bug

Using `getItem(partitionKey: PK)` where the `partitionKey` is a String in the data class which generates the schema produces the following error:
```kotlin
kotlin.NotImplementedError: An operation is not implemented: not yet implemented
	at aws.sdk.kotlin.hll.dynamodbmapper.model.internal.TableImplKt$tableImpl$1.getItem(TableImpl.kt:38) ~[dynamodb-mapper-jvm-1.4.58-beta.jar:na]
```

Trying to use `getItem { key = "a-string-key" }` following the example on the [blog](https://aws.amazon.com/blogs/developer/announcing-dev-preview-of-dynamodb-mapper-for-kotlin/), it said the type of the `key` required is the class which generates the schemas, instead of String.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected behavior

- Calling getItem() without any error. 

- The type of the `key` in `getItem` builder block should be the type of `DynamoDbPartitionKey` defined in the class.

### Current behavior

- Calling getItem() throws NotImplementedError

- `key` in `getItem` builder block only accepts the class which generate the schemas, instead of the type of `DynamoDbPartitionKey` defined in the class.

### Steps to Reproduce

Using Java 23 and Kotlin 2.1.10

Data class to generate Schema
```kotlin
@DynamoDbItem
data class Planet(
    @DynamoDbPartitionKey
    val name: String,
    val id: Long
)
```

Use `getItem()`, calling the getPlanet() with a Spring boot controller gives the error
```kotlin
    suspend fun getPlanet(): Planet? {
        val planetTable = mapper.getTable("planet-table", PlanetSchema)
        return planetTable.getItem("Jupiter")
    }
```

Use `getItem { }`  
```kotlin
    suspend fun getPlanet(): Planet? {
        val planetTable = mapper.getTable("planet-table", PlanetSchema)
        return planetTable.getItem { key = "Jupiter" }.item
    }
```

### Possible Solution

_No response_

### Context

_No response_

### AWS SDK for Kotlin version

1.4.58

### Platform (JVM/JS/Native)

JVM

### Operating system and version

macOS

ianbotsf:
## Issue \#

Resolves #1577 

## Description of changes

Adds support for `getItem(partitionKey: PK)` and `getItem(partitionKey: PK, sortKey: SK)` in DynamoDB Mapper tables by way of a new private interceptor that sets the low-level key before invocation.

This change also updates the `Operation` type to a `data class` to provide an easy `copy` method which can insert the new interceptor into an operation created by one of the codegen factory methods. The `Operation` type is `internal` so there's no concern about public API compatibility.

### DIFF
--- a/hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/Operation.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/Operation.kt
@@ -5,20 +5,33 @@
 package aws.sdk.kotlin.hll.dynamodbmapper.pipeline.internal
 
 import aws.sdk.kotlin.hll.dynamodbmapper.items.ItemSchema
-import aws.sdk.kotlin.hll.dynamodbmapper.pipeline.*
+import aws.sdk.kotlin.hll.dynamodbmapper.pipeline.Interceptor
+import aws.sdk.kotlin.hll.dynamodbmapper.pipeline.InterceptorAny
 
-internal class Operation<T, HReq, LReq, LRes, HRes>(
-    private val initialize: (HReq) -> HReqContextImpl<T, HReq>,
-    private val serialize: (HReq, ItemSchema<T>) -> LReq,
-    private val lowLevelInvoke: suspend (LReq) -> LRes,
-    private val deserialize: (LRes, ItemSchema<T>) -> HRes,
-    interceptors: Collection<InterceptorAny>,
+internal data class Operation<T, HReq, LReq, LRes, HRes>(
+    val initialize: (HReq) -> HReqContextImpl<T, HReq>,
+    val serialize: (HReq, ItemSchema<T>) -> LReq,
+    val lowLevelInvoke: suspend (LReq) -> LRes,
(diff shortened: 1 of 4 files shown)


lauzadis (approved):

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/Operation.kt
```diff
@@ -5,20 +5,33 @@
...
+        lowLevelInvoke: suspend (LReq) -> LRes,
+        deserialize: (LRes, ItemSchema<T>) -> HRes,
+        interceptors: Collection<InterceptorAny>,
+    ) : this(
+        initialize,
+        serialize,
+        lowLevelInvoke,
+        deserialize,
+        interceptors.map {
+            // Will cause runtime ClassCastExceptions during interceptor invocation if the types don't match. Is that ok?
```
0marperez, line 30: Nit: The comment is useful but `Is that ok?` seems like it could be removed

0marperez (approved):

PR #1385, opened 2024-08-21
feat: add Schema generator Gradle plugin

lauzadis:
## Description of changes

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1585, opened 2025-04-24
fix: use custom Gradle distribution in schema generator tests

lauzadis:
## Description of changes

### DIFF
--- a/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
@@ -56,7 +56,6 @@ class SchemaGeneratorPluginTest {
             .create()
             .withProjectDir(testProjectDir)
             .withPluginClasspath()
-            .withGradleDistribution("https://d2pjps8lqszrgq.cloudfront.net/gradle-8.12.1-bin.zip") // TODO Parameterize Gradle versions
             .forwardOutput()
             .withArguments("--info", "build")
     }


hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
```diff
@@ -56,7 +56,7 @@ class SchemaGeneratorPluginTest {
             .create()
             .withProjectDir(testProjectDir)
             .withPluginClasspath()
-            .withGradleVersion("8.5") // TODO parameterize
+            .withGradleDistribution("https://d2pjps8lqszrgq.cloudfront.net/gradle-8.12.1-bin.zip") // TODO Parameterize Gradle versions
```
ianbotsf, line 59: **Question**: Do we _need_ to parameterize the Gradle version or do we really want to match the Gradle version already running? Because our CI [updates **gradle/wrapper/gradle-wrapper.properties** with the distribution URL](https://github.com/awslabs/aws-kotlin-repo-tools/blob/e263cc4dab8f64b08cc57257497a1b035f1a150d/.github/actions/configure-gradle/action.yml#L29-L32). It seems like we could reuse that for this runner configuration:

```kotlin
private fun detectGradleWrapperDistro(): URI {
    val propertiesFile = File("../../../gradle/wrapper/gradle-wrapper.properties")

    val properties = propertiesFile
        .readLines()
        .map { it.trim() }
        .filterNot { it.isBlank() || it.startsWith("#") }
        .associate { it.split("=", limit = 2).run { get(0).trim() to getOrNull(1)?.trim() } }

    return URI(requireNotNull(properties["distributionUrl"]))
}

...

runner = GradleRunner
    .create()
    .withProjectDir(testProjectDir)
    .withPluginClasspath()
    .withGradleDistribution(detectGradleWrapperDistro())
    .forwardOutput()
    .withArguments("--info", "build")
```

ianbotsf (approved):

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
lauzadis: I think you originally gave me feedback that we should test across multiple Gradle versions to ensure compatibility. I'm not too concerned about that, but I think synchronizing with `gradle-wrapper.properties` would be good

lauzadis: Nevermind, you gave the same feedback back then that you just did now 😆 
https://github.com/awslabs/aws-sdk-kotlin/pull/1385#discussion_r1727204188

I'll try to update this to pull the version from `gradle-wrapper.properties`

0marperez: What happens if we don't use `withGradleVersion` or `withGradleDistribution`. According to the Gradle runner [docs](https://docs.gradle.org/current/userguide/test_kit.html#sub:gradle-runner-gradle-version):

> By default, the runner will attempt to find a Gradle distribution based on where the GradleRunner class was loaded from

lauzadis: Nice suggestion, it seems to work locally, let's try it in CI...

0marperez (approved):

PR #1586, opened 2025-04-25
feat: release readiness - smithy-kotlin branch check

0marperez:
## Description of changes

### DIFF
--- a/.github/workflows/release-readiness.yml
+++ b/.github/workflows/release-readiness.yml
@@ -45,7 +45,7 @@ jobs:
           SMITHY_KOTLIN_BRANCH=$(git branch --show-current)
           echo "smithy-kotlin branch: $SMITHY_KOTLIN_BRANCH"
           
-          if [ "$SDK_BRANCH" == "$SMITHY_KOTLIN_BRANCH" ]; then
+          if [ "$GITHUB_HEAD_REF" == "$SMITHY_KOTLIN_BRANCH" ]; then
             cd ../aws-sdk-kotlin
             git fetch origin
 


.github/workflows/release-readiness.yml
```diff
@@ -11,22 +11,61 @@ jobs:
...
+            git fetch origin
+
+            DIFF=$(git diff origin/main -- gradle/libs.versions.toml | grep '^[-+][^-+]'; exit 0)
+            SMITHY_KOTLIN_VERSION_BUMP=$(echo "$DIFF" | grep "smithy-kotlin-runtime-version =\|smithy-kotlin-codegen-version ="; exit 0)
+
+            if [ -z "$SMITHY_KOTLIN_VERSION_BUMP" ]; then
+              echo "::error::Matching smithy-kotlin and aws-sdk-kotlin branches but no smithy-kotlin version bump"
+              exit 1
+            else
+              echo "::warning::Matching smithy-kotlin and aws-sdk-kotlin branches with smithy-kotlin version bump detected"
```
lauzadis, line 63: Why is this a warning?

.github/workflows/release-readiness.yml
```diff
@@ -11,22 +11,61 @@ jobs:
   release-readiness:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout sources
+      - name: Checkout source
         uses: actions/checkout@v4
+        with:
+          path: 'aws-sdk-kotlin'
+          ref: '${{ github.head_ref }}'
```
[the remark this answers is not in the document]
0marperez: I think so, when I was testing and making changes before, `git branch --show-current` was failing because the default behavior is to checkout a commit _only_ not the whole branch.

.github/workflows/release-readiness.yml
```diff
@@ -11,22 +11,61 @@ jobs:
...
+          fetch-depth: 0
```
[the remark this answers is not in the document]
0marperez: I think we can remove this, I don't think it matters. This is left over from testing. I'll remove it and see if the workflow still works.

.github/workflows/release-readiness.yml
0marperez: I think there's a small chance someone makes a change to the smithy version lines that's not a version bump, and a warning could help someone debug.

.github/workflows/release-readiness.yml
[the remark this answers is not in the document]
lauzadis: Instead of using `git branch --show-current` can't you just use the `GITHUB_HEAD_REF` environment variable that's set by default? https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

.github/workflows/release-readiness.yml
lauzadis: Not a strong opinion but I think if there's matching smithy-kotlin and aws-sdk-kotlin branches, and we do a version bump, that's expected behavior and shouldn't be a warning?

ianbotsf: Agreed, this should be info.

lauzadis (approved):

PR #1587, opened 2025-04-29
feat: add analytics to api ref docs

xinsong-cui:
## Description of changes

### DIFF
--- a/dokka-aws/src/test/kotlin/aws/sdk/kotlin/dokka/DokkaAwsTest.kt
+++ b/dokka-aws/src/test/kotlin/aws/sdk/kotlin/dokka/DokkaAwsTest.kt
@@ -5,7 +5,6 @@
 package aws.sdk.kotlin.dokka
 
 import org.jsoup.Jsoup
-import org.junit.jupiter.api.Assumptions.assumeTrue
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.TestInstance
 import java.io.File
@@ -19,7 +18,7 @@ class DokkaAwsTest {
 
         assertTrue(
             testFile.exists(),
-            "Test file does not exist: ${testFile.absolutePath}"
+            "Test file does not exist: ${testFile.absolutePath}",
         )
 
         val document = Jsoup.parse(testFile, "UTF-8")


docs/dokka-presets/templates/base.ftl
```diff
@@ -7,6 +7,8 @@
 <head>
     <meta name="viewport" content="width=device-width, initial-scale=1" charset="UTF-8">
     <@page_metadata.display/>
+    <#-- AWS Analytics tracking script - Implements site-wide analytics and tracking functionality -->
+    <script src="https://a0.awsstatic.com/s_code/js/3.0/awshome_s_code.js"></script>
```
lauzadis, line 11: Instead of editing the base template we should add the scripts using `customAssets`/ `customStyleSheets`: 

 https://github.com/awslabs/aws-sdk-kotlin/blob/0dc1f013b0b732222a4f853203659dc7bcbe344c/build.gradle.kts#L50-L58

xinsong-cui: Is there a reason not adding them into base template? 
AWS Analytics needs to be [loaded first before other script](https://w.amazon.com/bin/view/AWS_Marketing/DSE/MetricsEngineering/AnalyticsSCode/ImplementSCode#:~:text=Make%20sure%20that%20the%20script%20load%C2%A0before%20all%20other%20JS%20scripts.%C2%A0)
and ShortBread needs to be placed [before closing \<head\> tags](https://quip-amazon.com/xatbA1SwKdap/Getting-Started-with-Shortbread-CDN)

I tried to add them into `customAssets`/ `customStyleSheets` but it seems I cannot control the loading order.

xinsong-cui: I was wrong earlier, adding to `customAssets` or `customStyleSheets` won’t work, they only accept local files. From what I’ve found, the only way to integrate a CDN link is by modifying the base template.

xinsong-cui: I was wrong earlier, adding to `customAssets` or `customStyleSheets` won’t work, they only accept local files. From what I’ve found, the only way to integrate a CDN link is by modifying the base template.

dokka-aws/src/test/kotlin/aws/sdk/kotlin/dokka/DokkaAwsTest.kt
```diff
@@ -0,0 +1,40 @@
...
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class DokkaAwsTest {
+    @Test
+    fun testLoadScripts() {
+        val testFile = File("build/dokka/html/index.html")
+
+        assumeTrue(
+            testFile.exists(),
+            "Skipping test: test file does not exist",
+        )
```
ianbotsf, line 23: **Nit**: We should not skip the test if the test file doesn't exist. We should fail (i.e., `assertTrue`).

gradle/libs.versions.toml
```diff
@@ -26,6 +26,7 @@ kotlinx-benchmark-version = "0.4.12"
 kotlinx-serialization-version = "1.7.3"
 mockk-version = "1.13.13"
 slf4j-version = "2.0.16"
+jsoup-version = "1.19.1"
```
ianbotsf, line 29: **Nit**: The latest release is [**1.20.1**](https://github.com/jhy/jsoup/releases/tag/jsoup-1.20.1)

ianbotsf (approved):
Minor feedback, fix and ship!

PR #1592, opened 2025-05-08
kn: properly memscope `executeCommand`

lauzadis:
Fixes a new error seen when running `ExecuteCommandTest` on macosArm64 (and likely other platforms too). We weren't properly memscoping the use of `command`, and it was getting corrupted:
```
/bin/sh: �@p-: No such file or directory
```

Also fully qualify the shell path because iOS simulator does not have `sh` on the PATH.

## Description of changes

### DIFF
--- a/aws-runtime/aws-config/native/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommandNative.kt
+++ b/aws-runtime/aws-config/native/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommandNative.kt
@@ -48,8 +48,8 @@ internal actual suspend fun executeCommand(
         }
 
         val shell = when (platformProvider.osInfo().family) {
-            OsFamily.Windows -> "cmd.exe"
-            else -> "sh"
+            OsFamily.Windows -> "C:\\\\Windows\\\\System32\\\\cmd.exe"
+            else -> "/bin/sh"
         }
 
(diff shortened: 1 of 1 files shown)


aws-runtime/aws-config/native/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommandNative.kt
```diff
@@ -48,17 +48,19 @@ internal actual suspend fun executeCommand(
         }
 
         val shell = when (platformProvider.osInfo().family) {
-            OsFamily.Windows -> "cmd.exe"
-            else -> "sh"
+            OsFamily.Windows -> "C:\\\\Windows\\\\System32\\\\cmd.exe"
```
ianbotsf, line 51: **Question**: Why are these backslashes _double_ escaped?

ianbotsf (approved):

aws-runtime/aws-config/native/src/aws/sdk/kotlin/runtime/auth/credentials/executeCommandNative.kt
lauzadis: Cancelling in favor of #1593. I don't think they should be double escaped.

PR #1598, opened 2025-05-16
misc: only run release branch workflow on public repo

0marperez:
## Issue \#

N/A

## Description of changes

### DIFF
--- a/.github/workflows/update-release-branch.yml
+++ b/.github/workflows/update-release-branch.yml
@@ -29,6 +29,8 @@ env:
 
 jobs:
   update-release:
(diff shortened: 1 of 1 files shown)


lauzadis:
Should we just create a **release** branch in the mirrored repository?

0marperez:
@lauzadis We can but it's not going to be used. It would only be there to stop the workflow from failing

lauzadis:
That sounds fine, it's better than adding extra checks into our CI scripts in my opinion. The mirror should be an exact mirror, even if we don't use the branch

PR #1599, opened 2025-05-16
feat: api ref doc sidebar optimization

xinsong-cui:
## Description of changes


build.gradle.kts
```diff
@@ -119,6 +122,83 @@ allprojects {
...
+            dokkaOutputDir.walkTopDown()
+                .filter { it.isFile && it.name.endsWith(".html") }
+                .forEach { file ->
+                    val updatedContent = file.readLines().filterNot { line ->
+                        line.contains("""scripts/navigation-loader.js""")
+                    }.joinToString("\n")
+
+                    file.writeText(updatedContent)
+                }
+        }
```
lauzadis, line 199: Instead of replacing the HTML content of every file, could we just create our own file **navigation-loader.js** and add it as a `customAsset`? Would that override Dokka's default navigation loader script? It seems like a simpler solution

xinsong-cui: No I don't think it will override Dokka's loader. The file from `customAsset` will be put into `images` folder like `images/navigation-loader.js`
The Dokka's own navigation-loader is in `scripts` folders. So both files will be in the generated html.

build.gradle.kts
```diff
@@ -119,6 +122,83 @@ allprojects {
...
+            }
+
+            dokkaOutputDir.listFiles { file ->
+                file.isDirectory && file.resolve("navigation.html").exists()
+            }?.forEach { moduleDir ->
+                val moduleName = moduleDir.name
+
+                val navFile = File(moduleDir, "navigation.html")
+
+                if (navFile.exists()) {
```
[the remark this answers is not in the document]
xinsong-cui: Oh yes, you are right, let me remove that

build.gradle.kts
```diff
@@ -119,6 +122,83 @@ allprojects {
...
+
+            if (!dokkaOutputDir.exists()) {
+                logger.warn("Dokka output directory not found: ${dokkaOutputDir.absolutePath}")
+                logger.warn("Skipping navigation trimming")
+                return@doLast
+            }
+
+            dokkaOutputDir.listFiles { file ->
+                file.isDirectory && file.resolve("navigation.html").exists()
+            }?.forEach { moduleDir ->
```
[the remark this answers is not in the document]
xinsong-cui: Do we still want to prune them? Now we are trimming and using these navigations

build.gradle.kts
lauzadis: Oh ok, I see that here: https://github.com/Kotlin/dokka/blob/540fe084aadac0edb333350210ee22ee9901a3de/dokka-subprojects/plugin-base/src/main/kotlin/org/jetbrains/dokka/base/renderers/html/htmlPreprocessors.kt#L37

It sounds like a good feature request to Dokka if you can open one, but we can work around it for this PR

build.gradle.kts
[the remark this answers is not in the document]
lauzadis: I didn't realize your navigation loader would use each module's navigation.html. Ok, that will work but we'll need to update our build process to _not_ remove duplicate files (and make sure the docs artifact size doesn't increase significantly)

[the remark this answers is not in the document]
xinsong-cui: Yep, I'll remove that. I did a quick estimation, the doc size should only increased by 300 - 400 MB

build.gradle.kts
xinsong-cui: Sounds good, I can open a request to Dokka

docs/dokka-presets/scripts/custom-navigation-loader.js
```diff
@@ -0,0 +1,152 @@
...
+
+    // Hide sidebar for root index page
+    if (moduleName === "index.html") {
+        hideSidebar()
+        return Promise.resolve('');
+    }
+
+    const navigationPath = moduleName
+        ? `${pathToRoot}${moduleName}/navigation.html`
+        : `${pathToRoot}navigation.html`;
```
0marperez, line 48: Should this be `${pathToRoot}/navigation.html` ?

buildSrc/src/main/kotlin/dokka-customization.gradle.kts
```diff
@@ -0,0 +1,74 @@
...
+                val trimmedHref = originalHref.replace("../", "")
+                anchor.attr("href", trimmedHref)
+            }
+
+            val sideMenuParts = doc.select("div.sideMenu > div.sideMenuPart")
+
+            sideMenuParts.forEach { submenu ->
+                val submenuId = submenu.id()
+                // If this is not the current module's submenu, remove all its nested content
+                if (submenuId != "$moduleName-nav-submenu") {
```
0marperez, line 36: Nit: We could get rid of `submenuId`:

 `if (submenu.id() != "$moduleName-nav-submenu") {`

buildSrc/src/main/kotlin/dokka-customization.gradle.kts
```diff
@@ -0,0 +1,74 @@
...
+                    val overviewDiv = submenu.select("> div.overview").first()
+                    overviewDiv?.select("span.navButton")?.remove()
```
0marperez, line 38: Are we going to have a way for users to go back a module?

0marperez (commented):
Should we do this for smithy-kotlin as well? The navigation sidebar isn't troublesome but for consistency it would be nice to do the same.

xinsong-cui:
> Should we do this for smithy-kotlin as well? The navigation sidebar isn't troublesome but for consistency it would be nice to do the same.
Yeah, I agree we should keep consistency. I am open to do for smithy-kotlin.

docs/dokka-presets/scripts/custom-navigation-loader.js
xinsong-cui: I think`/` was included in `pathToRoot`

buildSrc/src/main/kotlin/dokka-customization.gradle.kts
xinsong-cui: Yeah, users can go back a module pretty much like before (via navigation or the breadcrumb). The only difference is users can not jump in to another module's submenu directly.

buildSrc/build.gradle.kts
```diff
@@ -5,3 +5,7 @@ plugins {
 repositories {
     mavenCentral()
 }
+
+dependencies {
+    implementation(libs.jsoup)
+}
```
ianbotsf, line 11: **Question**: Why does this dependency need to be stated in two places (here and in **\<root>/build.gradle.kts**)?

xinsong-cui: The one in **root/build.gradle.kts** looks like leftover code from refactoring. I'll clean that up.

PR #1619, opened 2025-06-20
misc: run release readiness check when labled and more

0marperez:
## Issue \#

-

## Description of changes

Currently the workflow doesn't run again when the `ready-for-release` label is added. This makes it run when a label is added, removed, and more. Removing the need to re-run the workflow manually after adding the label


lauzadis (approved):
Nice

PR #1622, opened 2025-06-21
docs: fix navigation links on left sidebar

lauzadis:
Fixes two issues with our latest docs' left sidebar:
1) `extractModulePath` used the presence of periods to detect the module name, which failed for versioned docs, which use periods in the version number... simplified function implementation to fix this
2) Broken sidebar links due to excessive parent directory elements (`../`). Dokka already handles these in their navigation loader, we need to prune all the parent directory elements rather than just some of them

## Description of changes


services/timestreamquery/e2eTest/src/EndpointDiscoveryTest.kt
ianbotsf, line 1: **Correctness**: Why are we deleting this file?

lauzadis: This e2e test fails with

```
    java.lang.IllegalStateException: Unexpected response Failure(java.lang.IllegalStateException: Unexpected response Failure(aws.sdk.kotlin.services.timestreamwrite.model.TimestreamWriteException: Only existing Timestream for LiveAnalytics customers can access the service. Reach out to AWS support, for more information.))
```

lauzadis: ~~I should have deleted the test for **timestreamwrite**, not **timestreamquery**, fixed that. ~~

edit: both **timestreamwrite** and **timestreamquery** need to be disabled

PR #1687, opened 2025-09-16
feat: minor version strategy rules

0marperez:
## Issue \#

N/A

## Description of changes

- Configure minor-version-strategy rules (deprecatedUntilVersion)
- GH workflow to run new rule

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -12,8 +12,8 @@ atomicfu-version = "0.29.0"
 binary-compatibility-validator-version = "0.18.0"
 
 # smithy-kotlin codegen and runtime are versioned separately
-smithy-kotlin-runtime-version = "1.5.8"
-smithy-kotlin-codegen-version = "0.35.8"
+smithy-kotlin-runtime-version = "1.5.10"
+smithy-kotlin-codegen-version = "0.35.10"
 
 # codegen
 smithy-version = "1.60.2"


ktlint-rules/minor-version-strategy/src/main/kotlin/aws/sdk/kotlin/ktlintrules/minorversionstrategy/MinorVersionStrategyRuleSetProvider.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.pinterest.ktlint.cli.ruleset.core.api.RuleSetProviderV3
+import com.pinterest.ktlint.rule.engine.core.api.RuleProvider
+import com.pinterest.ktlint.rule.engine.core.api.RuleSetId
+import software.aws.ktlint.rules.apisScheduledForRemovalRule
+
+/**
+ * Ruleset provider for AWS SDK Kotlin minor-version-bump-specific Ktlint rules.
+ */
+class MinorVersionStrategyRuleSetProvider : RuleSetProviderV3(RuleSetId("minor-version-strategy-rules")) {
+    private val sdkVersion = System.getProperty("sdkVersion").split(".")
```
lauzadis: correctness: this looks at JVM system properties, not Gradle properties right?

0marperez:
> I don't like that every project now needs to define a ktlint-rules module. Can we move the rule registration to aws-kotlin-repo-tools? We already have a reference to the Gradle Project there and should be able to get the sdkVersion and anything else you need

We can't move it to repo tools because the rule would need to be published and then consumed downstream i.e. the minor and major version would be hardcoded. We can access the minor and major version using the `Project`, but rule registration can't be done in the `build.gradle.kts` file downstream.

ktlint-rules/minor-version-strategy/src/main/kotlin/aws/sdk/kotlin/ktlintrules/minorversionstrategy/MinorVersionStrategyRuleSetProvider.kt
0marperez: Correct, I don't think we can access `gradle.properties` at runtime unless we explicitly read the file. We're setting the Gradle property as a system property [here](https://github.com/aws/aws-sdk-kotlin/blob/ff65f6f4a5a31fbd30d2468030ec44beccf0c524/build.gradle.kts#L112-L115).

build.gradle.kts
```diff
@@ -106,3 +107,9 @@ val lintPaths = listOf(
 )
 
 configureLinting(lintPaths)
+configureMinorVersionStrategyRules(lintPaths)
+
+// Set SDK version from gradle.properties as a system property for 'configureMinorVersionStrategyRules' to use
+tasks.withType<JavaExec> {
```
lauzadis: this can be refined to only a single task using tasks.named("...")

ktlint-rules/minor-version-strategy/src/main/kotlin/aws/sdk/kotlin/ktlintrules/minorversionstrategy/MinorVersionStrategyRuleSetProvider.kt
lauzadis: Oh I missed that, I'm not a fan of that approach but I'm not sure there's anything better

lauzadis (approved):

PR #1694, opened 2025-09-23
misc: collect ci metrics

xinsong-cui:
## Description of changes

Collect workflow CI metrics and upload to Cloudwatch

Metrics being collected:
- execution time 
- success rate

### DIFF
--- a/.github/workflows/codebuild-ci.yml
+++ b/.github/workflows/codebuild-ci.yml
@@ -75,6 +75,7 @@ jobs:
           fi
       - name: Calculate duration
         id: end
+        if: always()
         run: |
           printf -v now '%(%s)T'
           duration=$(( now - ${{ steps.start.outputs.timestamp }} ))
@@ -185,6 +186,7 @@ jobs:
           }
       - name: Calculate duration
         id: end
+        if: always()
         run: |
           printf -v now '%(%s)T'
           duration=$(( now - ${{ steps.start.outputs.timestamp }} ))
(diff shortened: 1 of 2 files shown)


.github/workflows/codebuild-ci.yml
```diff
@@ -43,6 +43,11 @@ jobs:
     if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
```
0marperez: Question: This was added in the previous revision. Do we still need to checkout sources?

.github/workflows/continuous-integration.yml
```diff
@@ -4,7 +4,9 @@ on:
   pull_request:
   workflow_dispatch:
 
-permissions: { }
+permissions:
+  id-token: write
+  contents: read
```
0marperez, line 9: Question: Why do we need these new permissions?

.github/workflows/codebuild-ci.yml
xinsong-cui: Nice catch, I don't think we need them now

.github/workflows/continuous-integration.yml
xinsong-cui: also carry from last revision, we no longer need this

xinsong-cui: We need them for configure credentials

0marperez (approved):

PR #1701, opened 2025-10-03
misc: refactor artifact size metrics & service check batch

0marperez:
## Issue \#

N/A

## Description of changes

Refactors service check batch, e2e tests, and artifact size metrics
See: https://github.com/aws/aws-kotlin-repo-tools/pull/150

### DIFF
--- a/.github/workflows/service-ci.yml
+++ b/.github/workflows/service-ci.yml
@@ -28,7 +28,7 @@ jobs:
           env-vars-for-codebuild: GITHUB_REPOSITORY, UPLOAD, RELEASE_METRICS, IDENTIFIER
 
       - name: Process metrics
-        uses: awslabs/aws-kotlin-repo-tools/.github/actions/artifact-size-metrics/download-and-process@asm-v2 # TODO: Change to main before merging
+        uses: awslabs/aws-kotlin-repo-tools/.github/actions/artifact-size-metrics/download-and-process@main
         with:
           download: 'true'
 


lauzadis (approved):

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,41 @@
...
+            aws codebuild stop-build --id $BUILD_ID
+          fi
+
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  id-token: write
+  contents: read
```
lauzadis, line 40: nit: these typically go at the top of the file

build.gradle.kts
```diff
@@ -34,14 +34,6 @@ plugins {
     // ensure the correct version of KGP ends up on our buildscript classpath
     id(libs.plugins.kotlin.multiplatform.get().pluginId) apply false
     id(libs.plugins.kotlin.jvm.get().pluginId) apply false
-    alias(libs.plugins.aws.kotlin.repo.tools.artifactsizemetrics)
-}
-
-artifactSizeMetrics {
-    artifactPrefixes = setOf(":services", ":aws-runtime")
```
lauzadis, line 41: How are we filtering the artifact size computation to only these projects now?

0marperez: We're not filtering for artifacts from those two projects anymore. Now we're filtering for artifacts that are configured for publishing to Maven. Our updated codebuild service check job does this by running `publishAllPublicationsToTestLocalRepository` and then looking in the test local repository.

It should be better because we won't have to maintain hard coded lists anymore like this one. We never updated this to include hll's.

lauzadis: Ok right, that sounds great

PR #1751, opened 2025-12-12
misc: add `DynamoDbAttributeConverter` annotation

lauzadis:
## Issue \#

Addresses https://github.com/aws/aws-sdk-kotlin/issues/1457

## Description of changes

### DIFF
--- a/hll/dynamodb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
@@ -19,7 +19,6 @@ import com.google.devtools.ksp.KspExperimental
 import com.google.devtools.ksp.getAnnotationsByType
 import com.google.devtools.ksp.getConstructors
 import com.google.devtools.ksp.isAnnotationPresent
-import com.google.devtools.ksp.processing.KSPLogger
 import com.google.devtools.ksp.symbol.*
 
 /**


hll/dynamodb-mapper/dynamodb-mapper-annotations/common/src/aws/sdk/kotlin/hll/dynamodbmapper/Annotations.kt
```diff
@@ -4,20 +4,28 @@
...
  * included then the attribute name matches the property name.
  */
 @Target(AnnotationTarget.PROPERTY)
 public annotation class DynamoDbAttribute(val name: String)
 
+/**
+ * Specifies the type of [ValueConverter] to be used when processing this attribute.
+ */
+public annotation class DynamoDbAttributeConverter(val converter: KClass<out ValueConverter<*>>)
+
```
ianbotsf, line 22: **Comment**: I always seem to forget that annotation classes can only take `KClass`/`Class` arguments, not actual class instances. 😖 Several of the value converters (e.g., [`BooleanValueConverter`](https://github.com/aws/aws-sdk-kotlin/blob/983bcdb3f6f59fea14d92b8bf1fb07ad6e63daa6/hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/values/scalars/BooleanValueConverter.kt#L15), the [various number value converters](https://github.com/aws/aws-sdk-kotlin/blob/983bcdb3f6f59fea14d92b8bf1fb07ad6e63daa6/hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/values/scalars/NumberValueConverters.kt), etc.) are defined as `val` instances. We'll need to update those `class` definitions (or maybe `object`?) to allow users to select them in annotations. Can you either leave a `FIXME` somewhere and/or create a second task to handle updating the converter declarations?

hll/dynamodb-mapper/dynamodb-mapper-annotations/common/src/aws/sdk/kotlin/hll/dynamodbmapper/Annotations.kt
```diff
@@ -4,20 +4,28 @@
...
 /**
  * Specifies that this class/interface describes an item type in a table. All public properties of this type will be mapped to
  * attributes unless they are explicitly ignored.
  * @param converterName The fully qualified name of the item converter to be used for converting this class/interface.
  * If not set, one will be automatically generated.
  */
-// FIXME Update to take a KClass<ItemConverter>, which will require splitting codegen modules due to a circular dependency
+// FIXME Update to take a KClass<ItemConverter>?
 @Target(AnnotationTarget.CLASS)
 public annotation class DynamoDbItem(val converterName: String = "")
```
ianbotsf, line 31: **Comment**: Yes, we'll definitely still want to do this. I think that'll be far more convenient.

hll/dynamodb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/HighLevelRenderer.kt
```diff
@@ -42,7 +42,7 @@ internal class HighLevelRenderer(
                 attributes,
             )
 
-            val annotation = SchemaRenderer(annotated, renderCtx)
+            val annotation = SchemaRenderer(logger, annotated, renderCtx)
             annotation.render()
```
ianbotsf, line 46: **Question**: Isn't the logger already available in `renderCtx`?

hll/dynamodb-mapper/dynamodb-mapper-annotations/build.gradle.kts
```diff
@@ -6,3 +6,14 @@
...
+kotlin {
+    sourceSets {
+        commonMain {
+            dependencies {
+                // For ValueConverter
+                implementation(project(":hll:dynamodb-mapper:dynamodb-mapper"))
+            }
+        }
+    }
+}
```
ianbotsf, line 19: **Comment**: I'd hoped to keep the dependencies clean for the annotations package but we obviously need the `ValueConverter` (and soon, `ItemConverter`) types. I wonder if we should extract a **dynamodb-mapper-core-api** package which contains some important interfaces but no implementations.

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
```diff
@@ -569,4 +569,41 @@ class SchemaGeneratorPluginTest {
...
+        // Test cross-package converter
+        assertContains(schemaContents, "import a.different.pkg.HealthcareConverter")
+        assertContains(
+            schemaContents,
+            """        AttributeDescriptor(
+            "healthcare",
+            Employee::healthcare,
+            Employee::healthcare::set,
+            a.different.pkg.HealthcareConverter(),
+        ),""",
```
ianbotsf, line 606: **Nit**: String indentation is a little wonky.

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/resources/attribute-converter/Employee.kt
```diff
@@ -0,0 +1,41 @@
...
+    @DynamoDbPartitionKey
+    var id: Int = 1,
+    var givenName: String = "Johnny",
+    var surname: String = "Appleseed",
+
+    @property:DynamoDbAttributeConverter(OccupationConverter::class)
+    var occupation: Occupation = Occupation("Student", 0),
+
+    @property:DynamoDbAttributeConverter(HealthcareConverter::class)
+    var healthcare: Healthcare = Healthcare(false),
```
ianbotsf: **Question**: Having to qualify the target in the annotation line seems kind of ugly and unintuitive. Is this still required if the annotation class is annotated with `@Target(AnnotationTarget.PROPERTY)`?

hll/dynamodb-mapper/dynamodb-mapper-annotations/build.gradle.kts
lauzadis: It might be better organized that way but in practice users can't use the annotations standalone, they will still need to add a dependency on the mapper package. I don't think breaking out a core-api package will reduce the dependency closure if that's what you're thinking about?

hll/dynamodb-mapper/dynamodb-mapper-annotations/common/src/aws/sdk/kotlin/hll/dynamodbmapper/Annotations.kt
lauzadis: Created a new task: SDK-KT-1001

hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
lauzadis: These can't be easily fixed, I tried before sending this out for a review

hll/dynamodb-mapper/dynamodb-mapper-annotations/common/src/aws/sdk/kotlin/hll/dynamodbmapper/Annotations.kt
lauzadis: I made a new task for this: SDK-KT-1002

hll/dynamodb-mapper/dynamodb-mapper-annotations/build.gradle.kts
ianbotsf: I agree that _some_ application package will need a dependency on the mapper package but that's not necessarily the same as the package which holds their data classes, which could theoretically be used in many non-SDK contexts. In any event, splitting up the module dependencies is outside the scope of this PR se we can leave it for now and make a final decision closer to launch time.

ianbotsf (approved):

PR #1757, opened 2025-12-17
chore: annotate DDB Mapper-generated declarations as @GeneratedApi

ianbotsf:
## Issue \#

(none)

## Description of changes

Annotates mapper-generated declarations with `@GeneratedApi` and sets that as a "non-public marker" for the purposes of API validation.

### Examples

Examples of updated generated schemas are present in the PR in **SchemaGeneratorPluginTest.kt**.

Generated operations are a little harder to see in PR so check out **dynamodb-mapper.api** to see which members are no longer validated. Despite no longer appearing in the API dump, the members are still present in the generated operation. For example:

```kotlin
public interface ScanRequest<T> {
    public companion object { }

    @GeneratedApi
    public val consistentRead: Boolean?
    public val exclusiveStartKey: T?
    public val filter: BooleanExpr?
    @GeneratedApi
    public val limit: Int?
    @GeneratedApi
    public val returnConsumedCapacity: ReturnConsumedCapacity?
    @GeneratedApi
    public val segment: Int?
    @GeneratedApi
    public val select: Select?
    @GeneratedApi
    public val totalSegments: Int?
}

public class ScanRequestBuilder<T> {
    @GeneratedApi
    public var consistentRead: Boolean? = null
    public var exclusiveStartKey: T? = null

    public var filter: BooleanExpr? = null

    public fun filter(block: Filter.() -> BooleanExpr?) {
        filter = FilterImpl.run(block)
    }

    @GeneratedApi
    public var limit: Int? = null
    @GeneratedApi
    public var returnConsumedCapacity: ReturnConsumedCapacity? = null
    @GeneratedApi
    public var segment: Int? = null
    @GeneratedApi
    public var select: Select? = null
    @GeneratedApi
    public var totalSegments: Int? = null

    public fun build(): ScanRequest<T> {
        // implementation omitted
    }
}
```

**Companion PR**: https://github.com/smithy-lang/smithy-kotlin/pull/1471

### DIFF
--- a/hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/MemberCodegenBehavior.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/MemberCodegenBehavior.kt
@@ -146,10 +146,6 @@ private val rules = listOf(
     Rule("attributesToGet", Types.Kotlin.list(Types.Kotlin.String), Drop),
     Rule("attributeUpdates", Types.Kotlin.stringMap(llType("AttributeValueUpdate")), Drop),
 
-    // FIXME Add support for workloadProfileName as a pass-through field by applying a @Synthetic annotation to
-    //  pass-through members and ignoring API validation for them
-    Rule("workloadProfileName", Types.Kotlin.String, Drop),
-
     // Hoisted members
     Rule("tableName", Types.Kotlin.String, Hoist),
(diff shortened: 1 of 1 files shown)


hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/model/HllTypes.kt
```diff
@@ -0,0 +1,18 @@
+package aws.sdk.kotlin.hll.codegen.model
+
+import aws.sdk.kotlin.runtime.InternalSdkApi
+
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
0marperez, line 4: Nit: The copyright header isn't at the top of the file

PR #1773, opened 2026-01-06
feat: `@DynamoDbTtlSeconds` annotation

lauzadis:
## Description of changes

### DIFF
--- a/hll/dynamodb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
@@ -349,9 +349,9 @@ internal class SchemaRenderer(
                 }
             } else {
                 write(
-                    "override val attributes: #T = #T { }",
+                    "override val attributes: #T = #T()",
                     Type.from(RuntimeTypes.Core.Collections.Attributes),
-                    Type.from(RuntimeTypes.Core.Collections.attributesOf),
+                    Type.from(RuntimeTypes.Core.Collections.emptyAttributes),
                 )
             }
         }
--- a/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper-schema-generator-plugin/src/test/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/plugins/SchemaGeneratorPluginTest.kt
@@ -142,7 +142,7 @@ class SchemaGeneratorPluginTest {
                 public object UserSchema : ItemSchema.PartitionKey<User, KeyType.Key1<Int>> {
                     override val converter: UserConverter = UserConverter
                     override val partitionKey: KeySpec.Key1<Int> = KeySpec.number<Int>("id")
-                    override val attributes: Attributes = attributesOf { }
+                    override val attributes: Attributes = emptyAttributes()
                 }
             """.trimIndent(),
         )
@@ -413,7 +413,7 @@ class SchemaGeneratorPluginTest {
                 public object CustomUserSchema : ItemSchema.PartitionKey<CustomUser, KeyType.Key1<Int>> {
                     override val converter: MyCustomUserConverter = MyCustomUserConverter
                     override val partitionKey: KeySpec.Key1<Int> = KeySpec.number<Int>("id")
-                    override val attributes: Attributes = attributesOf { }
+                    override val attributes: Attributes = emptyAttributes()
                 }
             """.trimIndent(),
         )
@@ -570,7 +570,7 @@ class SchemaGeneratorPluginTest {
                 public object RenamedPartitionKeySchema : ItemSchema.PartitionKey<RenamedPartitionKey, KeyType.Key1<Int>> {
                     override val converter: RenamedPartitionKeyConverter = RenamedPartitionKeyConverter
                     override val partitionKey: KeySpec.Key1<Int> = KeySpec.number<Int>("user_id")
-                    override val attributes: Attributes = attributesOf { }
+                    override val attributes: Attributes = emptyAttributes()
                 }
             """.trimIndent(),
         )


hll/dynamodb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -163,6 +163,16 @@ internal class SchemaRenderer(
             }
 
             // converter
+
+            // Handle TTL annotation
+            prop.annotations
+                .singleOrNull { it.annotationType.resolve().declaration.qualifiedName?.asString() == DynamoDbTTL::class.qualifiedName }
+                ?.let {
+                    val lifetimeSeconds = it.arguments.single().value as? Long ?: error("@DynamoDbTTL annotation argument could not be evaluated at compile time. Use a literal value like @DynamoDbTTL(3600) instead of expressions like @DynamoDbTTL(1.hours.inWholeSeconds).")
```
lauzadis: I first tried configuring the annotation like `@DynamoDbTTL(1.hours.inWholeSeconds)`, but trying to access this argument value in KSP always returned null. Only constants like `@DynamoDbTTL(3600)` could be resolved by KSP

hll/dynamodb-mapper/dynamodb-mapper-annotations/common/src/aws/sdk/kotlin/hll/dynamodbmapper/Annotations.kt
```diff
@@ -49,3 +48,11 @@ public annotation class DynamoDbSortKey
...
 @Target(AnnotationTarget.PROPERTY)
 public annotation class DynamoDbIgnore
+
+/**
+ * Specifies that this property is used to track the item's time-to-live (TTL).
+ * https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html
+ * @param lifetimeSeconds The lifetime of this item, in seconds
+ */
+@Target(AnnotationTarget.PROPERTY)
+public annotation class DynamoDbTTL(val lifetimeSeconds: Long)
```
ianbotsf: **Style**: Since we're confined to using a scalar type and, by implication, a fixed time unit let's make that part of the annotation class name to make calling code more self-documenting:

```kotlin
@DynamoDbTtl(3600) // 3600 what?
val foo: String

// versus

@DynamoDbTtlSeconds(3600) // 3600 seconds, obvs
val foo: String
```

ianbotsf: **Style**: `DynamoDbTTL` → `DynamoDbTtl`

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/values/TTLValueConverter.kt
```diff
@@ -0,0 +1,36 @@
...
+    override val left: MonoConverter<AttributeValue, Long> = MonoConverter {
+        it.asN().toLong()
+    }
+
+    override val right: MonoConverter<Long, AttributeValue> = MonoConverter {
+        val currentTimeSeconds = clock.now().epochSeconds
+        val ttlValue = currentTimeSeconds + lifetimeSeconds
+        AttributeValue.N(ttlValue.toString())
+    }
+}
```
ianbotsf: **Correctness**: I hadn't considered supporting TTL (or RVN, timestamp, etc.) via a `ValueConverter` but I don't think it'll achieve what we're hoping for. I can see how it'll work on `PutItem` calls but `UpdateItem` is an important API we'll need to support too...one which won't use value converters at all because it'll operate on low-level attributes.

I think interceptors may be the better approach here, although we'll likely need some additional types to allow a schema to convey that certain entity members represent special values. FWIW, the `Structure` and `Member` types defined in **hll-codegen** already have `val attributes: Attributes` members which may work. If we intend to use them, though, we'd need to convert them from `@InternalSdkApi data class` to something safer like interfaces with backing `private data class` implementations and factory functions.

hll/dynamodb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -320,6 +321,38 @@ internal class SchemaRenderer(
                 keySpecInstantiation(sortKeyProps)
                 write()
             }
+
+            // Handle TTL annotation
+            val ttlFields = properties.mapNotNull { prop ->
+                prop.annotations
+                    .singleOrNull { it.annotationType.resolve().declaration.qualifiedName?.asString() == DynamoDbTtlSeconds::class.qualifiedName }
+                    ?.let { annotation ->
+                        val lifetime = annotation.arguments.single().value as? Long ?: error("@DynamoDbTtlSeconds annotation argument could not be evaluated at compile time. Use a literal value like @DynamoDbTtlSeconds(3600) instead of expressions like @DynamoDbTtlSeconds(1.hours.inWholeSeconds).")
```
ianbotsf: **Question**: Do KSP's troubles with annotation args ever manifest as the `arguments` collection being completely empty? Because if so, `single()` will throw an exception before you can attempt to retrieve the value. If that's a concern, `singleOrNull` would be better. If that's not a concern then disregard.

ianbotsf: **Nit**: Let's include the property name in this error message as well.

**Side comment**: I think we may want a more robust error-throwing strategy for our schema codegen since users will have to deal with the exceptions. Ideally we'd point them directly to the class, property, file, line, and position of the error (where applicable). Let's brainstorm.

hll/dynamodb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -320,6 +321,38 @@ internal class SchemaRenderer(
...
+                        require(lifetime > 0) { "@DynamoDbTtlSeconds must be positive, got $lifetime seconds on property ${prop.ddbName}" }
+                        prop.simpleName.getShortName() to lifetime
+                    }
+            }
+            require(ttlFields.size <= 1) { "Only one @DynamoDbTtlSeconds annotation is allowed, found ${ttlFields.size} on properties: ${ttlFields.joinToString { it.first }}" }
+            val ttlField = ttlFields.singleOrNull()
```
ianbotsf: **Question**: _Is_ multiple TTL annotations on a single entity a problem? It doesn't seem very useful but DDB generally favors denormalized schemata where things like data duplication/aliasing might be more acceptable.

hll/dynamodb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -320,6 +321,38 @@ internal class SchemaRenderer(
...
+                    "override val attributes: #T = #T {",
+                    "}",
+                    Type.from(RuntimeTypes.Core.Collections.Attributes),
+                    Type.from(RuntimeTypes.Core.Collections.attributesOf),
+                ) {
+                    write(
+                        "#T.#L to #T(#S, #LL)",
+                        MapperTypes.Model.SchemaAttributes,
+                        "TtlField",
+                        TypeRef("kotlin", "Pair"),
```
ianbotsf: **Nit**: Let's add this to `Types.Kotlin`.

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/model/Type.kt
```diff
@@ -39,6 +40,15 @@ public sealed interface Type {
...
         }
+
+        /**
+         * Derives a [TypeRef] from a Smithy [Symbol]
+         */
+        public fun from(symbol: Symbol, nullable: Boolean = false): TypeRef = TypeRef(
+            pkg = symbol.namespace,
+            shortName = symbol.name,
+            nullable = nullable,
+        )
```
ianbotsf, line 51: **Comment**: Nice!

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/interceptors/TtlInterceptor.kt
```diff
@@ -0,0 +1,41 @@
...
+            // TODO Support UpdateItem
+            else -> request
+        }
+    }
+
+    private fun handlePutItem(request: LowLevelPutItemRequest, ttlField: Pair<String, Long>): LowLevelPutItemRequest {
+        val newItem = request.item?.toMutableMap() ?: return request
+        val (fieldName, lifetime) = ttlField
+
+        val ttlValue = AttributeValue.N((clock.now().epochSeconds + lifetime).toString())
```
ianbotsf: **Nit**: Could simplify with the `attr` helper function:

```kotlin
val ttlValue = attr(clock.now().epochSeconds + lifetime) // Returns AttributeValue.N
```

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/items/ItemSchema.kt
```diff
@@ -24,6 +26,12 @@ public sealed interface ItemSchema<T> {
      */
     public val keyAttributeNames: List<String>
 
+    /**
+     * Generic attributes for associating metadata with this schema
+     */
+    public val attributes: Attributes
+        get() = emptyAttributes()
```
ianbotsf: **Question**: Why do we need a default implementation for this field and why should it be `emptyAttributes()`?

ianbotsf (approved):

hll/dynamodb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
lauzadis: I haven't seen that. I've seen a `ClassNotFoundException` in the case of `KClass` arguments and a null argument `.value` in the case of expressions like `1.hours.inWholeSeconds`

hll/dynamodb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
lauzadis: DynamoDB only supports one TTL attribute: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/time-to-live-ttl-how-to.html

My thinking was that if we let users configure TTL on multiple attributes, but only one is actually used, we'd be misleading them

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/items/ItemSchema.kt
lauzadis: We don't need one, I will remove it

PR #1796, opened 2026-02-09
feat: key-projected request/response types

ianbotsf:
## Issue \#

Related to #472 

## Description of changes

This change expands generated high-level operation requests and responses into subtypes based on the presence of item key fields. The end effect is strongly-typed keys in high-level operation inputs/outputs. This change is a big one so maybe start with **dynamodb-mapper-ops-codegen/README.md** and explore outward from there.

### DIFF
--- a/tests/benchmarks/service-benchmarks/build.gradle.kts
+++ b/tests/benchmarks/service-benchmarks/build.gradle.kts
@@ -2,7 +2,7 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
-import aws.sdk.kotlin.gradle.dsl.skipPublishing
+import aws.sdk.kotlin.gradle.publishing.skipPublishing
 
 plugins {
     id(libs.plugins.kotlin.multiplatform.get().pluginId)


hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/model/internal/TableImpl.kt
```diff
@@ -100,68 +59,5 @@ internal fun <T, PK : KeyType, SK : KeyType> tableImpl(
...
-                    interceptors = it.interceptors.prepend(interceptor),
-                )
-            }
-            val hRes = op.execute(GetItemRequest { })
-            return hRes.item
-        }
-
-        override suspend fun putItem(request: PutItemRequest<T>): PutItemResponse<T> {
-            val ttlInterceptor = schema.ttlInterceptor
-            val counterInterceptor = schema.counterInterceptor
```
lauzadis, line 118: When do these interceptors get registered now?

ianbotsf: Whoops, they _don't_ get registered now. I needed to relocate this but never re-added it. Will fix.

PR #1803, opened 2026-02-18
chore: refactor DynamoDB Mapper ahead of support for update expressions

ianbotsf:
## Issue \#

(none)

## Description of changes

In preparation for update expressions and `UpdateItem` support, this change refactors a few key things in **dynamodb-mapper**:
* The `internal`-only `attr` functions are renamed to `av`. `attr` as a name for these was always a bad choice since we _also_ have the `public` `attr` function in `Filter`, which is confusing (for us as devs, not for users).
* The **jvmTest** unit tests are moved to **commonTest** and updated to use platform-neutral dependencies for file reading and mocking.
* The `Filter` and `SortKeyFilter` interfaces are renamed to `FilterDsl` and `SortKeyFilterDsl`, respectively. This change makes it clearer how the interfaces are used (i.e., as DSL receivers) and heads off some tricky naming semantics with the impending update expressions support.

### DIFF
--- a/hll/dynamodb-mapper/dynamodb-mapper/build.gradle.kts
+++ b/hll/dynamodb-mapper/dynamodb-mapper/build.gradle.kts
@@ -3,10 +3,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import software.amazon.dynamodb.services.local.main.ServerRunner
-import software.amazon.dynamodb.services.local.server.DynamoDBProxyServer
 import dev.mokkery.verify.VerifyMode
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask
+import software.amazon.dynamodb.services.local.main.ServerRunner
+import software.amazon.dynamodb.services.local.server.DynamoDBProxyServer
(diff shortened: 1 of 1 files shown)


gradle/libs.versions.toml
lauzadis: This doesn't have to be fixed in this PR, but FYI I found out the coordinate for this dependency changed, the new one is https://mvnrepository.com/artifact/software.amazon.dynamodb/DynamoDBLocal

hll/dynamodb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/OperationTest.kt
```diff
@@ -30,18 +37,17 @@ private const val TABLE_NAME = "foo-table"
 // FIXME Should be in commonTest but mockk is JVM-only and finding a good KMP mocking library is hard
```
0marperez: This comment no longer applies

PR #1808, opened 2026-02-20
feat: implementation of UpdateItem and update expressions

ianbotsf:
## Issue \#

This change introduces [update expressions](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.UpdateExpressions.html) and support for the [`UpdateItem`](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html) operation. It's another biggie but a lot of it is docs and DSL overloads.

I recommend starting with the KDocs for `UpdateDsl`, `UpdateDsl.Set`, `UpdateDsl.Remove`, etc. These give a good overview of the capabilities of the DSL and of update expressions in general. Also the tests in `UpdateItemTest` and `UpdateDslTest` may be helpful for seeing the DSL in action.

Aside from enabling update expressions, this PR also refactors attribute paths in expressions. Previously, top-level attributes were accessed via an `attr()` function but nested attributes/elements were accessed with the `get` operator, as in `attr("foo")["bar"]`. The punctuation difference felt awkward so I've changed the `attr()` function into a `val attr: Attr` which supports indexed access, as in `attr["foo"]["bar"]`.

## Description of changes

### DIFF
--- a/hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/internal/UpdateDslImpl.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/internal/UpdateDslImpl.kt
@@ -45,6 +45,14 @@ internal class UpdateSetImpl :
     override fun AttributePath.set(index: Int, value: Expression) = get(index).update(value)
     override fun AttributePath.set(key: String, value: Expression) = get(key).update(value)
 
+    override fun AttributePath.appending(
+        value: Expression,
+    ): Expression = ScalarFuncExpr(ScalarFunc.LIST_APPEND, this, value)
+
+    override fun List<Any?>.appending(
+        value: Expression,
+    ): Expression = ScalarFuncExpr(ScalarFunc.LIST_APPEND, LiteralExpr(this), value)
+
     override fun AttributePath.orElse(
         value: Expression,
     ): Expression = ScalarFuncExpr(ScalarFunc.IF_NOT_EXISTS, this, value)
--- a/hll/dynamodb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/expressions/UpdateDslTest.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/expressions/UpdateDslTest.kt
@@ -53,7 +53,7 @@ class UpdateDslTest {
     @Test
     fun testExpressions() {
         testUpdates(
-            "SET foo = oof, bar[0] = rab[0], baz.qux = zab.xuq, nox = xon + onx, pix = xip - ipx, moo = if_not_exists(oom, omo) ADD bur rub DELETE zip piz" to {
+            "SET foo = oof, bar[0] = rab[0], baz.qux = zab.xuq, nox = xon + onx, pix = xip - ipx, moo = if_not_exists(oom, omo), paz = list_append(paz, zap) REMOVE yum, muy ADD bur rub DELETE zip piz" to {
                 set {
                     attr["foo"] = attr["oof"]
                     attr["bar"][0] = attr["rab"][0]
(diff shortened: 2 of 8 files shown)


hll/dynamodb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/operations/UpdateItemTest.kt
```diff
@@ -0,0 +1,168 @@
...
+
+        private val converter = SimpleItemConverter(
+            ::Entity,
+            { this },
+            AttributeDescriptor("id", Entity::id, Entity::id::set, NumberValueConverters.Int),
+        )
+        private val schema = ItemSchema(converter, KeySpec.int("id"))
+    }
+
+    @BeforeAll
```
lauzadis, line 34: nit: missing explicit `@BeforeAll` import, can we add one? I've run into issues with this when moving from a JVM-only build to KMP.

hll/dynamodb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/expressions/UpdateDslTest.kt
```diff
@@ -0,0 +1,652 @@
...
+package aws.sdk.kotlin.hll.dynamodbmapper.expressions
+
+import aws.sdk.kotlin.hll.dynamodbmapper.expressions.internal.ParameterizingExpressionVisitor
+import aws.sdk.kotlin.hll.dynamodbmapper.expressions.internal.UpdateDslImpl
+import aws.sdk.kotlin.hll.dynamodbmapper.util.av
+import aws.sdk.kotlin.services.dynamodb.model.AttributeValue
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class UpdateDslTest {
```
lauzadis, line 14: This suite is missing tests for `remove`, should we have some here?

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/internal/UpdateDslImpl.kt
```diff
@@ -0,0 +1,86 @@
...
+internal abstract class UpdateClauseImpl(val action: UpdateAction) {
+    private val _updates = mutableListOf<UpdateClauseExpr>()
+    val updates: List<UpdateClauseExpr> get() = _updates
+
+    fun AttributePath.update(value: Expression) {
+        _updates += UpdateClauseExpr(action, this, value)
+    }
+}
+
+internal class UpdateSetImpl :
```
lauzadis, line 38: Should we support the `list_append` function? 

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.UpdateExpressions.html#Expressions.UpdateExpressions.SET.UpdatingListElements

hll/dynamodb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/operations/UpdateItemTest.kt
ianbotsf: These tests use **kotest** (not **kotlin-test**), wherein test classes inherit some **kotest**-defined superclass to provide the desired testing functionality. In our case, we inherit `AnnotationSpec` which provides `@BeforeAll` and other annotations as child classes. Thus they do not need to be imported and will not change if these tests change source sets (which they won't because they're already KMP tests).

hll/dynamodb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/expressions/UpdateDslTest.kt
ianbotsf: Good catch! Yes, will add.

hll/dynamodb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/operations/UpdateItemTest.kt
lauzadis: Ah ok, thanks. I thought this was `org.junit.jupiter.api.BeforeAll`

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/internal/UpdateDslImpl.kt
ianbotsf: Oh yes, we should. I seem to have missed that during my pass through the docs.

PR #1809, opened 2026-03-04
fix: actually run DDB mapper operation tests; actually convert high-level update expressions into low-level

ianbotsf:
## Issue \#

(none)

## Description of changes

The following items were missed in the previous [implementation of `UpdateItem` and update expressions](https://github.com/aws/aws-sdk-kotlin/pull/1808):
* Switch back to **kotlin-test** runner for DynamoDB Mapper operation tests. Presently, none of the operation tests are actually running because of a **kotest** misconfiguration.
  * Fixing the config is possible but **kotest** 6.0 ended support for JDK 8 so that library is effectively a dead end.
* Actually generate the code for conversion of high-level update expressions into low-level update expressions. Presently, the low-level update expression value is never changed from the default of `null`.

Upstream PR: https://github.com/smithy-lang/smithy-kotlin/pull/1514

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -60,10 +60,12 @@ allprojects {
             )
         }
 
-        // Required to enable reflective access in testing.
-        // See smithy-kotlin/runtime/testing/jvm/src/aws/smithy/kotlin/runtime/testing/SystemOverrides.kt
-        // for more info
-        jvmArgs("--add-opens=java.base/java.util=ALL-UNNAMED")
+        if (testJavaVersion == null || testJavaVersion.asInt() >= 9) {
+            // Required to enable reflective access in testing in JDK 9+.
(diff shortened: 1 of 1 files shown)


hll/dynamodb-mapper/dynamodb-mapper/build.gradle.kts
```diff
@@ -24,6 +24,7 @@ buildscript {
 
 plugins {
     alias(libs.plugins.ksp)
+    alias(libs.plugins.kotest)
```
lauzadis: Does this plugin application work for native builds?

ianbotsf: Yes, see [the **kotest** docs on setup](https://kotest.io/docs/framework/project-setup.html). The **Multiplatform** section mentions that the **kotest** and **ksp** plugins are both required.

PR #1813, opened 2026-03-10
feat: add support for `BatchGetItem`

ianbotsf:
## Issue \#

Related to #472 

## Description of changes

This change adds support for the `BatchGetItem` operation. Unlike previous operations, this is the first of a series of cross-table operations which work differently than normal operations like `GetItem` or `Scan`. Because of this, the operation is invoked on `DynamoDbMapper` itself rather than a `Table` or `Index` instance. Under the hood, this means that factory methods, high-low conversions, and operation configuration work differently leading to some changes in codegen:
* The `Operation` type has been refreshed to pass around full context objects instead of only specific members. This simplifies codegen in all cases and enables operations to use prior context (e.g., a request input) to control later transformations (e.g., forming the response output).
* `DynamoDbMapper` is now referenced as an `ItemSourceKind`, which controls how factory methods and `*Operations` interfaces are generated. Similar to `Table` and `Index`, `DynamoDbMapper` now inherits from `DynamoDbMapperOperations`, which is code-generated by KSP. This also necessitates the use of a dummy persistence spec to streamline invocation.
* `MemberCodegenBehavior` has a new `CustomTransformation` implementation which is used for `BatchGetItem`'s request and response types to provide user-friendly type-safe DSLs. This class _may_ change more as additional cross-table operations are implemented (e.g., `BatchWriteItem`, `TransactGetItems`, etc.).
* `DslInfo` is now more configurable to enable different types of DSL invocation, finalization, and naming.

## How to read this PR

This is another large feature so I recommend following this guide:
* Start with **BatchGetItemTest.kt** which shows two examples of `BatchGetItem`. This should help familiarize you with the API before diving into...
* **BatchGetItemRequestTable.kt** which contains the high-level representation of a single table in a `BatchGetItem` request. Invocations with multiple `table(...) { ... }` calls will get a corresponding number of `BatchGetItemRequestTable<*>` instances. These are what's converted down into the low-level `Map<String, KeysAndAttributes>` used by the DDB client.
* **BatchGetItemRequestTableDsl.kt** contains the (very small) DSL features for a table in a `BatchGetItem` call. This is the receiver inside a `table(...) { }` block.
* Finally, browse the changes in **MemberCodegenBehavior.kt** to see how the new `CustomTransformation` behavior is declared for `BatchGetItem` requests/responses. This type will be reused for the rest of the batch/transact APIs.
* Then, all the rest!

### DIFF
--- a/hll/build.gradle.kts
+++ b/hll/build.gradle.kts
@@ -99,11 +99,11 @@ val projectsToIgnore = listOf(
     "dynamodb-mapper-codegen",
     "dynamodb-mapper-ops-codegen",
     "dynamodb-mapper-schema-codegen",
-    "dynamodb-mapper-schema-generator-plugin-test",
 ).filter { it in subprojects.map { it.name }.toSet() } // Some projects may not be in the build depending on bootstrapping
 
 apiValidation {
     ignoredProjects += projectsToIgnore
+    nonPublicMarkers += "aws.smithy.kotlin.runtime.GeneratedApi"
 }
 
 // Configure Dokka for subprojects
--- a/hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/MemberCodegenBehavior.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/MemberCodegenBehavior.kt
@@ -200,10 +200,6 @@ private val rules = listOf(
     Rule("attributesToGet", Types.Kotlin.list(Types.Kotlin.String), Drop),
     Rule("attributeUpdates", Types.Kotlin.stringMap(llType("AttributeValueUpdate")), Drop),
 
-    // FIXME Add support for workloadProfileName as a pass-through field by applying a @Synthetic annotation to
-    //  pass-through members and ignoring API validation for them
-    Rule("workloadProfileName", Types.Kotlin.String, Drop),
-
     // Hoisted members
     Rule("tableName", Types.Kotlin.String, Hoist),
     Rule("indexName", Types.Kotlin.String, Hoist),
--- a/hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/KeyProjectedTypeRenderer.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/KeyProjectedTypeRenderer.kt
@@ -229,7 +229,7 @@ internal class KeyProjectedTypeRenderer(
         BuilderRenderer(
             ctx = builderCtx,
             generator = this@KeyProjectedTypeRenderer,
-            builtType = projection.interfaceStruct.type,
+            builtStructure = projection.interfaceStruct,
             implementationType = projection.implStruct.type,
             members = projection.builderStruct.members,
             builderNameOverride = projection.builderStruct.type.shortName,
--- a/hll/dynamodb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
+++ b/hll/dynamodb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
@@ -9,6 +9,8 @@ import aws.sdk.kotlin.hll.codegen.model.*
 import aws.sdk.kotlin.hll.codegen.rendering.BuilderRenderer
 import aws.sdk.kotlin.hll.codegen.rendering.RenderContext
 import aws.sdk.kotlin.hll.codegen.rendering.RendererBase
+import aws.sdk.kotlin.hll.codegen.util.generatedAnnotation
+import aws.sdk.kotlin.hll.codegen.util.plus
 import aws.sdk.kotlin.hll.codegen.util.visibility
 import aws.sdk.kotlin.hll.dynamodbmapper.*
 import aws.sdk.kotlin.hll.dynamodbmapper.codegen.annotations.AnnotationsProcessorOptions
@@ -33,6 +35,7 @@ internal class SchemaRenderer(
     private val ctx: RenderContext,
 ) : RendererBase(ctx, "${classDeclaration.qualifiedName!!.getShortName()}Schema") {
     private val className = classDeclaration.qualifiedName!!.getShortName()
+    private val classStructure = Structure.from(classDeclaration)
     private val classType = Type.from(classDeclaration)
 
     private val builderName = "${className}Builder"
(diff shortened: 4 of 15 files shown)


hll/dynamodb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/operations/BatchGetItemTest.kt
```diff
@@ -0,0 +1,143 @@
...
+            table(bigDataTable) {
+                keys = bigDataIds.map(::Key)
+            }
+        }
+
+        val bigDataResp = resp.table(bigDataTable)
+
+        val processedIds = bigDataResp.items.map { it.id }
+        assertEquals(54, processedIds.size) // only 54 items fit in the 16MB maximum response size
+
```
lauzadis, line 136: Will this break if/when DynamoDB increases their response size limits? How likely do we think they'll increase limits in the future?

hll/dynamodb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/operations/BatchGetItemTest.kt
```diff
@@ -0,0 +1,143 @@
...
+        val resp = mapper.batchGetItem {
+            table(bigDataTable) {
+                keys = bigDataIds.map(::Key)
+            }
+        }
+
+        val bigDataResp = resp.table(bigDataTable)
+
+        val processedIds = bigDataResp.items.map { it.id }
+        assertEquals(54, processedIds.size) // only 54 items fit in the 16MB maximum response size
```
lauzadis, line 135: DynamoDB's docs say [this should be 52](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html), I'm assuming that's a mistake in their docs?

hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/FactoryRenderer.kt
```diff
@@ -48,15 +48,15 @@ import aws.sdk.kotlin.hll.dynamodbmapper.codegen.operations.model.*
  *         HReqContextImpl(highLevelReq, spec.schema, MapperContextImpl(spec, "GetItem"))
  *     },
  *
- *     serialize = { highLevelReq, schema -> highLevelReq.convert(spec.tableName, schema) },
+ *     serialize = { ctx -> ctx.highLevelRequest.convert(spec.tableName, ctx.serializeSchema) },
  *
  *     lowLevelInvoke = { lowLevelReq ->
  *         spec.mapper.client.withWrappedClient { client ->
- *             client.getItem(lowLevelReq)
+ *             client.getItem(ctx.lowLevelRequest)
```
lauzadis, line 55: nit: Where does `ctx` come from in this scope? Should `lowLevelReq` be changed to `ctx`

hll/dynamodb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/internal/OperationTest.kt
```diff
@@ -189,10 +189,6 @@ private data class LFooRequest(val table: String, val foo: Item)
 private data class LFooResponse(val foo: Item)
 private data class HFooResponse(val foo: Foo)
 
-private fun HFooRequest.convert(table: String, schema: ItemSchema<Foo>) = LFooRequest(table, schema.converter.convertRight(foo))
-
 private fun LFooResponse.convert(schema: ItemSchema<Foo>) = HFooResponse(schema.converter.convertLeft(foo))
```
lauzadis: nit: I think this `LFooResponse.convert` could be removed too

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/operations/BatchGetItemRequestTable.kt
```diff
@@ -0,0 +1,75 @@
...
+internal fun List<BatchGetItemRequestTable<*>>.convert(): Map<String, KeysAndAttributes> {
+    fun <PK : KeyType> pkMap(table: BatchGetItemRequestTable.PartitionKey<*, PK>) = table
+        .keys
+        .map { key -> keysToItem(table.schema, key) }
+
+    fun <PK : KeyType, SK : KeyType> ckMap(table: BatchGetItemRequestTable.CompositeKey<*, PK, SK>) = table
+        .keys
+        .map { (pk, sk) -> keysToItem(table.schema, pk, sk) }
+
+    return associate { table ->
```
lauzadis, line 64: If a user calls `table(sameTable) { ... }` multiple times in the same batch request, wouldn't this only keep the final entry? Do we support that use case? If so, should we merge `BatchGetItemRequestTable` instances? 

Also a unit test for this case would be helpful

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/core/TemplateEngine.kt
```diff
@@ -104,7 +104,9 @@ public class TemplateEngine(processors: List<TemplateProcessor>) {
                         append(pos)
                     }
                     append(" is outside of args bounds ")
-                    append(args.indices)
+                    appendLine(args.indices)
+                    appendLine("""Full template: "$template"""")
+                    appendLine("Args: $args")
```
lauzadis, line 109: Were these just for debugging or did you intend to keep these?

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/operations/BatchTransactDsls.kt
lauzadis, line 1: sneak peek!

hll/dynamodb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/operations/BatchGetItemTest.kt
ianbotsf: Yes, this would break if DDB changed their response size limits. They've been the same [since at least 2014](https://web.archive.org/web/20141009004138/http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html) so I think we're safe.

hll/dynamodb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/operations/BatchGetItemTest.kt
ianbotsf: Or an ambiguity. Calculating item size precisely is difficult as it depends on the names, values, and data types of all attributes. The DDB documentation does not describe the 300KB items precisely so I made my own schema and 54 was the number of items returned.

hll/dynamodb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/FactoryRenderer.kt
ianbotsf: Good catch, that should be `ctx ->` instead of `lowLevelReq ->`.

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/operations/BatchGetItemRequestTable.kt
ianbotsf: Yes the current behavior is to keep only the final entry for a given `Table` instance:

```kotlin
mapper.batchGetItem {
    table(foo) { keys(k1, k2, k3) } // New entry for `foo`
    table(bar) { keys(k4, k5, k6) } // New entry for `bar`
    table(foo) { keys(k7, k8, k9) } // Overwrite entry for `foo`; keys k1, k2, k3 no longer used
}
```

It's unclear to me why users would invoke the `table` method multiple times for the same table instance. We should discuss this in more depth.

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/core/TemplateEngine.kt
ianbotsf: They were useful in debugging so I chose to leave them in place for future debugging.

hll/dynamodb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/operations/BatchTransactDsls.kt
ianbotsf: Yeah, I accidentally committed my scratch file (removed in [latest commit](https://github.com/aws/aws-sdk-kotlin/pull/1813/changes/6808470cfcfaa6bef646515309afec3815feea12)). Everything in this file will _eventually_ become batch/transact APIs but they need a lot more work before they're ready.

PR #1824, opened 2026-03-24
fix: mrap e2e test logging & flakiness

0marperez:
## Issue \#

`SDK-KT-259`

## Description of changes

- Better logging
- Random postfix to MRAP to avoid naming collisions when test is running multiple times at the same time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## [1.6.43] - 03/24/2026
+
+### Features
+* (**bedrockagentcorecontrol**) Adds SDK support for 1) Persist session state in AgentCore Runtime via filesystemConfigurations in CreateAgentRuntime, UpdateAgentRuntime, and GetAgentRuntime APIs, 2) Optional name-based filtering on AgentCore ListBrowserProfiles API.
+* (**mediapackagev2**) Reduces the minimum allowed value for startOverWindowSeconds from 60 to 0, allowing customers to effectively disable the start-over window.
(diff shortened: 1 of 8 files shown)


services/s3/e2eTest/src/MultiRegionAccessPointTest.kt
```diff
@@ -31,12 +31,13 @@ import org.junit.jupiter.params.ParameterizedTest
...
 import java.util.stream.Stream
+import kotlin.random.Random
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.minutes
 import kotlin.time.Duration.Companion.seconds
 
 private const val MRAP_BUCKET_PREFIX = "s3-mrap-test-bucket-"
-private const val MULTI_REGION_ACCESS_POINT_NAME = "aws-sdk-for-kotlin-test-multi-region-access-point"
+private val MULTI_REGION_ACCESS_POINT_NAME = "aws-sdk-for-kotlin-mrap-${Random.nextInt(0, 9999)}"
 private const val TEST_OBJECT_KEY = "test.txt"
```
ianbotsf, line 41: **Question**: Why are we randomizing the access point name? We create and delete it with every test run so it shouldn't collide.

0marperez: I think the issue that is causing MRAP creation to fail is naming collisions when we have two different CI runs of this E2E test running at the same time. 

So even though we create and delete it every test run. There are situations in which two tests try to create it at the same time.

PR #1855, opened 2026-04-09
feat: download object 

0marperez:
## Issue \#

N/A

## Description of changes

- Downstream changes from https://github.com/smithy-lang/smithy-kotlin/pull/1551
- Miscelanneous KDoc changes and fixes to existing S3 TM code
- Download object operation
- Refactor`ceilDiv`so it can be used in download object implementation
- Misc changes so KN builds succeed (in `TransferInterceptorTest`)


hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/interceptors/TransferContext.kt
```diff
@@ -53,8 +53,8 @@ public interface TransferInterceptorContext {
 }
 
 /**
- * The context around an [aws.sdk.kotlin.hll.s3transfermanager.S3TransferManager] transfer.
- * Used to track transfer progress or to modify in progress transfers.
+ * The context around a [aws.sdk.kotlin.hll.s3transfermanager.S3TransferManager] transfer.
+ * Used to track in progress transfers.
  */
```
ianbotsf, line 58: **Nit**: "an" was correct, "a" is incorrect

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/interceptors/TransferContext.kt
```diff
@@ -68,8 +68,8 @@ public class TransferContext(
 ) : TransferInterceptorContext
 
 /**
- * The context around an [aws.sdk.kotlin.hll.s3transfermanager.S3TransferManager] transfer.
- * Used to track transfer progress or to modify in progress transfers.
+ * The context around a [aws.sdk.kotlin.hll.s3transfermanager.S3TransferManager] transfer.
+ * Used to modify in progress transfers.
  */
 public class MutableTransferContext(
```
ianbotsf, line 74: **Nit**: "an" was correct, "a" is incorrect

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+    private val networkOperation: Semaphore,
+    private val diskOperation: Semaphore,
+    private val maxInMemoryParts: Int,
+    private val targetPartSizeBytes: Long,
+    private val bufferCount: Semaphore,
+) {
+    /**
+     * Used to interact with the underlying systems files.
+     */
+    private val system: Filesystem = PlatformProvider.System
```
ianbotsf, line 54: **Question**: Why do we have a `private val` for this? We don't allow callers to inject their own `Filesystem` and we just use a singleton value so it seems like we could just replace `system` with `PlatformProvider.System`. Or make it injectable as a constructor parameter but you're not actually unit testing this, just the full E2E transfer, so that might be overkill.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+        if (needMoreParts) {
+            transferRemainingParts()
+        }
+        buildResponse()
+    }
+
+    private suspend fun transferFirstPart() {
+        executePhase(TransferPhase.BytesTransferred, context, interceptors) {
+            try {
+                networkOperation.acquire()
```
ianbotsf: **Nit**: Prefer the [`withPermit`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/with-permit.html) method over manually acquiring and releasing permits. Applies to `diskOperation` too.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+                s3Client.withTmBusinessMetric { client ->
+                    client.getObject(context.s3Request as GetObjectRequest) { getObjectResponse ->
+                        context.s3Response = getObjectResponse
+                        context.transferredBytes = getObjectResponse.contentLength
+                        context.transferableBytes = parseTransferableBytes(getObjectResponse)
+
+                        contentLength = context.transferableBytes
+                        etag = getObjectResponse.eTag ?: throw S3TransferManagerException("etag is null in initial get object response")
```
ianbotsf: **Question**: When would `eTag` be null?

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+                        partCount = getObjectResponse.partsCount
+
+                        val bytes = getObjectResponse.body?.toByteArray() ?: byteArrayOf()
```
ianbotsf, line 114: **Question**: When would `body` be null?

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+                            downloadPath?.let {
+                                system.write(
+                                    tempDownloadPath,
+                                    bytes,
+                                    WriteType.OVERWRITE,
+                                )
+                            }
+                        } finally {
+                            diskOperation.release()
+                        }
```
ianbotsf: **Question**: When would `downloadPath` be null? Do we still need to acuire the `diskOperation` permit if we're not going to write anything to disk?

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+
+                        if (partCount == 1 || context.transferredBytes == context.transferableBytes) {
```
ianbotsf: **Question**: Is OR (`||`) the right boolean operator here? When would `partCount == 1` be disjoint with `context.transferredBytes == context.transferableBytes`?

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+                system.delete(tempDownloadPath, mustExist = false)
+                throw t
+            } finally {
+                networkOperation.release()
+            }
+        }
+    }
+
+    private suspend fun transferRemainingParts() {
+        val buffer = Channel<Part>(capacity = maxInMemoryParts)
```
ianbotsf: **Correctness**: This bypasses your `bufferSemaphore` and will cause _more_ than the configured number of parts to be loaded into memory when two transfers are happening simultaneously.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+    private suspend fun downloadParts(
+        buffer: Channel<Part>,
+    ) = coroutineScope {
+        val jobs = mutableSetOf<Job>()
+
+        repeat(partCount!! - 1) { i ->
+            // -1 since first part is already downloaded
+            jobs += launch {
+                // Other jobs will be updating the global context so make a copy that won't change and we can modify safely
+                val localContext = context.copy()
```
ianbotsf: **Question**: `context` will have been modified before some coroutines launch. Is the initial state of `localContext` important?

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+                                localContext.transferredBytes = localContext.transferredBytes!! + downloadedBytes
+
+                                val bytes = getObjectResponse.body?.toByteArray() ?: byteArrayOf()
+                                bufferCount.acquire()
+                                buffer.send(
+                                    Part(bytes, i),
+                                )
+                            }
+                        }
+                    }
```
ianbotsf: **Question**: Why is `downloadedBytes` a `var`? It seems like we only set it in one place and it could be a `val`:

```kotlin
val downloadedBytes = executePhase(TransferPhase.BytesTransferred, localContext, interceptors) {
    s3Client.withTmBusinessMetric {
        it.getObject(localContext.s3Request as GetObjectRequest) { getObjectResponse ->
            localContext.s3Response = getObjectResponse
            localContext.transferredBytes = localContext.transferredBytes!! + getObjectResponse.contentLength!!

            val bytes = getObjectResponse.body?.toByteArray() ?: byteArrayOf()
            bufferCount.acquire()
            buffer.send(Part(bytes, i))

            getObjectResponse.contentLength
        }
    }
}
```

ianbotsf: **Correctness**: This `bufferCount.acquire()` call happens after the bytes are already loaded into memory. It must be moved _before_ the `toByteArray` call.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+
+    private suspend fun updateContext(localContext: MutableTransferContext, downloadedBytes: Long) {
+        // Note: coroutines will update the context as they complete, not sequentially based on part number
+        contextMutex.withLock {
+            // It doesn't matter what part the latest req/resp corresponds to
+            context.s3Request = localContext.s3Request
+            context.s3Response = localContext.s3Response
+
+            // This value should remain the same throughout the transfer unless a user changes it using an interceptor
+            context.transferableBytes = localContext.transferableBytes
```
ianbotsf, line 200: **Question**: What's the use case for allowing a caller to change `transferableBytes` in an interceptor? Wouldn't that cause the transfer to fail?

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+            buffer.consumeEach { bufferCount.release() } // Clears buffer and resets count if something went wrong, or no op if success
+        }
+    }
+
+    private suspend fun downloadParts(
+        buffer: Channel<Part>,
+    ) = coroutineScope {
+        val jobs = mutableSetOf<Job>()
+
+        repeat(partCount!! - 1) { i ->
```
ianbotsf: **Style**: There's no reason to use a mutable collection with a loop. Prefer `map` instead:

```kotlin
val jobs = (2..partCount).map { i ->
    launch {
        ...
    }
}
```

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+ */
+
+package aws.sdk.kotlin.hll.s3transfermanager.operations.downloadobject.phases
+
+import aws.sdk.kotlin.hll.s3transfermanager.interceptors.MutableTransferContext
+import aws.sdk.kotlin.hll.s3transfermanager.interceptors.TransferInterceptor
+import aws.sdk.kotlin.hll.s3transfermanager.interceptors.TransferPhase
+import aws.sdk.kotlin.hll.s3transfermanager.interceptors.executePhase
+import aws.sdk.kotlin.hll.s3transfermanager.model.MultipartDownloadType
+import aws.sdk.kotlin.hll.s3transfermanager.model.utils.toDownloadObjectResponse
```
ianbotsf, line 14: **Question**: Where is this `toDownloadObjectResponse` function defined? I don't see the definition in this PR nor in [**s3-tm-main/hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/utils**](https://github.com/aws/aws-sdk-kotlin/tree/2493670155f3f484f4055f00c7816463c1a2ced2/hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/utils).

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/S3TransferManager.kt
```diff
@@ -50,20 +53,28 @@ public class S3TransferManager private constructor(public val s3Client: S3Client
...
-    public val maxConcurrentPartUploads: Int = builder.maxConcurrentPartUploads
+    public val maxConcurrentNetworkOperations: Int = builder.maxConcurrentNetworkOperations
+
+    /**
+     * The maximum number of concurrent file read/write operations.
+     * The actual number of operations executing at any given time may be less than or equal but never greater.
+     *
+     * Defaults to 5.
+     */
+    public val maxConcurrentDiskOperations: Int = builder.maxConcurrentDiskOperations
```
ianbotsf: **Nit**: We should not call these "disk" operations. They're more accurately "file system" operations, which may be backed by a disk, system memory, network storage, etc.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+    private val context: MutableTransferContext,
+    private val s3Client: S3Client,
+    private val downloadPath: String?,
+    private val interceptors: List<TransferInterceptor>,
+    private val objectHandler: (suspend (ByteArray) -> T)?,
+    private val networkOperation: Semaphore,
+    private val diskOperation: Semaphore,
+    private val maxInMemoryParts: Int,
+    private val targetPartSizeBytes: Long,
+    private val bufferCount: Semaphore,
```
ianbotsf: **Nit**: `bufferCount` → `bufferSemaphore` (to match the name used elsewhere)

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+
+    suspend fun transfer() {
+        transferFirstPart()
+        if (needMoreParts) {
+            transferRemainingParts()
+        }
+        buildResponse()
+    }
+
+    private suspend fun transferFirstPart() {
```
ianbotsf: **Correctness**: This function never acquires a permit from `bufferCount` and so will violate the `maxInMemoryParts` constraint.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+                try {
+                    diskOperation.acquire()
+                    downloadPath?.let {
+                        system.write(
+                            tempDownloadPath,
+                            part.bytes,
+                            WriteType.OFFSET(targetPartSizeBytes * (part.part + 1)), // +1 since first part is already written
+                        )
+                    }
+                    bufferCount.release()
```
ianbotsf: **Correctness**: At the moment `bufferCount.release()` executes, the part is still loaded into memory.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+
+    private suspend fun transferRemainingParts() {
+        val buffer = Channel<Part>(capacity = maxInMemoryParts)
+        partCount = partCount ?: ceilDiv(context.transferableBytes!!, targetPartSizeBytes).toInt()
+
+        try {
+            coroutineScope {
+                launch { downloadParts(buffer) }
+                launch { processParts(buffer) }
+            }
```
ianbotsf: **Question**: What's the reason to launch separate coroutines for downloading and processing?

One of the drawbacks to this approach is that it requires different coroutines to acquire and release a permit for a given part, which leaves a lot of room for subtle concurrency errors.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+    /**
+     * Concurrent coroutines will be updating a shared context using this mutex.
+     */
+    private val contextMutex = Mutex()
+
+    /**
+     * If a download path is provided, the object will be downloaded to a temporary file and then renamed to the
+     * provided path at the end.
+     */
+    private val tempDownloadPath = "$downloadPath.s3tmp.${Random.nextInt(0, 10_000_000)}"
```
ianbotsf, line 65: **Question**: All file systems impose a maximum path length and maximum file length, either of which may be exceeded by concatenating onto `downloadPath`. Why do we need to base the temp name on `downloadPath` instead of simply using a random file name of short length?

Additionally, what if `tempDownloadPath` matches the name of `downloadPath` for some other object in the transfer or another file already in the directory? Won't that file be overwritten?

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+
+    /**
+     * Will be used to verify we're receiving the same version of the object for each part.
+     */
+    private lateinit var etag: String
```
ianbotsf: **Nit**: `etag` → `eTag`

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+     */
+    private var needMoreParts: Boolean = true
+
+    suspend fun transfer() {
+        transferFirstPart()
+        if (needMoreParts) {
+            transferRemainingParts()
+        }
+        buildResponse()
+    }
```
ianbotsf: **Question**: Could we avoid a mutable `var` with a possibly-inaccurate starting value by returning the boolean from `transferFirstPart`?

```kotlin
suspend fun transfer() {
    val needMoreParts = transferFirstPart()
    if (needMoreParts) {
        transferRemainingParts()
    }
    buildResponse()
}

private suspend fun transferFirstPart(): Boolean {
    ...
    return !(partCount == 1 || context.transferredBytes == context.transferableBytes)
}
```

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+                                    WriteType.OVERWRITE,
+                                )
+                            }
+                        } finally {
+                            diskOperation.release()
+                        }
+
+                        if (partCount == 1 || context.transferredBytes == context.transferableBytes) {
+                            needMoreParts = false
+                            renameTempFile()
```
ianbotsf: **Nit** `renameTempFile()` is called from both `transferFirstPart` and `transferRemainingParts` under different conditions. It could be moved to `transfer` to simplify.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+        for (part in buffer) {
+            processors += launch {
+                objectHandler?.invoke(part.bytes)
+                try {
+                    diskOperation.acquire()
+                    downloadPath?.let {
+                        system.write(
+                            tempDownloadPath,
+                            part.bytes,
+                            WriteType.OFFSET(targetPartSizeBytes * (part.part + 1)), // +1 since first part is already written
```
ianbotsf: **Correctness**: This offset is only valid for range downloads. In the case of part downloads, the sizes are not guaranteed to be `targetPartSizeBytes`, nor are they guaranteed to be the same for each part.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/DownloadObject.kt
```diff
@@ -0,0 +1,77 @@
...
+        transferredBytes = 0L,
+        s3Request = when (multipartDownloadType) {
+            MultipartDownloadType.Part -> downloadObjectRequest.toGetObjectRequest(partNumber = 1)
+            MultipartDownloadType.Range -> downloadObjectRequest.toGetObjectRequest(range = "bytes=0-${targetPartSizeBytes - 1}")
+        },
+    )
+
+    if (objectHandler == null && downloadPath == null) {
+        throw S3TransferManagerException("Please specify what to do with the downloaded object by setting a download path or an object handler.")
+    }
```
ianbotsf, line 47: **Question**: Could we use a union type to enforce that strictly one of `objectHandler` and `downloadPath` are set?

```kotlin
sealed interface DownloadSink {
    public data class FileSystem(val path: String) : DownloadSink
    public data class Handler<T>(val partHandler: suspend (ByteArray) -> T) : DownloadSink
}
```

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+/**
+ * Orchestrates downloading an S3 object, potentially in multiple parts, to a local file **and/or to a handler function**.
+ */
+internal class TransferBytes<T>(
+    private val multipartDownloadType: MultipartDownloadType,
+    private val context: MutableTransferContext,
+    private val s3Client: S3Client,
+    private val downloadPath: String?,
+    private val interceptors: List<TransferInterceptor>,
+    private val objectHandler: (suspend (ByteArray) -> T)?,
```
ianbotsf: **Correctness**: This handler signature doesn't provide callers the part number or any way to figure out how it fits into the overall object. With concurrency, parts may be processed out of order, causing `objectHandler` to be invoked for later parts _before_ being invoked for earlier parts.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -0,0 +1,288 @@
...
+    private val networkOperation: Semaphore,
+    private val diskOperation: Semaphore,
+    private val maxInMemoryParts: Int,
+    private val targetPartSizeBytes: Long,
+    private val bufferCount: Semaphore,
+) {
```
ianbotsf, line 50: **Style**: This many parameters is a code smell and an indication that this single class is likely doing too much. Please refactor to pass less parameters and to simplify the responsibilities of this class.

ianbotsf (changes requested):
**Question**: I think I expected more of the transfer code to be reused between upload and download. Have you explored options to make `TransferBytes` more generic so that it could move data to/from arbitrary sources?

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
0marperez: It's defined here:
https://github.com/aws/aws-sdk-kotlin/blob/b1ccdbdab8e0513a9d5df2c013f9ff4f3e7b7143/hll/s3-transfer-manager-codegen/src/main/kotlin/aws/sdk/kotlin/hll/s3transfermanager/codegen/mappings/downloadobject/Converters.kt#L51-L107

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
0marperez: Just personal preference / style really

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
0marperez: This is part of the spec (it's a SHOULD so we can do something else if you want)

It would but it's unlikely, we can use UUIDs instead

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
0marperez: In practice never but it's nullable because of the model

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
0marperez: I think this should always be present but same thing as above it's nullable because of the model

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
0marperez: The download path can be null if the user only provided a handler. I can fix the permit issue.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
0marperez: When doing a ranged get or if the object wasn't uploaded in parts then `partCount` is null

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
0marperez: I want them to run at the same time so that's why they're in different coroutines

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
0marperez: The most important thing here is the get object request. The user is free to modify it, so if they do something weird with it, it might break things, but that's on them. If the user never modifies the request then it doesn't matter.

Users can inspect/modify the response but it's not used by us anywhere.

Only problem is that the interceptor might be out of date when it comes to how many bytes have been transferred but it's part of the problem with running things concurrently. Unless we use a mutex and a single context.

Users can also change how many bytes are transferrable but that shouldn't stop us from fully downloading the full object since we calculate how many parts are needed independent of that.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
0marperez: Not sure why a user would do that. We keep a separate variable where we know the object's total size and we do everything based on that so a change wouldn't fail the transfer but it would result in some weird interceptor contexts.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/DownloadObject.kt
0marperez: I'd like for users to be able to provide both

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
ianbotsf: OK then let me flip the question. If `context.transferredBytes == context.transferableBytes` does it _matter_ what `partCount` is?

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
ianbotsf: Ah right, it's codegenned. Thanks for the pointer!

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
ianbotsf: Each individual part download/processing is already a separate coroutine since you call `launch` inside of the loops in `downloadParts` and `processParts`. I'm wondering why they have separate parent coroutines or why downloading/processing even needs to happen in separate coroutines?

This seems like a more straightforward approach:

```kotlin
val jobs = (2..partCount!!).map { partNum ->
    launch {
        bufferSemaphore.withPermit {
            downloadAndProcessPart(partNum)
        }
    }
}
jobs.awaitAll()
```

This eliminates the need for a channel to send parts between coroutines and simplifies semaphore usage for parts in memory. Would this work?

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
ianbotsf: UUIDs would exceed the 8-character limitation imposed by the spec. I'll pursue answers to this question offline and we can leave your file naming code as-is for now.

ianbotsf: Seems like the spec is authoritative in this regard. Your temp file naming code is correct. 👍

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
0marperez: It would work but the current approach with a channel allows downloads to proceed while earlier parts are 
still being written to disk, maximizing network and disk usage.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
0marperez: Agreed, I'd like to defer this refactoring until the remaining TM operations are implemented. Those will likely reveal shared patterns we can extract into common abstractions.

I'll leave a TODO here to track it.

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
ianbotsf: I think a single coroutine for downloading/processing could offer the same benefit. You'd still have all three semaphores, just accessed by a single coroutine:

```
val jobs = (2..partCount!!).map { partNum ->
    launch {
        bufferSemaphore.withPermit {
            val partContext = networkOperation.withPermit {
                downloadPart(partNum) // returns `PartContext`
            }
            fileSystemSemaphore.withPermit {
                processPart(partContext)
            }
        }
    }
}
jobs.awaitAll()
```

This code should allow new downloads to start even if some prior parts are waiting/performing the write to disk, so long as there are sufficient permits left in `bufferSemaphore` and `networkOperation`. Are there any problems I'm missing?

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/operations/downloadobject/phases/TransferBytes.kt
```diff
@@ -205,27 +190,24 @@ internal class TransferBytes<T>(
...
+                            WriteType.OFFSET(part.offset),
                         )
                     }
-                    bufferCount.release()
-                } finally {
-                    diskOperation.release()
                 }
             }
         }
+        bufferSemaphore.release()
```
ianbotsf: **Correctness**: Doesn't this just release a single permit regardless of how many parts were in the buffer?

hll/s3-transfer-manager/common/src/aws/sdk/kotlin/hll/s3transfermanager/model/PartContext.kt
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @param partNumber The part number.
+ * @param bytes The bytes of the part.
+ * @param offset The byte offset of this part within the overall object.
+ */
+public data class PartContext(
+    val partNumber: Int,
+    val bytes: ByteArray,
+    val offset: Long,
+)
```
ianbotsf, line 19: **Correctness**: Data classes are [not suitable for public library APIs](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api) because they cannot be modified without breaking backwards-compatibility. I recommend either making `PartContext` a normal (non-`data`) class or exposing `PartContext` as a public interface and using a private `data class` implementation.

ianbotsf (changes requested):

ianbotsf (approved):
Looks good!

PR #1864, opened 2026-04-20
chore: Fix a flaky Kinesis test by always creating new resources

ianbotsf:
## Issue \#

(none)

## Description of changes

We've had an increasingly-flaky Kinesis E2E test causing problems in CI and merge groups. This change rewrites the test to _always_ create new resources instead of potentially relying on existing ones.


services/kinesis/e2eTest/src/KinesisSubscribeToShardTest.kt
```diff
@@ -34,159 +35,144 @@ private val TEST_DATA = "Bees, bees, bees, bees!"
...
-                        startingPosition = StartingPosition {
-                            type = ShardIteratorType.TrimHorizon
-                        }
-                    },
-                ) {
-                    val event = it.eventStream?.first()
-                    val record = event?.asSubscribeToShardEvent()?.records?.single()
-                    assertEquals(TEST_DATA, record?.data?.decodeToString())
+        val dataStreamArn = client.createStream(testStreamName)
+        val dataStreamConsumerArn = client.registerStreamConsumerAndWait(testConsumerName, dataStreamArn)
```
luigi617: If `dataStreamArn` is assigned and `client.registerStreamConsumerAndWait` fails, then `dataStreamArn` won't be cleaned up as the test fails directly. I think `@BeforeTest`/`@AfterTest` can handle it, something like

```
 private lateinit var dataStreamArn: String                                                                                                                                                                                                                                                                                                                              
  private var dataStreamConsumerArn: String? = null                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                          
  @BeforeTest                                                                                                                                                                                                                                                                                                                                                             
  fun setup() = runBlocking {                                                                                                                                                                                                                                                                                                                                             
      dataStreamArn = client.createStream(testStreamName)                                                                                                                                                                                                                                                                                                                 
      dataStreamConsumerArn = client.registerStreamConsumerAndWait(...)                                                                                                                                                                                                                                                                                                   
  }                                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                          
  @AfterTest                                                                                                                                                                                                                                                                                                                                                              
  fun cleanup() = runBlocking {                                                                                                                                                                                                                                                                                                                                           
      dataStreamConsumerArn?.let {                                                                                                                                                                                                                                                                                                                                        
          client.deregisterStreamConsumer { consumerArn = it; streamArn = dataStreamArn }                                                                                                                                                                                                                                                                                 
      }                                                                                                                                                                                                                                                                                                                                                                   
      if (::dataStreamArn.isInitialized) {                                                                                                                                                                                                                                                                                                                                
          client.deleteStream { streamArn = dataStreamArn }                                                                                                                                                                                                                                                                                                               
          client.waitUntilStreamNotExists { streamArn = dataStreamArn }                                                                                                                                                                                                                                                                                                   
      }                                                                                                                                                                                                                                                                                                                                                                   
  } 
```

or a nested `try/finally` would work as well

services/kinesis/e2eTest/src/KinesisSubscribeToShardTest.kt
```diff
@@ -34,159 +35,144 @@ private val TEST_DATA = "Bees, bees, bees, bees!"
...
+    }
 
-        val newStreamArn = waitUntilStreamExists({ streamName = randomStreamName })
-            .getOrThrow()
-            .streamDescription!!
-            .streamArn!!
+    val newStreamArn = waitUntilStreamExists { streamName = name }
+        .getOrThrow()
+        .streamDescription!!
+        .streamArn
```
luigi617, line 114: Why `!!` is dropped? I didn't find where `null` is handled or is it a non-null variable?

luigi617 (changes requested):

services/kinesis/e2eTest/src/KinesisSubscribeToShardTest.kt
ianbotsf: It's a non-null property. The compiler was emitting a warning about the unnecessary `!!`.

services/kinesis/e2eTest/src/KinesisSubscribeToShardTest.kt
ianbotsf: You're right, I'll go with nested `try`/`finally`.

I moved away from `@BeforeAll` and `@AfterAll` because:
* I needed to execute this test in a loop to reproduce the failure (it happened about ~20% of the time for me). Having separate methods for setup, execution, and teardown made it difficult to write the loop.
* There's only one test in this file so these resources weren't being shared by other tests, which is the point of `@BeforeAll`/`@AfterAll`
* I'm suspicious of mutable data and late-bound references. I prefer the predictability of immutable `val`s whenever possible.

`@BeforeAll`/`@AfterAll` definitely have good uses but this test didn't seem like one of them.

0marperez (approved):

luigi617 (approved):

PR #1877, opened 2026-04-26
feat: write cached auth files with `600` POSIX perms

ianbotsf:
## Description of changes

This change begins writing cached auth token files with permissions `600`—user read/write—on POSIX OSes (e.g., Linux and Mac). 

This is a **behavioral change** so a new configuration property is added (`AWS_CACHED_AUTH_FILE_PERMISSIONS` et al.) allowing users to opt back into the old behavior (i.e., `OS_DEFAULT`—use OS-default permissions for new files in the given directory). Otherwise, the default behavior is `USER_READ_WRITE`.

**Upstream PR**: https://github.com/smithy-lang/smithy-kotlin/pull/1578


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CachedAuthFilePermissions.kt
```diff
@@ -0,0 +1,46 @@
...
+import aws.sdk.kotlin.runtime.config.AwsSdkSetting
+import aws.sdk.kotlin.runtime.config.profile.AwsProfile
+import aws.sdk.kotlin.runtime.config.profile.cachedAuthFilePermissions
+import aws.smithy.kotlin.runtime.config.resolve
+import aws.smithy.kotlin.runtime.util.LazyAsyncValue
+import aws.smithy.kotlin.runtime.util.PlatformProvider
+
+/**
+ * Identifies a set of permissions to apply when writing cached auth files (e.g., during cache token refresh for SSO or
+ * AWS Login credentials). These values have no effect on Windows.
```
luigi617: Question: How permission is provided for windows? or We know that cashed auth files won't never be used in window

luigi617 (approved):
Leave a question in the comment to understand the context, but it looks good to me

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CachedAuthFilePermissions.kt
ianbotsf: Users on Windows get OS-default permissions. This was an intentional choice by the specification.

PR #1883, opened 2026-04-28
Retry standard strategy

luigi617:
## Description of changes
- Introduced `AWS_NEW_RETRIES_2026` flag for enabling new standard retry strategy.
- Enabled new standard retry strategy.
- Preserved old standard strategy. 
- Added unit tests


aws-runtime/aws-config/api/aws-config.api
```diff
@@ -595,8 +596,10 @@ public final class aws/sdk/kotlin/runtime/config/profile/ConfigSection {
 }
 
 public final class aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategyKt {
-	public static final fun resolveRetryStrategy (Laws/smithy/kotlin/runtime/util/PlatformProvider;Laws/smithy/kotlin/runtime/util/LazyAsyncValue;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
-	public static synthetic fun resolveRetryStrategy$default (Laws/smithy/kotlin/runtime/util/PlatformProvider;Laws/smithy/kotlin/runtime/util/LazyAsyncValue;Lkotlin/coroutines/Continuation;ILjava/lang/Object;)Ljava/lang/Object;
+	public static final fun configureRetryDefaults (Laws/smithy/kotlin/runtime/retries/StandardRetryStrategy$Config$Builder;Ljava/lang/String;Ljava/lang/Integer;Z)V
```
luigi617: Not sure why this is a api break as these functions uses `InternalSdkApi`

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/AwsHttpSetting.kt
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Settings resolved from environment variables and system properties for the AWS HTTP runtime.
+ */
+@InternalSdkApi
+public object AwsHttpSetting {
+    /**
+     * Enables the new retry behavior. When set, takes precedence over
+     * the `SMITHY_NEW_RETRIES_2026` environment variable defined in smithy-kotlin.
+     */
+    public val AwsNewRetries: EnvironmentSetting<Boolean> = boolEnvSetting("aws.newRetries2026", "AWS_NEW_RETRIES_2026")
```
luigi617: both **aws-http** (for error type) and **aws-config** (for resolving retry strategy) needs these flag, so added in **aws-http** as **aws-config** depends on this.

gradle/libs.versions.toml
```diff
@@ -22,6 +22,7 @@ ddb-local-version = "2.6.1"
 kotest-version = "5.9.1"
 kotlinx-benchmark-version = "0.4.16"
 kotlinx-serialization-version = "1.11.0"
+kaml-version = "0.55.0"
```
luigi617, line 25: Used for DynamoDB specific test as its variable is defined in **aws-sdk-kotlin**

aws-runtime/aws-config/api/aws-config.api
luigi617: In order to make it not api breaking, I think we can add a new function with serviceName and preserve the old `resolveRetryStrategy ` function, and move the actual implementation to another function that will be called. E.g. 

```
@InternalSdkApi
public suspend fun resolveRetryStrategy(
    platformProvider: PlatformProvider = PlatformProvider.System,
    profile: LazyAsyncValue<AwsProfile> = asyncLazy { loadAwsSharedConfig(platformProvider).activeProfile },
): RetryStrategy = resolveRetryStrategyImpl(platformProvider, profile, serviceName = null)

@InternalSdkApi
public suspend fun resolveRetryStrategy(
    platformProvider: PlatformProvider = PlatformProvider.System,
    profile: LazyAsyncValue<AwsProfile> = asyncLazy { loadAwsSharedConfig(platformProvider).activeProfile },
    serviceName: String,
): RetryStrategy = resolveRetryStrategyImpl(platformProvider, profile, serviceName)

private suspend fun resolveRetryStrategyImpl(
    platformProvider: PlatformProvider,
    profile: LazyAsyncValue<AwsProfile>,
    serviceName: String?,
): RetryStrategy {
...
}
```

ianbotsf: Regarding the "why is this breaking" question, discussed offline but the summary is:

`@InternalApi` and `@InternalSdkApi` are opt-in annotations we use to ensure users understand that these methods are `public` because they _have_ to be shared across modules but they are _not_ intended for public use. That said, these methods _are_ `public` and that means inter-module calls are susceptible to runtime failures (e.g., [`NoSuchMethodError`](https://docs.oracle.com/javase/8/docs/api/java/lang/NoSuchMethodError.html)) if the modules differ in versions. We must try to avoid breaking binary compatibility on these outside of a minor version bump.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
```diff
@@ -43,11 +64,41 @@ public suspend fun resolveRetryStrategy(
...
+        timeoutRetryCost = STANDARD_THROTTLING_RETRY_COST
+    }
+
+    maxAttempts = configuredMaxAttempts
+        ?: if (isDynamoDb) {
+            DYNAMODB_DEFAULT_MAX_ATTEMPTS
+        } else {
+            StandardRetryStrategy.Config.DEFAULT_MAX_ATTEMPTS
+        }
 }
```
ianbotsf: **Correctness**: This still couples the retry strategy resolution to particular services, which makes it a leaky abstraction. The better way to do this is to customize the codegen for these two services to pre-initialize some retry strategy defaults before passing it to this resolver method.

The order of operations should look roughly like:

1. Generated client builder code (e.g., `DynamoDbClient.Config.Builder`) initializes a default retry strategy config
   * For most services there are no customizations made to the strategy config
   * For DynamoDB services, update the max attempts and initial delay
2. Then resolve the relevant environmental configuration parameters (e.g., `AWS_MAX_ATTEMPTS`)
   * If any parameters have been customized then update the config accordingly
   * If a parameter _hasn't_ been customized then don't touch the corresponding config value—it should still be the default initialized by generated client code
3. Then execute the config builder block
   * If users specify a custom value for any config that overrides the defaults/resolved config
   * If a user does not customize a config value then it will still be the default/resolved value

This offers a clean and separated architecture that keeps customizations and edge cases away from the central abstractions (e.g., `StandardRetryStrategy`) and closer to their actual usage.

aws-runtime/aws-config/api/aws-config.api
ianbotsf: In general, the pattern you identified is the correct way to make backwards-compatible changes—preserve the existing method and add a new overload with the new parameter. Typically we'd also then decide on a deprecation strategy for the original method so that eventually we'd be back to the clean state of having only the single desired implementation.

As I called out in another comment, though, I don't think relying on service name inside the central resolver method is correct so I don't believe we actually need to change this method's signature presently.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
```diff
@@ -20,19 +22,38 @@ import aws.smithy.kotlin.runtime.retries.StandardRetryStrategy
...
+ * If `AWS_NEW_RETRIES_2026` is set to `true`, the standard retry strategy behavior is enabled.
+ * Falls back to smithy-kotlin's `SMITHY_NEW_RETRIES_2026`.
  */
 @InternalSdkApi
 public suspend fun resolveRetryStrategy(
     platformProvider: PlatformProvider = PlatformProvider.System,
     profile: LazyAsyncValue<AwsProfile> = asyncLazy { loadAwsSharedConfig(platformProvider).activeProfile },
+    serviceName: String? = null,
 ): RetryStrategy {
+    val useNewRetries = AwsHttpSetting.AwsNewRetries.resolve(platformProvider) ?: CoreSettings.resolveNewRetriesEnabled()
```
ianbotsf: **Nit**: Pass `platformProvider` to `CoreSettings.resolveNewRetriesEnabled()`

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
```diff
@@ -43,11 +64,41 @@ public suspend fun resolveRetryStrategy(
...
+        configureRetryDefaults(serviceName, maxAttempts, useNewRetries)
+    }
+}
+
+/**
+ * Configures the retry strategy builder with the resolved max attempts and, when new retries are enabled,
+ * all standard defaults as defined in the New Retry Behavior.
+ */
+@InternalSdkApi
+public fun StandardRetryStrategy.Config.Builder.configureRetryDefaults(
```
ianbotsf: **Nit**: This method is only called from `resolveRetryStrategy` in the same file so it should be `internal` or `private`.

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/retries/AwsRetryPolicy.kt
```diff
@@ -43,9 +48,20 @@ import aws.smithy.kotlin.runtime.retries.policy.StandardRetryPolicy
...
  * If none of those conditions match, this policy delegates to [StandardRetryPolicy]. See that class's documentation for
  * more information about how it evaluates exceptions.
+ *
+ * @param serviceName The optional sdkId of the service (e.g. `"STS"`).
  */
-public open class AwsRetryPolicy : StandardRetryPolicy() {
+public open class AwsRetryPolicy internal constructor(
+    private val serviceName: String? = null,
+    platformProvider: PlatformEnvironProvider = PlatformProvider.System,
+) : StandardRetryPolicy() {
```
ianbotsf: **Correctness**: As noted elsewhere we should avoid coupling individual service behavior to the cross-service retry policy. If we decide to pursue a service-specific override, we should customize codegen and config resolution as appropriate.

All that said, I see that our current implementation treats `IDPCommunicationError` as retriable regardless of which service is throwing it, which is technically a violation of the spec (both pre-update and post-update). I think that's okay because no other AWS service _throws_ `IDPCommunicationError`. (IAM service can throw `OpenIdIdpCommunicationError` but that's not the same.) I don't think we need to customize how this error is handled.

ianbotsf (changes requested):

aws-runtime/aws-http/common/src/aws/sdk/kotlin/runtime/http/retries/AwsRetryPolicy.kt
luigi617: Great! then the previous error classification is already compliant with the sep

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
luigi617: I agree that it is better to have the default retry strategy config of each serivce. However, `resolveRetryStrategy` still needs to know more information than it had, i.e. we need to pass the default value for each service into this function (only DynamoDB would have values, rest of services pass null). So that  `resolveRetryStrategy` can generate the correct `RetryStrategy`.

Since `RetryStrategy` is immutable, it is possible to build `RetryStrategy` first (one that use always the default value) and then rebuild it so that it set the default value and env parameter value. But the behavior of `resolveRetryStrategy` changes.

As we want to [perserve this function signature](https://github.com/aws/aws-sdk-kotlin/pull/1883#discussion_r3163543482). I thought about the following code but this is essentially the same as [this](https://github.com/aws/aws-sdk-kotlin/pull/1883#discussion_r3162720719):

What do you think?

```
internal data class ResolvedRetryConfig(
    val maxAttempts: Int?,
    val retryMode: RetryMode,
    val useNewRetries: Boolean,
)

internal suspend fun resolveRetryConfig(
    platformProvider: PlatformProvider,
    profile: LazyAsyncValue<AwsProfile>,
): ResolvedRetryConfig {
    ....
    return ResolvedRetryConfig(maxAttempts, retryMode, useNewRetries)
}

@InternalSdkApi
public suspend fun resolveRetryStrategy(
    platformProvider: PlatformProvider = PlatformProvider.System,
    profile: LazyAsyncValue<AwsProfile> = asyncLazy { loadAwsSharedConfig(platformProvider).activeProfile },
): RetryStrategy {
    val config = resolveRetryConfig(platformProvider, profile)
    return buildRetryStrategy(config)
}

internal fun buildRetryStrategy(config: ResolvedRetryConfig): RetryStrategy =
    buildRetryStrategy(config, defaultMaxAttempts = null, defaultInitialDelay = null)

internal fun buildRetryStrategy(
    config: ResolvedRetryConfig,
    defaultMaxAttempts: Int?,
    defaultInitialDelay: Duration?,
): RetryStrategy {
    val factory = when (config.retryMode) {
        RetryMode.STANDARD, RetryMode.LEGACY -> StandardRetryStrategy
        RetryMode.ADAPTIVE -> AdaptiveRetryStrategy
    }

    return factory {
        configureRetryDefaults(
            configuredMaxAttempts = config.maxAttempts,
            useNewRetries = config.useNewRetries,
            defaultMaxAttempts = defaultMaxAttempts,
            defaultInitialDelay = defaultInitialDelay,
        )
    }
}

@InternalSdkApi
public fun StandardRetryStrategy.Config.Builder.configureRetryDefaults(
    configuredMaxAttempts: Int? = null,
    useNewRetries: Boolean = false,
    defaultMaxAttempts: Int? = null,
    defaultInitialDelay: Duration? = null,
) {
    if (!useNewRetries) {
        configuredMaxAttempts?.let { maxAttempts = it }
        return
    }

    delayProvider {
        initialDelay = defaultInitialDelay ?: STANDARD_INITIAL_DELAY
        scaleFactor = STANDARD_SCALE_FACTOR
    }

    tokenBucket {
        retryCost = STANDARD_RETRY_COST
        timeoutRetryCost = STANDARD_THROTTLING_RETRY_COST
    }

    maxAttempts = configuredMaxAttempts
        ?: defaultMaxAttempts
        ?: StandardRetryStrategy.Config.DEFAULT_MAX_ATTEMPTS
}

```

ianbotsf: I like the direction of your `resolveRetryConfig` method. Passing around config builders seems better than passing around immutable built objects. That said, `ResolvedRetryConfig` feels a bit redundant to our existing retry strategy config builder. Could we use `RetryStrategy.Config.Builder` instead?

luigi617: `ResolvedRetryConfig` is not redundant because it includes `retryMode` which determines the factory and it is not a field in `RetryStrategy.Config.Builder`. Alternative is to return sth like `Pair<RetryMode, RetryStrategy.Config.Builder>` but is essentially the same having `ResolvedRetryConfig`

ianbotsf: OK then let's go with `ResolvedRetryConfig`. 👍

luigi617: done!

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/dynamodb/DynamoDbRetryDefaultsIntegration.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    override val sectionWriters: List<SectionWriterBinding>
+        get() = listOf(
+            SectionWriterBinding(
+                ServiceClientGenerator.Sections.CompanionObject,
+                AppendingSectionWriter { writer ->
+                    writer.write("")
+                    writer.write("override val defaultMaxAttempts: Int = $DYNAMODB_MAX_ATTEMPTS")
+                    writer.write("override val defaultInitialDelay: #T = $DYNAMODB_INITIAL_DELAY_MS.#T", KotlinTypes.Time.Duration, KotlinTypes.Time.milliseconds)
+                },
```
ianbotsf, line 47: **Style**: Do not mix Kotlin string substitution and Smithy string substitution in the same command. Always use one or the other...preferably Smithy string substitution:

```kotlin
writer.write("override val defaultMaxAttempts: Int = #L", DYNAMODB_MAX_ATTEMPTS)
writer.write(
    "override val defaultInitialDelay: #T = #L.#T",
    KotlinTypes.Time.Duration,
    DYNAMODB_INITIAL_DELAY_MS,
    KotlinTypes.Time.milliseconds,
)
```

ianbotsf (changes requested):
Please add tests for the new env var and system property.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/dynamodb/DynamoDbRetryDefaultsIntegration.kt
luigi617: my bad, I forgot the `#L`

ianbotsf (approved):

PR #1891, opened 2026-04-30
chore: enable customization of Dokka build via Gradle properties

ianbotsf:
## Description of changes

This change parameterizes Dokka's isolation mode via Gradle properties. This allows for easier testing of Dokka performance without modifying source code to set memory levels, isolation levels, etc.


buildSrc/src/main/kotlin/dokka-convention.gradle.kts
```diff
@@ -52,16 +56,24 @@ dokka {
...
+                val smithyKotlinPackageList = properties["dokka.smithyKotlin.packageList"]?.toString()
+                if (smithyKotlinPackageList != null) {
+                    packageListUrl(smithyKotlinPackageList)
+                }
             }
         }
     }
+
+    dokkaPublications.html {
+        val offline = properties["dokka.offlineMode"]?.toString()?.lowercase()?.toBooleanStrict() ?: false
```
luigi617, line 75: If someone sets `dokka.offlineMode=yes or 1`, this throws an `IllegalArgumentException` instead of falling back gracefully. Is this intentional? I think it is bettere to use `toBoolean()` or `toBooleanStrictOrNull() ?: false`

buildSrc/src/main/kotlin/dokka-convention.gradle.kts
```diff
@@ -13,8 +13,12 @@ dokka {
...
     moduleVersion.set(sdkVersion)
 
-    dokkaGeneratorIsolation = ProcessIsolation {
-        maxHeapSize = "16g"
+    val isolationMode = properties["dokka.isolationMode"]?.toString()?.lowercase() ?: "process"
+    val heapSize = properties["dokka.maxHeapSize"]?.toString() ?: "16g"
+
+    dokkaGeneratorIsolation = when (isolationMode) {
+        "classloader" -> ClassLoaderIsolation { }
+        else -> ProcessIsolation { maxHeapSize = heapSize }
```
luigi617: If `isolationMode` has any value different from `classloader` it would silently falls through to `ProcessIsolation`, probably something like this is better:
```
"classloader" -> ClassLoaderIsolation { }                                                                                                                                                                                                         
"process" -> ProcessIsolation { maxHeapSize = heapSize }                                                                                                                                                                                          
else -> error("Unknown dokka.isolationMode: $isolationMode (expected 'process' or 'classloader')") 
```

buildSrc/src/main/kotlin/dokka-convention.gradle.kts
```diff
@@ -52,16 +56,24 @@ dokka {
...
-    // Configure Dokka to link to smithy-kotlin types
-    dokkaSourceSets.configureEach {
+        // Configure Dokka to link to smithy-kotlin types
         externalDocumentationLinks {
             create("smithy-kotlin") {
                 url("https://docs.aws.amazon.com/smithy-kotlin/api/latest")
+
+                val smithyKotlinPackageList = properties["dokka.smithyKotlin.packageList"]?.toString()
+                if (smithyKotlinPackageList != null) {
+                    packageListUrl(smithyKotlinPackageList)
```
luigi617: Question: `smithyKotlinPackageList` is surely expected to be url right? This name `smithyKotlinPackageList` does not seem like an url

buildSrc/src/main/kotlin/dokka-convention.gradle.kts
```diff
@@ -52,16 +56,24 @@ dokka {
             "OPERATIONS.md", // Generated by `kat` tool, all service operations
         ).mapNotNull { project.file(it).takeIf { it.exists() } }
         includes.from(includeFiles)
-    }
 
-    // Configure Dokka to link to smithy-kotlin types
-    dokkaSourceSets.configureEach {
+        // Configure Dokka to link to smithy-kotlin types
         externalDocumentationLinks {
```
luigi617, line 62: By looking at Dokka document,`externalDocumentationLinks` are always inside `dokkaSourceSets`. But now `dokkaSourceSets` is deleted. What is the new behavior?

luigi617 (changes requested):

buildSrc/src/main/kotlin/dokka-convention.gradle.kts
ianbotsf: Yes, this is intentional. We should eagerly fail on configuration errors because we cannot accurately determine the user's intent.

When a user does not pass a flag, it's clear they want the default behavior. When the user does pass a flag and the value is valid (in this case `true` or `false`, case-insensitive), it's clear they want specific (possibly non-default) behavior. When the user specifies an _invalid_ value, their intent is no longer clear. They definitely _wanted_ some specific behavior but we cannot determine which—`true` or `false`. Silently falling back to the default gives the user the erroneous impression that their config was successfully applied.

Ironically, I did not follow this pattern for `isolationMode` as you correctly pointed out.

buildSrc/src/main/kotlin/dokka-convention.gradle.kts
ianbotsf: You are absolutely correct. I'll use an `else` case to catch situations where someone chooses an invalid option.

buildSrc/src/main/kotlin/dokka-convention.gradle.kts
ianbotsf: Will change.

buildSrc/src/main/kotlin/dokka-convention.gradle.kts
ianbotsf: This is still happening inside the `dokkaSourceSets.configureEach` block started on L41. I removed the close of the previous block and the start of this block because we didn't need two blocks.

**Before**:

```kotlin
dokkaSourceSets.configureEach {
    // Configure samples
    // Configure includes
}

dokkaSourceSets.configureEach {
    // Configure external doc links
}
```

**After**:

```kotlin
dokkaSourceSets.configureEach {
    // Configure samples
    // Configure includes
    // Configure external doc links
}
```

luigi617 (approved):

PR #1897, opened 2026-05-07
feat: generate specialized, single-attribute key extension methods for operations

ianbotsf:
## Issue \#

Related to #472 

## Description of changes

This change adds codegenned extension methods on operations over single-attribute keys.

### Background

Keys are modeled specially in DynamoDB Mapper. Namely, they must be a subtype of [`KeyType`](https://github.com/aws/aws-sdk-kotlin/blob/e3ebeff8d6d8cd8377fa9d75c6bd80afd95e18c4/hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/items/KeyType.kt#L38) such as `KeyType.Key1<K1>`, `KeyType.Key2<K1, K2>`, etc., where each <code>K<em>n</em></code> is one of `ByteArray`, `Byte`, `Int`, `Long`, `Short`, or `String`.

This key modeling is used heavily within schemas and converters. For example, the schema for a table with a string partition key and an integer sort key would have the following type signature:

```kotlin
ItemSchema.CompositeKey<
    MyDataType,            // The mapped type (irrelevant in this example)
    KeyType.Key1<String>,  // The partition key consists of a single String attribute
    KeyType.Key1<Int>,     // The sort key consists of a single Int attribute
>
```

These key types come into play when interacting with operations which receive/return keys, such as `deleteItem`:

```kotlin
val schema: ItemSchema.CompositeKey<MyDataType, KeyType.Key1<String>, KeyType.Key1<Int>> = …

// table: Table.CompositeKey<MyDataType, KeyType.Key1<String>, KeyType.Key1<Int>>
val table = mapper.getTable(myTableName, schema)

table.deleteItem {
    // partitionKey: KeyType.Key1<String>
    partitionKey = Key("Foo")

    // sortKey: KeyType.Key1<Int>
    sortKey = Key(42)
}
```

### Ergonomics

The preceding method/property signatures carefully preserve type information and safety throughout DynamoDB Mapper and eliminate the need for runtime reflection or dynamic typing. That said, dealing with key values themselves can be a bit awkward (e.g., `Key("Foo")` instead of just `"Foo"`).

To resolve that issue, this PR now generates extension methods for every combination of single-attribute key types on a table. These extension methods delegate to their appropriate canonical request-accepting method but provide cleaner signatures which enable simpler call patterns:

```kotlin
table.deleteItem("Foo", 42)
```

As with canonical request-accepting methods, these key-specific extension methods are also type-safe. For instance, `deleteItem(partitionKey: String, sortKey: Int)` is only defined on a receiver of `TableOperations.CompositeKey<*, KeyType.Key1<String>, KeyType.Key1<Int>>`, preventing consumers from accidentally passing a byte array or a non-`Int` sort key.

### Rationale

Single-attribute key fields are the most common types of key structures in DynamoDB. Support for multi-attribute key fields is relatively new and not yet widely adopted or required. Consequently, most consumers will interact with tables/indexes that use a single-attribute partition key and possibly a single-attribute sort key. For this reason, providing simplifying extension functions cleans up user code and makes the intent clearer.

At present there are 6 accepted key types (`ByteArray`, `Byte`, `Int`, `Long`, `Short`, or `String`), leading to 6 overloads for operations on a single-attribute partition key with no sort key. For tables that additionally have a sort key, each of the 6 possible partition key types are combined with each of the 6 possible sort key types, leading to an additional 36 overloads. Thus, each operation now gains 6 + 36 = 42 overloads.

If we were to provide extension methods for both single-attr and dual-attr keys, then the following overloads would be generated per operation:

| Partition key | Sort key    | Overloads |
|:--------------|:------------|----------:|
| single-attr   | none        |         6 |
| single-attr   | single-attr |        36 |
| single-attr   | dual-attr   |       216 |
| dual-attr     | none        |        36 |
| dual-attr     | single-attr |       216 |
| dual-attr     | dual-attr   |     1,296 |
| **Total**     |             | **1,806** |

The artifact size overhead of 1,806 new overloads outweighs the benefits of providing convenience methods for multi-attribute keys. This decision may be revisited in the future.


hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
```diff
@@ -4,14 +4,12 @@
  */
 package aws.sdk.kotlin.hll.dynamodbmapper.codegen.operations.rendering
 
-import aws.sdk.kotlin.hll.codegen.model.Operation
-import aws.sdk.kotlin.hll.codegen.model.Type
-import aws.sdk.kotlin.hll.codegen.model.TypeRef
-import aws.sdk.kotlin.hll.codegen.model.genericVars
+import aws.sdk.kotlin.hll.codegen.model.*
```
luigi617, line 7: **Question**: Is `import ...*` a preferred method / better practice than importing only required parts? My impression is importing everything (*) is 1) less readable 2) might be slower due to importing more than needed

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
```diff
@@ -124,20 +139,19 @@ internal class OperationsTypeRenderer(
...
                     write(
                         "override suspend fun #L(request: #T) = #L(spec).execute(request)",
                         op.methodName,
-                        requestProjection.interfaceStruct.type,
+                        request,
                         FactoryRenderer.factoryFunctionName(op),
                     )
 
-                    if (paginationInfo != null) blankLine()
+                    if (isPaginated) blankLine()
```
luigi617, line 153: **Question**: why blank line is needed only when `isPaginated==true`?

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
ianbotsf: We don't have a strong policy or style rule about star imports. You're correct that `import ...*` is less readable and can lead to confusion about where a particular declaration comes from. Star imports do not make runtime code any slower since imports are a source code feature only. All unqualified references are turned into qualified ones by the compiler and the imports don't exist at runtime. Star imports may _slightly_ reduce compilation speed but not enough to matter IMHO.

Typically, I don't bother writing/maintaining the imports myself. The IDE auto-imports everything I need and automatically folds a given namespace into a star import when 5 or more declarations are used. I haven't cared enough to fix/change this but you asking is a good signal to revisit and make a more intentional decision independent of this PR.

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
ianbotsf: It spaces out the operation declarations which require multiple lines vs those that only require one line:

```kotlin
interface TableOperations<T> {
    suspend fun deleteItem(request: DeleteItemRequest.PartitionKey<PK>): DeleteItemResponse<T>
    suspend fun getItem(request: GetItemRequest.PartitionKey<PK>): GetItemResponse<T>
    suspend fun updateItem(request: UpdateItemRequest.PartitionKey<PK>): UpdateItemResponse<T>

    @ManualPagination(paginatedEquivalent = "queryPaginated")
    suspend fun query(request: QueryRequest.PartitionKey<PK>): QueryResponse.PartitionKey<T, PK>

    @ManualPagination(paginatedEquivalent = "scanPaginated")
    suspend fun scan(request: ScanRequest.PartitionKey<PK>): ScanResponse.PartitionKey<T, PK>
}
```

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
```diff
@@ -240,4 +332,22 @@ internal class OperationsTypeRenderer(
         }
 }
 
-private fun Operation.appliesToKindOrAncestor(kind: ItemSourceKind): Boolean = kind in itemSourceKinds || (kind.parent?.let { appliesToKindOrAncestor(it) } ?: false)
+private fun Operation.appliesToKindOrAncestor(
+    kind: ItemSourceKind,
+): Boolean = kind in itemSourceKinds || (kind.parent?.let { appliesToKindOrAncestor(it) } ?: false)
+
+private fun <T, R> List<T>.permutations(length: Int, transform: (List<T>) -> List<R>): List<List<R>> = buildList {
```
luigi617: **nit**: this function is a cartesian product rather than permutation. Not sure which behavior is expected
Cartesian product is sampling with replacement (e.g. `cartesianProduct([1, 2]) = [(1, 1), (1, 2), (2, 1), (2, 2)]`). Whereas permutation is sampling without replacement (e.g. `permutations([1, 2]) = [(1, 2), (2, 1)]`)

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
```diff
@@ -240,4 +332,22 @@ internal class OperationsTypeRenderer(
...
+    val values = this@permutations
+
+    fun generate(current: List<T>) {
+        if (current.length == length) {
```
luigi617: Question: the idiomatic Kotlin approach is `.size`, why using `.length`? and what is the reson of having `.length`?

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
```diff
@@ -179,6 +193,80 @@ internal class OperationsTypeRenderer(
...
+
+            val response = when (pagination) {
+                PaginationType.PAGE_FLOW -> Types.Kotlinx.Coroutines.Flow.flow(genericResponse.replaceGenericArgs())
+                else -> genericResponse.replaceGenericArgs()
+            }
+
+            val modifier = if (pagination == PaginationType.PAGE_FLOW) "" else "suspend "
+
+            if (pagination == PaginationType.SINGLE_PAGE) renderManualPaginationAnnotation(op)
+            withBlock(
```
luigi617, line 247: **Question**: Do we care warnings for unused functions? or in general, do we care warnings caused by codegen?

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
```diff
@@ -179,6 +193,80 @@ internal class OperationsTypeRenderer(
...
+        keyType: KeyProjectionType,
+        pagination: PaginationType,
+    ) {
+        val requestProjection = op.request.keyProjections[keyType]
+        val genericReceiver = itemSourceKind.opsType(keyType)
+        val genericResponse = op.response.keyProjections[keyType].interfaceStruct.type
+
+        val methodName = when (pagination) {
+            PaginationType.PAGE_FLOW -> PaginatorRenderer.paginatorName(op)
+            else -> op.methodName
```
luigi617: **Nit**: it might be better to indicate explicitly the case for else. Not a problem today, but if in the future `PaginationType` needs to be changed, the `else` will silently swallows it.

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
```diff
@@ -179,6 +193,80 @@ internal class OperationsTypeRenderer(
         }
     }
 
+    private fun renderKeyExtensions() {
+        keyedOperations
+            .filterNot { it.appliesToAncestorKind() }
+            .forEach { op ->
+                val paginationType = if (op.isPaginated) PaginationType.SINGLE_PAGE else PaginationType.UNPAGINATED
```
luigi617, line 199: **Nit**: it might be better to indicate explicitly the case for else. Not a problem today, but if in the future `PaginationType` needs to be changed, the `else` will silently swallows it.

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
```diff
@@ -179,6 +193,80 @@ internal class OperationsTypeRenderer(
...
+        }
+
+        val keyedMembers = requestProjection.interfaceStruct.members.filter { it.keyType != null }
+        val keyPermutations = keyDataTypes.permutations(keyedMembers.size) { permutation ->
+            keyedMembers.zip(permutation) { member, dataType ->
+                member.copy(type = dataType)
+            }
+        }
+
+        keyPermutations.forEach { keyPermutation ->
```
luigi617: If `permutations` receives an empty list of members, then `keyPermutations = listOf(listOf())`. Then it will generate something like `table.deleteItem(): Response = deleteItem { }`. How is this case handled? Is it prevented elsewhere?

luigi617 (changes requested):

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
ianbotsf: You're right, this should be called "cartesian product" not "permutations". 👍

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
ianbotsf: Good catch, I always confuse `size` and `length`. I should use `size` here.

As for why [`Collection<T>.length`](https://github.com/smithy-lang/smithy-kotlin/blob/ce7f1e449f76380cded1982c4b2b8a93d9bf31ac/runtime/runtime-core/common/src/aws/smithy/kotlin/runtime/util/JMESPath.kt#L46-L51) exists, it's because [JMESPath](https://jmespath.org/), which is [used by our waiter definitions](https://smithy.io/2.0/additional-specs/waiters.html), has a [`length`](https://jmespath.org/specification.html#length) function which operates over strings, arrays, and objects, returning either the `size` or the `length` depending on the receiver type. For that reason, it was convenient to create a generic `Collection<T>.length` extension to simplify codegen.

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
ianbotsf: Yes we care about unused functions, including in codegen. It's not always practical/desirable to remove unused functions from codegen but we should keep the code as clean as possible. Why do you ask? Is there some function which would be generated which is unused?

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
ianbotsf: `permutations` is applied to `keyDataTypes`, which is statically defined at the top of the file with a list of acceptable key types. The only way `keyPermutations == listOf(listOf())` is if `keyDataTypes == listOf()`, which should never happen but, if it did, the `keyPermutations.forEach` loop would be skipped. We wouldn't generate `table.deleteItem(): Response = deleteItem { }` because we wouldn't generate _any_ extension functions.

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
luigi617: This is just a question that came up to my mind when I was looking at this becuase this effectivly generates multiple overloaded functions that user might not use all of them. So I am not sure if `@Suppress("UNUSED")` is needed. But then I realize it is just a IDE-level annotation and not warning.

ianbotsf: In general, the purpose of a library is to provide a toolbox of functionality for consumers. Users may not use some of the APIs but they could if they wanted and those APIs served a useful purpose. We should avoid polluting our code with unused `private`/`internal`/`@InternalApi` declarations but the `public` ones may very well be unused (for now).

luigi617 (approved):

PR #1911, opened 2026-05-22
Fix dokka memory

luigi617:
## Issue \#

- `Task :dokkaGeneratePublicationHtml`: Runs indefinitely with the default 16g heap. Aggregator projects (`:`, `:services`, `:aws-runtime`, `:hll`) must load and merge all 426 modules in a single process — with insufficient heap, the JVM gets stuck in continuous GC making no forward progress.
- `Task :trimNavigationFiles`: The previous implementation parsed the full `navigation.html` independently per module via Jsoup — 426 redundant parses of a multi-MB file, causing ~150MB+ of concurrent DOM/String allocations and slow execution.
    
## Description of changes
   - Differentiate heap sizes by project role: aggregator projects (`:`, `:services`, `:aws-runtime`, `:hll`) now default to 64g via `dokka.aggregator.maxHeapSize`, leaf modules default to 16g via `dokka.module.maxHeapSize`. Both configurable via Gradle properties for CI tuning.
  - Rewrote `TrimNavigation` from a worker-based N-parse approach (426 independent Jsoup parses) to a single-parse architecture: parse root `navigation.html` once, pre-encode each module's collapsed/expanded HTML to `ByteArray`, then write all 426 output files in parallel via `parallelStream()` using direct byte I/O. Eliminates ~150MB of intermediate String/DOM allocations and the Gradle Worker API overhead.
  - Added diagnostic logging to `TrimNavigation` (parse time, segment count, memory footprint, write time) for observability in CI logs.
  - Pre-compiled the regex in `sdkIdToModelFilename()` which previously created a new `Regex`/`Pattern` object on every invocation (~426 times during codegen discovery).
  - Fixed resource leak in `PackageSettings.fromFile()` where `inputStream()` was never closed — wrapped in `.use { }` to ensure file descriptors are released across 426 service `package.json` files.

## Trade-offs and decisions
  - **Root navigation.html as single source of truth**: Uses the root file as template for all modules. If per-module `navigation.html` files ever diverge from the root, the new approach would overwrite those differences.
  - **parallelStream() thread pool**: Uses the JVM common `ForkJoinPool` (simpler) instead of Gradle's Worker API, so parallelism isn't governed by `--max-workers`. Fine in practice since `trimNavigationFiles` runs after `dokkaGenerate` via `finalizedBy`.

 ## Benchmark results (TrimNavigation)

  Local benchmark with 426 synthetic modules (359 KB navigation.html):

  | Implementation | avg | min | max |
  |---|---|---|---|
  | Old (N parses, Worker API) | 1082ms | 973ms | 1175ms |
  | New (single-parse, byte segments, parallel write) | 148ms | 138ms | 165ms |
  | **Speedup** | **7.3x** | | |

  Note: production `navigation.html` is ~190 MB (larger than the 359 KB synthetic file), so real-world speedup is expected to be even greater.

 ##   Test
  - Successfully ran it multiple times with max-workers=16, dokka.module.maxHeapSize=6g, Pdokka.aggregator.maxHeapSize=84g. However, this number is an upper bound and needs to be finetuned to find lower number that still success (use less resource as possible)
      - max-workers=16, dokka.module.maxHeapSize=8g, Pdokka.aggregator.maxHeapSize=64g fails sometimes
      - max-workers=16, dokka.module.maxHeapSize=8g, Pdokka.aggregator.maxHeapSize=44g fails
      - max-workers=16, dokka.module.maxHeapSize=6g, Pdokka.aggregator.maxHeapSize=64g works
  - Verified locally that both implementations produce byte-for-byte identical output (after HTML normalization) across all 426 modules via a correctness test.


buildSrc/src/main/kotlin/aws/sdk/kotlin/dokka/TrimNavigation.kt
```diff
@@ -29,64 +24,75 @@ abstract class TrimNavigation @Inject constructor(private val workerExecutor: Wo
...
-        val logger = Logging.getLogger(TrimNavigation::class.java)
-        logger.info("Trimming $relativePath...")
+        if (!rootNavFile.exists()) {
+            logger.warn("Root navigation.html not found at $rootNavFile, skipping trim")
+            return
+        }
 
-        val doc = Jsoup.parse(navigation)
+        val rootNavSize = rootNavFile.length() / 1024 / 1024
+        logger.lifecycle("[TrimNavigation] Parsing root navigation.html ($rootNavSize MB)...")
```
ianbotsf: **Nit**: `/` rounds down (i.e., floor division) which will lead to inaccurate reporting. I suggest printing either:
* The full number of bytes with thousands separators (e.g., `String.format("%,d", rootNavSize)`) –**_or_**–
* Megabytes with a few decimal places and proper rounding (e.g., `String.format("%.3f", rootNavSize)`)

buildSrc/src/main/kotlin/aws/sdk/kotlin/dokka/TrimNavigation.kt
```diff
@@ -29,64 +24,75 @@ abstract class TrimNavigation @Inject constructor(private val workerExecutor: Wo
...
+        val parseStart = System.currentTimeMillis()
+        val rootDoc = Jsoup.parse(rootNavFile)
+        logger.lifecycle("[TrimNavigation] Parsed in ${(System.currentTimeMillis() - parseStart) / 1000}s")
```
ianbotsf: **Nit**: Use either [`measureTime`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/measure-time.html) or [`measureTimedValue`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/measure-timed-value.html) instead of `currentTimeMillis`, which is lower precision and returns values as `Long` (vs `Duration`):

```kotlin
val (rootDoc, parseTime) = measureTimedValue {
    Jsoup.parse(rootNavFile)
}
logger.lifecycle("[TrimNavigation] Parsed in $parseTime")
```

buildSrc/src/main/kotlin/aws/sdk/kotlin/dokka/TrimNavigation.kt
```diff
@@ -29,64 +24,75 @@ abstract class TrimNavigation @Inject constructor(private val workerExecutor: Wo
...
+            Segment(id, collapsedPart.outerHtml().toByteArray(), part.outerHtml().toByteArray())
+        }
+
+        val moduleNavFiles: List<Pair<String, File>> = sourceDir.walk()
+            .filter { it.name == "navigation.html" && it.parentFile != sourceDir }
+            .map { it.parentFile.name to it }
+            .toList()
+
+        val totalSegmentBytes = segments.sumOf { it.collapsed.size.toLong() + it.full.size.toLong() } / 1024 / 1024
+        logger.lifecycle("[TrimNavigation] ${segments.size} segments pre-extracted ($totalSegmentBytes MB in memory)")
```
ianbotsf: **Nit**: Same comment about precision loss from floor division.

buildSrc/src/main/kotlin/aws/sdk/kotlin/dokka/TrimNavigation.kt
```diff
@@ -29,64 +24,75 @@ abstract class TrimNavigation @Inject constructor(private val workerExecutor: Wo
...
+                    if (seg.id.startsWith("$moduleName-nav-submenu")) {
+                        out.write(seg.full)
+                    } else {
+                        out.write(seg.collapsed)
+                    }
+                    out.write(newline)
+                }
+                out.write(footer)
             }
+        }
```
ianbotsf: **Question**:

> **parallelStream() thread pool**: Uses the JVM common `ForkJoinPool` (simpler) instead of Gradle's Worker API, so parallelism isn't governed by `--max-workers`. Fine in practice since `trimNavigationFiles` runs after `dokkaGenerate` via `finalizedBy`.

Why is allowing more than `--max-workers` of threads a good thing? It seems to me that respecting the `--max-workers` flag and using Gradle's Worker API is the best way to keep our build idiomatic and well-behaved.

Are there other reasons to drop the Workers API?

buildSrc/src/main/kotlin/aws/sdk/kotlin/dokka/TrimNavigation.kt
```diff
@@ -29,64 +24,75 @@ abstract class TrimNavigation @Inject constructor(private val workerExecutor: Wo
...
 
-        // Update navigation.html
-        val trimmedSideMenuParts = doc.select("div.sideMenu > div.toc--part")
-        navigation.writeText("<div class=\"sideMenu\">\n$trimmedSideMenuParts\n</div>")
+        val writeElapsed = (System.currentTimeMillis() - writeStart) / 1000
+        logger.lifecycle("[TrimNavigation] Done trimming navigation files in ${writeElapsed}s")
```
ianbotsf: **Nit**: Same comment about `measureTime`/`measureTimedValue`.

buildSrc/src/main/kotlin/aws/sdk/kotlin/dokka/TrimNavigation.kt
luigi617: It's not that exceeding `--max-workers` is a good thing, I think that it doesn't matter here.

The old implementation submitted 426 WorkActions, each of which is CPU intensive, so running without `--max-workers` would thrash the CPU.

In the new implementation, since root navigation.html is used as single source of truth , each parallel work doesn't need to recompute everything, it needs only to write, so now it is a I/O-bound. From my understanding `--max-workers` protects resource (CPU cores) that this work doesn't consume. Moreover, since the parallel works are ran temporal isolation via `finalizedBy`, there won't be any other gradle work fighting for CPU cores. In summary, I think Worker API solves a problem that no longer exists and Worker API adds overhead that can be avoided.

ianbotsf: I see a few potential downsides to this decision.

The Gradle Worker API respects `--max-workers` and coordinates access to Gradle resources. All of Gradle's other task work will happen in a thread pool it can manage and understand. Using `ForkJoinPool` breaks out of that model and means our `TrimNavigation` subtasks will use threads already in use by other Gradle tasks. Maybe this isn't a problem today since `TrimNavigation` is one of the last tasks to execute but that's a very brittle assumption that could be changed easily by our build.

Additionally, `ForkJoinPool` is not I/O-aware. It atttempts to balance load based on CPU utilization but has no visibility into other kinds of work or blocking. From the [`ForkJoinPool` JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html):

> The pool attempts to maintain enough active (or available) threads by dynamically adding, suspending, or resuming internal worker threads, even if some tasks are stalled waiting to join others. However, no such adjustments are guaranteed in the face of blocked I/O or other unmanaged synchronization.

To me, this makes `ForkJoinPool` _less_ suitable for I/O-intensive workflows than the Worker API.

Ultimately, what's important is that this works. If your testing has shown this to be stable and consistently successful then we can go with it for now. But let's document the decision to eschew Worker API for `ForkJoinPool` and why we made the decision in case we need to revisit later. Code comments in the task definition should be sufficient.

ianbotsf (approved):

PR #1914, opened 2026-05-28
refactor Converter type hierarchy to provide classes instead of vals to enable their use in @DynamoDbAttributeConverter annotations

ianbotsf:
## Issue \#

(none)

## Description of changes

Providing `ItemConverter` and `ValueConverter` types as `val`s has the drawback of them being unusable in annotations like `@DynamoDbItem` and `@DynamoDbAttributeConverter` (since annotations require compile-time constants as arguments). This change reworks the various standard converters to be classes instead. Where possible, the class's companion object is the delegate for implementation which allows referring to singletons by a simple name (e.g., `CharArrayValueConverter`) instead of instantiating new values at the call site.

This change simplifies various aspects of converter construction, rendering the `MonoConverter` type unnecessary.


hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/values/smithytypes/DocumentValueConverter.kt
```diff
@@ -4,67 +4,39 @@
...
-
+) : DocumentConverterBase(
+    numberValueConverter,
+    stringValueConverter,
+    booleanValueConverter,
+    nullValueConverter,
+    attributeValueListValueConverter,
+    attributeValueMapValueConverter,
+),
+    ValueConverter<Document?> {
```
luigi617, line 35: Returning `Document?` is a behavioral change, previously AttributeValue.Null would throw `IllegalArgumentException`, now it silently returns `null`. Is this intentional?

luigi617:
The stated motivation is enabling standard converters in @DynamoDbAttributeConverter annotations. But I did not see any test that actually exercises this, e.g.`@DynamoDbAttributeConverter(BooleanValueConverter::class)` or `@DynamoDbAttributeConverter(NumberValueConverters.Int::class)` run through the schema generator? The existing tests only cover custom converters, which already worked before this change.

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/values/collections/MapValueConverter.kt
```diff
@@ -22,37 +22,43 @@ import aws.sdk.kotlin.services.dynamodb.model.AttributeValue
  * val instantMapConv2 = MapValueConverter.mapValuesFrom(InstantConverter.Default) // same as above
  * ```
  */
-public val AttributeValueMapValueConverter: ValueConverter<Map<String, AttributeValue>> =
-    Converter(AttributeValue::M, AttributeValue::asM)
+public class AttributeValueMapValueConverter : ValueConverter<Map<String, AttributeValue>> by AttributeValueMapValueConverter {
```
luigi617: **Question**: what is the purpose of using `public class Foo: ValueConverter<X> by Foo`? is there feature that self delegation bring?

luigi617 (changes requested):
Most of the codes are good for me, I left some comments for understanding better. And I want to make sure that these codes are currently only used by us now, so we api-break change is fine

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/values/smithytypes/DocumentValueConverter.kt
ianbotsf: This is an intentional breaking change. [Smithy document types](https://smithy.io/2.0/spec/simple-types.html#document) may consist of any of these primitives:
* Strings
* Numbers
* Booleans
* Lists
* Maps (string keys only)
* null

Any one of those primitives forms a valid document on its own. So for instance `{"key": "val"}` is a valid document, `42` is a valid document, and `null` is a valid document. In our low-level clients/runtime we decided against modeling an explicit `Document.Null` variant and just use `null`, which means that every modeled document member is `Document?` _regardless_ of whether it's marked `@Required`, `@Default`, etc.

You're correct that previously this converter was `ValueConverter<Document>` (non-null) and that was erroneous. I think the reasoning at the time was that a `NullableConverter` wrapper could trivially convert this to provide a `ValueConverter<Document?>` (nullable) but this short-changed one of the acceptable primitives for documents and made certain documents unrepresentable in the base converter.

While I do not believe we should revert to the previous non-nullable converter type in this class, I do now wonder whether we should tighten the converter on `DocumentItemConverter` from `ItemConverter<Document>` to `ItemConverter<Document.Map>` since that's the only document type variant we can serialize into an item (e.g., we cannot form an item from otherwise-valid documents like `42` or `null`). What do you think?

ianbotsf:
> The stated motivation is enabling standard converters in @DynamoDbAttributeConverter annotations. But I did not see any test that actually exercises this, e.g.`@DynamoDbAttributeConverter(BooleanValueConverter::class)` or `@DynamoDbAttributeConverter(NumberValueConverters.Int::class)` run through the schema generator? The existing tests only cover custom converters, which already worked before this change.

You're right, I'll add some tests for explicit use of standard converters to make sure we don't regress on this feature.

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/values/collections/MapValueConverter.kt
ianbotsf: This is subtle because it's not actually _self_ delegation—it's delegating to the companion object which by default has the same name as the class. Note that self delegation would look like `by AttributeValueMapValueConverter()` with parentheses to invoke the constructor (which would fail because it would be a recursive loop).

The reason I converted many of our previous `class` converters to `val`s was because I wanted more singletons and less allocations. Delegating to the companion object is a middle ground that allows manual uses of the old `val` names to continue unchanged (because the companion object is named the same thing). Users manually constructing their own schemas can use, for instance, `BooleanValueConverter` instead of `BooleanValueConverter()` and avoid unnecessary allocations.

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/values/smithytypes/DocumentValueConverter.kt
luigi617: Agreed, I think tightening `DocumentItemConverter` to `ItemConverter<Document.Map>` makes sense. If an Item is always a map structure (`Map<String, AttributeValue>`). Making the type system enforce this (rather than failing at runtime) is better. It also makes the API self-documenting, e.g. users won't wonder what happens if they try to serialize `Document.Number(42)` as an item.

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/values/collections/MapValueConverter.kt
luigi617: Why use the class + companion delegation pattern rather than a plain object? Both preserve backward compatibility for Kotlin callers (e.g., `BooleanValueConverter.convertRight(...)` still works). An object would also be usable in annotations (`@DynamoDbAttributeConverter(BooleanValueConverter::class)`), produce a simpler API surface, avoid the confusing `class Foo : I by Foo` syntax, and eliminate unnecessary allocation if someone calls BooleanValueConverter(). Is there any motivation that I am missing?

Another thing that I don't understand is why we need to have it backward compatible if this is a new feature that no one has used (or is anyone using it?)

ianbotsf: I initially considered that but then discarded it as unworkable because codegen couldn't know when to generate `MyValueConverter()` for classes vs `MyValueConverter` for objects. But your question made me take a second look and turns out KSP can determine whether a type declaration is an object or not. 🤯 So I will refactor these into objects instead of classes where possible!

luigi617 (approved):

PR #1918, opened 2026-06-01
Benchmark performance 2026

luigi617:
## Description of changes
Benchmark performance analysis for endpoint resolution and serde


benchmarks/endpoint-resolution/src/main/kotlin/aws/sdk/kotlin/benchmarks/endpoint/LambdaEndpointResolutionBenchmark.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    // For region us-east-1 with FIPS disabled and DualStack disabled
+    @Benchmark
+    fun usEast1Standard(blackhole: Blackhole) {
+        val params = LambdaEndpointParameters {
+            region = "us-east-1"
+            useFips = false
+            useDualStack = false
+        }
```
ianbotsf: **Nit**: Move outside of benchmark iteration to eliminate unnecessary allocation overhead. Applies throughout this PR.

benchmarks/endpoint-resolution/src/main/kotlin/aws/sdk/kotlin/benchmarks/endpoint/OverheadBenchmark.kt
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    private val precomputedEndpoint = Endpoint(Url.parse("https://lambda.us-east-1.amazonaws.com"))
+
+    private suspend fun noOpResolve(): Endpoint = precomputedEndpoint
+
+    @Benchmark
+    fun directCall(blackhole: Blackhole) {
+        blackhole.consume(precomputedEndpoint)
+    }
```
ianbotsf: **Question**: I don't see what this benchmark is checking. Is this just a baseline for comparison from the other two benchmarks? If so, we might want to add some comments to explain it better.

benchmarks/endpoint-resolution/src/main/kotlin/aws/sdk/kotlin/benchmarks/endpoint/LambdaEndpointResolutionBenchmark.kt
```diff
@@ -0,0 +1,66 @@
...
+    fun usGovEast1FipsDualStack(blackhole: Blackhole) {
+        val params = LambdaEndpointParameters {
+            region = "us-gov-east-1"
+            useFips = true
+            useDualStack = true
+        }
+        val result = suspend { provider.resolveEndpoint(params) }
+            .startCoroutineUninterceptedOrReturn(completion)
+        check(result !== COROUTINE_SUSPENDED) { "resolveEndpoint suspended unexpectedly" }
+        blackhole.consume(result)
```
ianbotsf: **Nit**: This seems like a common pattern that could be factored into a common function. You can make it `inline` to avoid increasing runtime overhead from method jumping or lambda arguments.

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkHarness.kt
```diff
@@ -0,0 +1,173 @@
...
+            appendLine("      \"p90\": ${result.stats.p90},")
+            appendLine("      \"p95\": ${result.stats.p95},")
+            appendLine("      \"p99\": ${result.stats.p99},")
+            appendLine("      \"std_dev\": ${result.stats.stdDev}")
+            append("    }")
+            if (i < results.size - 1) appendLine(",") else appendLine()
+        }
+        appendLine("  ]")
+        appendLine("}")
+    }
```
ianbotsf: **Correctness**: Do not construct complex JSON, XML, YAML, etc. by string concatenation and template substitution. It's too brittle, error-prone, and makes debugging & refactoring difficult. Instead, use language or library constructs to build documents in an idiomatic way where the compiler can help catch more mistakes.

In the case of Kotlin, I recommend using [**kotlinx-serialization**](https://kotlinlang.org/docs/serialization.html). You declare `data classes` to represent your object shape, annotate them with `@Serializable`, and invoke the format-specific serializer:

```kotlin
@Serializable
data class BenchmarkReport(
    metadata: BenchmarkMetadata,
    results: List<BenchmarkResult>,
)

@Serializable
data class BenchmarkMetadata(...)

@Serializable
data class BenchmarkResult(...)

fun toJson(report: BenchmarkReport): String = Json.encodeToString(report)
```

We already use this approach for thins like [writing package manifest files](https://github.com/aws/aws-sdk-kotlin/blob/main/build-support/src/main/kotlin/aws/sdk/kotlin/gradle/sdk/PackageManifest.kt) and [parsing test definitions from documents](https://github.com/aws/aws-sdk-kotlin/blob/main/aws-runtime/aws-config/jvm/test/aws/sdk/kotlin/runtime/config/retries/AwsRetryIntegrationTestResources.kt).

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkInterceptor.kt
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    override fun readBeforeDeserialization(context: ProtocolResponseInterceptorContext<Any, HttpRequest, HttpResponse>) {
+        deserializationStartNanos = System.nanoTime()
+    }
+
+    override fun readAfterDeserialization(context: ResponseInterceptorContext<Any, Any, HttpRequest, HttpResponse>) {
+        deserializationEndNanos = System.nanoTime()
+    }
+}
```
ianbotsf, line 61: **Style**: Consider using [the `kotlin.time` APIs](https://kotlinlang.org/docs/time-measurement.html) instead of `nanoTime`. We take this approach in [`OperationTelemetryInterceptor`](https://github.com/smithy-lang/smithy-kotlin/blob/ec5486c7a56010359b46c15fb732691a4aec5073/runtime/protocol/http-client/common/src/aws/smithy/kotlin/runtime/http/interceptors/OperationTelemetryInterceptor.kt#L41-L44) for a very related use case. This typically simplifies the code and also makes it KMP-proof since `System.nanoTime()` is JVM-only. (Fun fact: [`System.nanoTime()` is the actual JVM implementation](https://github.com/JetBrains/kotlin/blob/9c58a07826905ae022fee40e7bb3049cef02da99/libraries/stdlib/jvm/src/kotlin/time/MonoTimeSource.kt#L12-L13) of these `kotlin.time` APIs!)

benchmarks/serde-benchmarks/build.gradle.kts
```diff
@@ -0,0 +1,177 @@
...
+    implementation(libs.smithy.kotlin.aws.credentials)
+    implementation(libs.smithy.kotlin.aws.protocol.core)
+    implementation(libs.smithy.kotlin.aws.json.protocols)
+    implementation(libs.smithy.kotlin.aws.xml.protocols)
+    implementation(libs.smithy.kotlin.aws.signing.common)
+    implementation(libs.smithy.kotlin.aws.signing.default)
+    implementation(libs.smithy.kotlin.identity.api)
+    implementation(libs.smithy.kotlin.telemetry.defaults)
+    implementation(libs.smithy.kotlin.smithy.rpcv2.protocols)
+}
```
ianbotsf, line 104: **Question**: That's a lot of dependencies! Are all of these really necessary?

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkRunner.kt
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Discovers and runs all generated serde benchmark classes.
+ *
+ * Each generated benchmark class follows the pattern:
+ * - Has a no-arg constructor
+ * - Has a `suspend fun benchmarks(): List<BenchmarkResult>` method
+ *
+ * This runner uses reflection to find all benchmark classes on the classpath.
+ */
+fun main() = runBlocking {
```
ianbotsf, line 9: **Correctness**: Runtime reflection makes this KMP incompatible. Plus, since this is a build-time task, it's better suited to Gradle code and the native features of our benchmark framework.

**kotlinx-benchmark** has a concept of [configurations](https://github.com/Kotlin/kotlinx-benchmark/blob/master/docs/configuration-options.md) which I think map nicely to the protocol projections. I think you could move benchmark discovery into **build.gradle.kts** by dynamically adding configurations based on the `benchmarkProjections` list. There are probably other approaches too that are more KMP-friendly and better integrated with our build system.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/SerdeBenchmarkIntegration.kt
```diff
@@ -0,0 +1,270 @@
...
+                        withBlock("#T.run(", ")", AwsRuntimeTypes.Benchmarks.BenchmarkHarness) {
+                            write("id = #S,", testCase.id)
+                            write("interceptor = interceptor,")
+                            write("extractNanos = #T::serializationNanos,", AwsRuntimeTypes.Benchmarks.BenchmarkInterceptor)
+                        }
+                        write("{ client.#L(#L) }", opName, fieldName)
+                    }
+                }
+                write("return results")
+            }
```
ianbotsf, line 155: **Nit**: Unfortunately, we've seen issues in the past with modeled shapes named the same as stdlib APIs like `List` or `String`. It's best to let the `KotlinWriter` verify the imports by using `#T` even for fundamental types like `List`:

```kotlin
withBlock("suspend fun benchmarks(): #T<#T> {", "}",
    KotlinTypes.Collections.List,
    AwsRuntimeTypes.Benchmarks.BenchmarkResult,
) {
```

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/SerdeBenchmarkIntegration.kt
```diff
@@ -0,0 +1,270 @@
...
+                )
+            } else {
+                writer.write("private val #L = #S.encodeToByteArray()", bodyFieldName, body)
+            }
+        }
+
+        writer.withBlock("private val #L = #T {", "}", fieldName, serviceSymbol) {
+            withBlock("httpClient = #T { _, request ->", "}", RuntimeTypes.HttpTest.TestEngine) {
+                if (testCase.headers.isNotEmpty()) {
+                    withBlock("val respHeaders = #T().apply {", "}.build()", RuntimeTypes.Http.HeadersBuilder) {
```
ianbotsf: **Nit**: `HeadersBuilder().apply { ... }.build()` can be replaced with [`Headers { ... }`](https://github.com/smithy-lang/smithy-kotlin/blob/ec5486c7a56010359b46c15fb732691a4aec5073/runtime/protocol/http/common/src/aws/smithy/kotlin/runtime/http/Headers.kt#L18-L19).

settings.gradle.kts
```diff
@@ -58,6 +58,18 @@ include(":hll")
...
+    if (endpointBenchmarkServices.all { it.isBootstrappedService }) {
+        include(":benchmarks:endpoint-resolution")
+    } else {
+        val missing = endpointBenchmarkServices.filterNot { it.isBootstrappedService }
+        logger.warn("Skipping :benchmarks:endpoint-resolution because these service(s) are not bootstrapped: $missing")
+    }
+}
+if (providers.gradleProperty("includeBenchmarks").orNull?.toBoolean() == true) {
+    include(":benchmarks:serde-benchmarks")
+}
```
ianbotsf: **Nit**: `if` statements with identical criteria could be combined.

ianbotsf: **Question**: Why is an `includeBenchmarks` Gradle property necessary? In [other benchmarks](https://github.com/aws/aws-sdk-kotlin/blob/8154650b09bd2091ae997404c1e6efcf6a9a2a53/settings.gradle.kts#L94-L112) we include the project so long as all the required services are present. Note that this doesn't automatically _run_ the benchmarks since that's a separate Gradle task not included in the generic `build` task.

This implementation allows the possibility that `includeBenchmarks` was set to `true` but not all the required services were bootstrapped, leading to an easily-missed warning.

ianbotsf (changes requested):

benchmarks/endpoint-resolution/src/main/kotlin/aws/sdk/kotlin/benchmarks/endpoint/OverheadBenchmark.kt
luigi617: That was used just to make sure that the `precomputedEndpoint` is actually not introducing any invisible overhead underneath, I will remove it

settings.gradle.kts
luigi617: I added this parameter `includeBenchmarks` because it was not passing the CI. I'll try the other benchmarks approach!

luigi617: I just remember why I added `includeBenchmarks`.  For serde benchmark, there is a `benchmarkModelsDir` parameter which is required, this is used for providing the path for the models (which are not in the repo). And since it was required here, I thought it would be consistent to add it also on endpoint resolution. Another option would be make it optional in compile but throw error in runtime. What do you think?

benchmarks/serde-benchmarks/build.gradle.kts
luigi617: I think so, because the generated code needs all dependencies (serde, http, aws runtime, ...)

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkRunner.kt
luigi617: Replaced the runtime reflection-based discovery with a codegen-generated registry. Each projection now emits a BenchmarkRegistration.kt file (via SerdeBenchmarkIntegration.writeAdditionalFiles) that directly instantiates its benchmark classes. The runner imports and calls these registration functions — no Class.forName, no classpath scanning.

The serde benchmarks measure sub-operation timing (only the serde portion via interceptor hooks), which doesn't map to kotlinx-benchmark's @Benchmark model where the entire function is timed. But as discussed in design doc, this was not a "good" option

benchmarks/endpoint-resolution/src/main/kotlin/aws/sdk/kotlin/benchmarks/endpoint/LambdaEndpointResolutionBenchmark.kt
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package aws.sdk.kotlin.benchmarks.endpoint
+
+import aws.sdk.kotlin.services.lambda.endpoints.DefaultLambdaEndpointProvider
+import aws.sdk.kotlin.services.lambda.endpoints.LambdaEndpointParameters
+import kotlinx.benchmark.*
+import org.openjdk.jmh.annotations.State
```
ianbotsf: **Nit**: Looks like we're importing the `@State` annotation from JMH, not from **kotlinx-benchmark**. I think we need `import kotlinx.benchmark.State` instead. Applies in other files too.

benchmarks/endpoint-resolution/src/main/kotlin/aws/sdk/kotlin/benchmarks/endpoint/BenchmarkUtils.kt
```diff
@@ -0,0 +1,23 @@
...
+    result.getOrThrow()
+}
+
+inline fun resolveEndpointSync(completion: Continuation<Endpoint>, crossinline block: suspend () -> Endpoint): Endpoint {
+    val result = suspend { block() }
+        .startCoroutineUninterceptedOrReturn(completion)
+    check(result !== COROUTINE_SUSPENDED) { "resolveEndpoint suspended unexpectedly" }
+    @Suppress("UNCHECKED_CAST")
+    return result as Endpoint
+}
```
ianbotsf: **Question**: Why accept a `completion: Continuation<Endpoint>` argument if every caller sets it to the same value of `benchmarkCompletion`?

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkHarness.kt
```diff
@@ -0,0 +1,176 @@
...
+     *   All warmup samples are discarded.
+     * - Measurement phase: run for MEASUREMENT_SECONDS (default 30s) or until MAX_ITERATIONS,
+     *   whichever comes first. Must record at least MIN_ITERATIONS (default 1,000).
+     */
+    suspend fun run(
+        id: String,
+        interceptor: BenchmarkInterceptor,
+        extractNanos: (BenchmarkInterceptor) -> Long,
+        operation: suspend () -> Unit,
+    ): BenchmarkResult {
```
ianbotsf, line 65: **Nit**: Use `inline` for `run` and `crossinline` for `operation`:

```kotlin
inline suspend fun run(
    id: String,
    interceptor: BenchmarkInterceptor,
    extractNanos: (BenchmarkInterceptor) -> Long,
    crossinline operation: suspend () -> Unit,
): BenchmarkResult {
    ...
}
```

This will optimize the runner to avoid lambda overhead when invoking `operation`.

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkHarness.kt
```diff
@@ -0,0 +1,176 @@
...
+        // Phase 1: Warmup — discard all results
+        val warmupStart = System.nanoTime()
+        while (System.nanoTime() - warmupStart < WARMUP_NANOS) {
+            interceptor.reset()
+            operation()
+        }
```
ianbotsf, line 73: **Nit**: `System.nanoTime` is JVM only. Replace with `kotlin.time` APIs:

```kotlin
val warmupEnd = timeSource.markNow() + WARMUP_NANOS.nanoseconds
while (warmupEnd.hasNotPassedNow()) {
    interceptor.reset()
    operation()
}
```

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkHarness.kt
```diff
@@ -0,0 +1,176 @@
...
+
+        // Flush garbage accumulated during warmup before measuring
+        System.gc()
+
+        // Phase 2: Measurement — collect individual samples and track 1-second windows
+        val samples = ArrayList<Long>(MAX_ITERATIONS.coerceAtMost(1_000_000))
```
ianbotsf: **Question**: The default value for `MAX_ITERATIONS` is 10,000,000 but this clamps it at 1,000,000. Why the discrepancy? Why allow a configurable maximum if we're going to cap it at a hardcoded limit?

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkHarness.kt
```diff
@@ -0,0 +1,176 @@
...
+        val windowMeans = ArrayList<Double>()
```
ianbotsf: **Nit**: Initialize the size of this array to avoid allocations during list resizing

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkHarness.kt
```diff
@@ -0,0 +1,176 @@
...
+        val measureStart = System.nanoTime()
+        var windowStart = measureStart
+        var windowSum = 0L
+        var windowCount = 0
+
+        while (samples.size < MAX_ITERATIONS) {
```
ianbotsf, line 86: **Correctness**: The `samples` list is initialized to a size of 1,000,000 at most but `MAX_ITERATIONS` may be greater than that, which will cause allocations when the list is resized.

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkHarness.kt
```diff
@@ -0,0 +1,176 @@
...
+        val mean = sum.toDouble() / n
+
+        // std_dev over 1-second window means (like JMH iteration means)
+        val stdDev = if (windowMeans.size > 1) {
+            val wmMean = windowMeans.average()
+            val variance = windowMeans.sumOf { (it - wmMean) * (it - wmMean) } / windowMeans.size
+            sqrt(variance)
+        } else {
+            0.0
+        }
```
ianbotsf: **Nit**: The `if` check is unnecessary since your stddev calculation will work with `size = 1`.

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkHarness.kt
```diff
@@ -0,0 +1,176 @@
...
+        return json.encodeToString(report)
+    }
+}
+
+data class BenchmarkMetadata(
+    val smithyKotlinVersion: String,
+    val sdkVersion: String,
+    val os: String = "${System.getProperty("os.name")} ${System.getProperty("os.version")}",
+    val instance: String = System.getProperty("benchmark.instance", "unknown"),
+)
```
ianbotsf: **Style**: It's confusing to have both `BenchmarkMetadataJson` and `BenchmarkMetadata`. I suggest renaming this to `KotlinBenchmarkMetadata` since it's specific to _our_ internal metadata.

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkRunner.kt
ianbotsf: The list of all protocols is still statically defined, meaning that adding a new protocol will involve touching both **build.gradle.kts** and **BenchmarkRunner.kt**. I suggest code-generating the `registerAllProtocols` function and then invoking it here.

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkRunner.kt
```diff
@@ -0,0 +1,48 @@
...
+        val benchmark = entry.create()
+        val results = benchmark.benchmarks()
+        allResults.addAll(results)
+        println("  Completed ${results.size} test cases")
+    }
+
+    val metadata = BenchmarkMetadata(
+        smithyKotlinVersion = System.getProperty("smithy.kotlin.version", "SNAPSHOT"),
+        sdkVersion = System.getProperty("aws.sdk.kotlin.version", "SNAPSHOT"),
+    )
```
ianbotsf: **Question**: How will these sysprops get configured?

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/SerdeBenchmark.kt
```diff
@@ -0,0 +1,22 @@
...
+    val create: () -> SerdeBenchmark,
+)
+
+object BenchmarkRegistry {
+    val entries: MutableList<BenchmarkRegistryEntry> = mutableListOf()
+
+    fun register(name: String, create: () -> SerdeBenchmark) {
+        entries.add(BenchmarkRegistryEntry(name, create))
+    }
+}
```
ianbotsf, line 24: **Nit**: This violates encapsulation by providing `entries` as a public mutable list. Please create an explicit backing field:

```kotlin
private val _entries = mutableListOf<BenchmarkRegistryEntry>()
val entries: List<BenchmarkRegistryEntry>
    get() = _entries
```

This will get slightly simpler when we upgrade to Kotlin 2.4.x which provides [explicit backing fields](https://kotlinlang.org/docs/properties.html#explicit-backing-fields) as a stable language feature:

```kotlin
// `entries` is a `List<...>` publicly but a `MutableList<...>` privately
val entries: List<BenchmarkRegistryEntry>
    field = mutableListOf()
```

benchmarks/serde-benchmarks/build.gradle.kts
```diff
@@ -0,0 +1,169 @@
...
+
+tasks.generateSmithyProjections {
+    val sdkVersion: String by project
+    doFirst {
+        System.setProperty("smithy.kotlin.codegen.clientRuntimeVersion", sdkVersion)
+    }
```
ianbotsf: **Question**: What does setting this property accomplish?

benchmarks/serde-benchmarks/build.gradle.kts
```diff
@@ -0,0 +1,169 @@
...
+    group = "benchmark"
+    description = "Run all serde benchmarks sequentially"
+    dependsOn("classes")
+    classpath = sourceSets["main"].runtimeClasspath
+    mainClass.set("aws.sdk.kotlin.benchmarks.serde.BenchmarkRunnerKt")
+    systemProperty("benchmark.warmupSeconds", project.findProperty("benchmark.warmupSeconds") ?: "10")
+    systemProperty("benchmark.measurementSeconds", project.findProperty("benchmark.measurementSeconds") ?: "30")
+    systemProperty("benchmark.minIterations", project.findProperty("benchmark.minIterations") ?: "1000")
+    systemProperty("benchmark.maxIterations", project.findProperty("benchmark.maxIterations") ?: "10000000")
+    systemProperty("benchmark.instance", project.findProperty("benchmark.instance") ?: "unknown")
```
ianbotsf, line 176: **Nit**: We should not define default values for these properties in multiple locations. We should default them in just one place, preferably here in **build.gradle.kts** since they're related to a build task.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/SerdeBenchmarkIntegration.kt
```diff
@@ -0,0 +1,293 @@
...
+        ctx: ProtocolGenerator.GenerationContext,
+        writer: KotlinWriter,
+        operation: OperationShape,
+        className: String,
+        testCases: List<HttpResponseTestCase>,
+    ) {
+        packageName = ctx.settings.pkg.name
+        generatedClassNames.add(className)
+        HttpProtocolSerdeBenchmarkGenerator(ctx, writer, operation).renderResponseBenchmarkClass(className, testCases)
+    }
```
ianbotsf: **Question**: Why are we capturing `packageName` inside of these `render*` methods but only using it in `writeAdditionalFiles`? Couldn't you just use the context that's available there?

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/SerdeBenchmarkIntegration.kt
```diff
@@ -0,0 +1,293 @@
...
+        if (generatedClassNames.isEmpty()) return
+
+        delegator.useTestFileWriter("BenchmarkRegistration.kt", packageName) { writer ->
+            writer.withBlock("internal fun registerBenchmarks() {", "}") {
+                for (className in generatedClassNames.sorted()) {
+                    write("#T.register(#S) { #L() }", AwsRuntimeTypes.Benchmarks.BenchmarkRegistry, className, className)
+                }
+            }
+        }
+    }
```
ianbotsf, line 70: **Correctness**: This relies on `generatedClassNames` being populated first, which depends on codegen execution order. If we change how codegen executes by reordering things, this integration may start silently failing.

Can't we determine the class names inside this method using the model?

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/SerdeBenchmarkIntegration.kt
```diff
@@ -0,0 +1,293 @@
...
+
+    private val idempotentFieldsInModel: Boolean by lazy {
+        operation.input.isPresent &&
+            model.expectShape(operation.input.get()).members().any { it.hasTrait(IdempotencyTokenTrait.ID.name) }
+    }
+
+    fun renderRequestBenchmarkClass(className: String, testCases: List<HttpRequestTestCase>) {
+        writer.write("")
+        writer.withBlock("class #L : #T {", "}", className, AwsRuntimeTypes.Benchmarks.SerdeBenchmark) {
+            write("")
```
ianbotsf, line 116: **Nit**: Unnecessary newline after class declaration

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/SerdeBenchmarkIntegration.kt
```diff
@@ -0,0 +1,293 @@
...
+                }
+                write("return results")
+            }
+        }
+
+        writer.write("")
+        writer.withBlock("fun main() = #T {", "}", RuntimeTypes.KotlinxCoroutines.runBlocking) {
+            write("val benchmark = #L()", className)
+            write("val results = benchmark.benchmarks()")
+            write("val metadata = #T(smithyKotlinVersion = #S, sdkVersion = #S)", AwsRuntimeTypes.Benchmarks.BenchmarkMetadata, "SNAPSHOT", "SNAPSHOT")
```
ianbotsf: **Question**: Why are we hardcoding `smithyKotlinVersion` and `sdkVersion` to `"SNAPSHOT"`? Shouldn't those be actual version numbers?

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/SerdeBenchmarkIntegration.kt
```diff
@@ -0,0 +1,293 @@
...
+                    ShapeValueGenerator(model, symbolProvider, explicitReceiver = true)
+                        .instantiateShapeInline(writer, inputShape, testCase.params)
+                }
+                .dedent()
+                .write("")
+        }
+    }
+
+    private fun renderResponseClientField(testCase: HttpResponseTestCase) {
+        val fieldName = "client_${sanitizeName(testCase.id)}"
```
ianbotsf: **Nit**: `fieldName` → `clientFieldName`

ianbotsf (changes requested):

benchmarks/endpoint-resolution/src/main/kotlin/aws/sdk/kotlin/benchmarks/endpoint/BenchmarkUtils.kt
luigi617: You're right, it is not required

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkHarness.kt
luigi617: You're right, this was used also to check windowMeans is non-empty, but it is guaranteed to be non-empty

benchmarks/serde-benchmarks/src/main/kotlin/aws/sdk/kotlin/benchmarks/serde/BenchmarkRunner.kt
luigi617: I've been changing them manually, I added so that they can be set using system property

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/SerdeBenchmarkIntegration.kt
luigi617: The entire `main` function was used to run each benchmark alone. But I think it is useless, I will add the option to run benchmark for each protocol.

luigi617:
Currently, I am generating the function that collects all protocol registry in the **build.gradle.kts** because this is the only place where I know all the existing protocols. Is there any better approach?

PR #1924, opened 2026-06-18
Cleanup todos and fixmes

ianbotsf:
## Issue \#

Related to #472 

## Description of changes

This PR is a cleanup of various `TODO`s and `FIXME`s in DynamoDB Mapper's codebase. In particular:
* Provide options for handling unknown values in `SimpleItemConverter`. This will allow users to gracefully handle cases where items are not fully modeled in a schema.
* Properly map Condition Expressions to `FilterDsl`. This has the effect of adding a new `condition: BooleanExpr` property to `DeleteItemRequest`, `PutItemRequest`, and `UpdateItemRequest`, along with adding a new `condition(block: FilterDsl.() -> BooleanExpr?)` method to each of those requests' builders.
* Pass a proper list of file dependencies to KSP's `CodeGenerator` when creating new files for ops codegen. This will speed up build times by narrowing the set of dependency files to check before potentially regenerating DDB Mapper operation files.
* Removed various `FIXME`s and `TODO`s which are not planned to be implemented/addressed. Some of these were speculative or have become obsolete.

### DIFF
--- a/hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/HighLevelOpsProcessor.kt
+++ b/hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/HighLevelOpsProcessor.kt
@@ -15,6 +15,7 @@ import aws.sdk.kotlin.hll.dynamodbmapper.codegen.operations.rendering.HighLevelR
 import aws.sdk.kotlin.services.dynamodb.DynamoDbClient
 import com.google.devtools.ksp.getClassDeclarationByName
 import com.google.devtools.ksp.getDeclaredFunctions
+import com.google.devtools.ksp.processing.Dependencies
 import com.google.devtools.ksp.processing.Resolver
 import com.google.devtools.ksp.processing.SymbolProcessorEnvironment
 import com.google.devtools.ksp.symbol.KSAnnotated
@@ -31,14 +32,20 @@ internal class HighLevelOpsProcessor(environment: SymbolProcessorEnvironment) : 
     private val opAllowlist = environment.options["op-allowlist"]?.split(";")
     private val pkg = environment.options["pkg"] ?: MapperPkg.Hl.Ops.Base
 
-    private val ctx by lazy {
-        val codegenFactory = CodeGeneratorFactory(codeGenerator, logger) // FIXME Pass dependencies
-        RenderContext(logger, codegenFactory, pkg, "dynamodb-mapper-ops-codegen")
-    }
-
     override fun processImpl(resolver: Resolver): List<KSAnnotated> {
         logger.info("Scanning low-level DDB client for operations and types")
-        val operations = getOperations(resolver)
+        val operationFunctions = resolver
+            .getClassDeclarationByName<DynamoDbClient>()!!
+            .getDeclaredFunctions()
+            .filter(::allow)
+
+        val dependencyFiles = operationFunctions.mapNotNull { it.containingFile }
+        val codegenDeps = Dependencies(aggregating = false, sources = dependencyFiles.toList().toTypedArray())
+        logger.info("Found dependency files:\n${codegenDeps.originatingFiles.joinToString("\n") { it.filePath }}")
+
+        val codegenFactory = CodeGeneratorFactory(codeGenerator, logger, codegenDeps)
+        val ctx = RenderContext(logger, codegenFactory, pkg, "dynamodb-mapper-ops-codegen")
+        val operations = getOperations(resolver, ctx)
 
         logger.info("Rendering high-level operations and types for ${operations.map { it.name }}")
         HighLevelRenderer(ctx, operations).render()
@@ -59,7 +66,7 @@ internal class HighLevelOpsProcessor(environment: SymbolProcessorEnvironment) : 
         return allowed ?: true
     }
 
-    private fun getOperations(resolver: Resolver): List<Operation> = resolver
+    private fun getOperations(resolver: Resolver, ctx: RenderContext): List<Operation> = resolver
         .getClassDeclarationByName<DynamoDbClient>()!!
         .getDeclaredFunctions()
         .filter(::allow)
--- a/hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/HighLevelOpsProcessor.kt
+++ b/hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/HighLevelOpsProcessor.kt
@@ -45,7 +45,11 @@ internal class HighLevelOpsProcessor(environment: SymbolProcessorEnvironment) : 
 
         val codegenFactory = CodeGeneratorFactory(codeGenerator, logger, codegenDeps)
         val ctx = RenderContext(logger, codegenFactory, pkg, "dynamodb-mapper-ops-codegen")
-        val operations = getOperations(resolver, ctx)
+
+        val operations = operationFunctions
+            .map(Operation::from)
+            .map { it.toHighLevel(ctx) }
+            .toList()
 
         logger.info("Rendering high-level operations and types for ${operations.map { it.name }}")
         HighLevelRenderer(ctx, operations).render()
@@ -65,12 +69,4 @@ internal class HighLevelOpsProcessor(environment: SymbolProcessorEnvironment) : 
 
         return allowed ?: true
     }
-
-    private fun getOperations(resolver: Resolver, ctx: RenderContext): List<Operation> = resolver
-        .getClassDeclarationByName<DynamoDbClient>()!!
-        .getDeclaredFunctions()
-        .filter(::allow)
-        .map(Operation::from)
-        .map { it.toHighLevel(ctx) }
-        .toList()
 }
--- a/hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/MemberCodegenBehavior.kt
+++ b/hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/MemberCodegenBehavior.kt
@@ -91,10 +91,8 @@ internal sealed interface MemberCodegenBehavior {
  * Identifies a type of expression literal supported by DynamoDB APIs
  */
 internal enum class ExpressionLiteralType {
-    Condition,
     Filter,
     KeyCondition,
-    Projection,
     Update,
 }
 
@@ -330,10 +328,6 @@ private val rules = listOf(
     Rule("keyConditionExpression", Types.Kotlin.String, ExpressionLiteral(KeyCondition)),
     Rule("filterExpression", Types.Kotlin.String, ExpressionLiteral(Filter)),
     Rule("updateExpression", Types.Kotlin.String, ExpressionLiteral(Update)),
-
-    // TODO add support for remaining expression types
-    Rule("conditionExpression", Types.Kotlin.String, Drop),
-    Rule("projectionExpression", Types.Kotlin.String, Drop),
 
     // Expression arguments
     Rule("expressionAttributeNames", Types.Kotlin.stringMap(Types.Kotlin.String), ExpressionArguments(AttributeNames)),
--- a/hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/MemberCodegenBehavior.kt
+++ b/hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/MemberCodegenBehavior.kt
@@ -301,7 +301,7 @@ private val transactWriteItemsRequestTables = CustomTransformation(
  * successfully matches with a member will be chosen.
  */
 private val rules = listOf(
-    // Deprecated expression members not to be carried forward into HLL
+    // Deprecated and/or unsupported expression members not to be carried forward into HLL
     Rule("conditionalOperator", llType("ConditionalOperator"), Drop),
     Rule("expected", Types.Kotlin.stringMap(llType("ExpectedAttributeValue")), Drop),
     Rule("queryFilter", Types.Kotlin.stringMap(llType("Condition")), Drop),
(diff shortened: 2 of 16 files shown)


hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/HighLevelOpsProcessor.kt
```diff
@@ -31,14 +32,20 @@ internal class HighLevelOpsProcessor(environment: SymbolProcessorEnvironment) :
...
+            .getDeclaredFunctions()
+            .filter(::allow)
+
+        val dependencyFiles = operationFunctions.mapNotNull { it.containingFile }
+        val codegenDeps = Dependencies(aggregating = false, sources = dependencyFiles.toList().toTypedArray())
+        logger.info("Found dependency files:\n${codegenDeps.originatingFiles.joinToString("\n") { it.filePath }}")
+
+        val codegenFactory = CodeGeneratorFactory(codeGenerator, logger, codegenDeps)
+        val ctx = RenderContext(logger, codegenFactory, pkg, "dynamodb-mapper-ops-codegen")
+        val operations = getOperations(resolver, ctx)
```
luigi617: `operationFunctions` has already done part of work for the `getOperations` function. Since `getOperations` is private, I assume that this is the only place this is called. I think this can be simplified. Maybe just do:
```
val operations = operationFunctions.map(Operation::from)
        .map { it.toHighLevel(ctx) }
        .toList()
```

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/FilterDsl.kt
```diff
@@ -655,9 +655,6 @@ public interface FilterDsl {
      */
     public fun AttributePath.isBetween(min: ByteArray, max: ByteArray): BooleanExpr = isBetween(LiteralExpr(min), LiteralExpr(max))
 
-    // TODO The following overloads support [ClosedRange] but [OpenEndRange] also exists. DynamoDB expressions don't
-    //  support it directly but we may be able to cheese it with two inequalities ANDed together.
-
```
luigi617, line 660: General question: when you remove only TODOs/FIXME without any implementation, does it mean that they are already solved or won't be fixed?

hll/ddb-mapper/dynamodb-mapper/common/test/aws/sdk/kotlin/hll/dynamodbmapper/items/SimpleItemConverterTest.kt
```diff
@@ -5,11 +5,16 @@
 package aws.sdk.kotlin.hll.dynamodbmapper.items
 
 import aws.sdk.kotlin.hll.dynamodbmapper.model.intersectKeys
+import aws.sdk.kotlin.hll.dynamodbmapper.model.itemOf
+import aws.sdk.kotlin.hll.dynamodbmapper.util.av
 import aws.sdk.kotlin.hll.dynamodbmapper.values.scalars.BooleanValueConverter
 import aws.sdk.kotlin.hll.dynamodbmapper.values.scalars.NumberValueConverters
 import aws.sdk.kotlin.hll.dynamodbmapper.values.scalars.StringValueConverter
+import aws.sdk.kotlin.services.dynamodb.model.AttributeValue
+import kotlinx.coroutines.channels.ProducerScope
```
luigi617: Nit: import never used

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/items/KeyConversion.kt
luigi617: Nit: With the nullable itemToCk overload removed, this annotation on the remaining overload is no longer needed.

luigi617 (changes requested):

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/expressions/FilterDsl.kt
ianbotsf: Good question! In hindsight, I should've annotated each removal with some reasoning.

In this particular case, the `TODO` won't be addressed at this time. For context, Kotlin has [two kinds of ranges](https://kotlinlang.org/docs/ranges.html): closed (end-inclusive) and open-ended (end-exclusive). Supporting [`ClosedRange`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.ranges/-closed-range/) matches the API capabilities of DynamoDB and should be sufficient for most users. [DDB's filter syntax](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.OperatorsAndFunctions.html) for `BETWEEN` only works with closed ranges so [`OpenEndRange`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.ranges/-open-end-range/) support would require forming our own boolean expressions with inequalities.

For example, the following mapper DSL:

```kotlin
attr["foo"] isIn 0..<10 // open-ended range: 0 ≤ foo < 10
```

Could be converted to the following DDB filter expression:

```
(foo >= 0 AND foo < 10)
```

We haven't fully fleshed out this use case so I'm inclined to do nothing until we get signals from users that they want this and why. Fortunately, all of the stdlib discrete range classes (e.g., [`IntRange`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.ranges/-int-range/), [`CharRange`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.ranges/-char-range/), etc.) already implement _both_ `ClosedRange` and `OpenEndRange` so they require no additional support. Implementing this feature would primarily benefit continuous ranges, such as `OpenEndRange<Double>` or `OpenEndRange<String>`.

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/MemberCodegenBehavior.kt
```diff
@@ -91,10 +91,8 @@ internal sealed interface MemberCodegenBehavior {
  * Identifies a type of expression literal supported by DynamoDB APIs
  */
 internal enum class ExpressionLiteralType {
-    Condition,
     Filter,
```
ianbotsf, line 94: **Comment**: Dropping `Condition` as an expression type distinct from `Filter`. DynamoDB's documentation refers to "[condition and filter expressions](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.OperatorsAndFunctions.html)" but they are functionally equivalent and so do not need distinct representations in our codegen layer.

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/MemberCodegenBehavior.kt
```diff
@@ -91,10 +91,8 @@ internal sealed interface MemberCodegenBehavior {
...
     KeyCondition,
-    Projection,
```
ianbotsf, line 97: **Comment**: [Projections and projection expressions](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ProjectionExpressions.html) will not be supported in the mapper at this time. Because of Kotlin's idiomatic gravity towards non-nullable properties in data classes, we cannot support projections by simply omitting fields from an entity like [the Java mapper does](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/ddb-en-client-adv-features-nested.html#ddb-en-client-adv-features-nested-projection). To truly support projection expressions, we'd need to support alternate entity representations, either user-provided or code generated. This may be a complex feature to implement in a type-safe manner so I prefer to wait for user signal that they want this feature and why.

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/model/MemberCodegenBehavior.kt
```diff
@@ -328,13 +327,9 @@ private val rules = listOf(
 
     // Expression literals
     Rule("keyConditionExpression", Types.Kotlin.String, ExpressionLiteral(KeyCondition)),
-    Rule("filterExpression", Types.Kotlin.String, ExpressionLiteral(Filter)),
+    Rule("filterExpression|conditionExpression".toRegex(), Types.Kotlin.String, ExpressionLiteral(Filter)),
     Rule("updateExpression", Types.Kotlin.String, ExpressionLiteral(Update)),
 
-    // TODO add support for remaining expression types
-    Rule("conditionExpression", Types.Kotlin.String, Drop),
-    Rule("projectionExpression", Types.Kotlin.String, Drop),
```
ianbotsf, line 336: **Comment**: As discussed above, condition expressions are just filter expressions and we will not implement projections at this time.

hll/ddb-mapper/dynamodb-mapper-ops-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/operations/rendering/OperationsTypeRenderer.kt
```diff
@@ -27,7 +27,6 @@ internal class OperationsTypeRenderer(
     private val operations: List<Operation>,
 ) : RendererBase(ctx, "${itemSourceKind.name}Operations") {
     private companion object {
-        // FIXME Support unsigned number types?
         val keyDataTypes = listOf(
             Types.Kotlin.Byte,
             Types.Kotlin.ByteArray,
```
ianbotsf, line 32: **Comment**: Adding this support is as trivial as expanding this list but would result in significantly more auto-generated extension methods, increasing our artifact size. See https://github.com/aws/aws-sdk-kotlin/pull/1897 for a discussion of the tradeoffs of auto-generated convenience extensions. Given the comparatively light usage of unsigned types vs signed types, I'm inclined to do nothing until we receive clearer user signal they want this feature and why.

Note that this `TODO` is solely about supporting _auto-generated convenience extensions_ for single-attribute unsigned keys. We still support unsigned key types (even for single-attribute keys) via the full syntax, such as:

```kotlin
val item = myTable.getItem {
    partitionKey = Key(42u) // UInt
}
```

Completing this feature would provide a compact syntax similar to what's achievable presently with signed types:

```kotlin
val item1 = table1.getItem(42)  // Int  — works today
val item2 = table2.getItem(42u) // UInt — no such method
```

hll/ddb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -154,7 +154,6 @@ internal class SchemaRenderer(
      * with our [ItemToValueConverter]
      */
     private fun renderValueConverter() {
-        // TODO Offer alternate serialization options besides AttributeValue.M?
         generatedAnnotation()
```
ianbotsf, line 157: **Comment**: This `TODO` speculated that users may want to represent nested entities in forms other than [DynamoDB's map data type](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes.Document.Map). We have no signal from our users presently about what other forms they'd want so I'm inclined to do nothing here until such time as we have concrete requests.

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/pipeline/MapperContext.kt
```diff
@@ -12,8 +12,6 @@ import aws.sdk.kotlin.hll.dynamodbmapper.pipeline.internal.MapperContextImpl
  * @param T The type of objects being converted to/from DynamoDB items
  */
 public interface MapperContext<T> {
-    // TODO what other fields would be useful in here?
-
```
ianbotsf, line 16: **Comment**: No other fields useful for interceptor context have been identified at this time. Fortunately, adding items to the interface will be backwards-compatible if done carefully.

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/util/AttributeValues.kt
```diff
@@ -64,7 +64,7 @@ internal fun dynamicAv(value: Any?): AttributeValue = when (value) {
     is Map<*, *> -> av(value as Map<String, Any?>)
     is Number -> av(value)
     is Set<*> -> when (val type = value.firstOrNull()) { // Attempt to determine set type by first element
-        null -> av(value as Set<String>) // FIXME Is this a bad idea for the empty set case?
+        null -> av(value as Set<String>)
```
ianbotsf, line 67: **Comment**: At runtime, we have no way to determine the generic type of an empty set because of [type erasure](https://kotlinlang.org/docs/generics.html#type-erasure). [DynamoDB's sets are typed](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes.SetTypes) as either strings, numbers, or binary strings, and so we must make a choice about which variant to use when persisting an empty set. There's really no reason to default to string but we have to pick something.

In practice, this should have very little impact. The set variants behave the same and offer similar capabilities for filtering, combining, etc., so it shouldn't make much difference to treat all empty sets as empty string sets in DDB.

hll/ddb-mapper/dynamodb-mapper/common/src/aws/sdk/kotlin/hll/dynamodbmapper/DynamoDbMapper.kt
```diff
@@ -61,8 +61,6 @@ public interface DynamoDbMapper : DynamoDbMapperOperations {
         schema: ItemSchema.CompositeKey<T, PK, SK>,
     ): Table.CompositeKey<T, PK, SK>
 
-    // TODO add multi-table operations like batchGetItem, transactWriteItems, etc.
-
```
ianbotsf, line 65: **Comment**: This work was already completed in https://github.com/aws/aws-sdk-kotlin/pull/1813, https://github.com/aws/aws-sdk-kotlin/pull/1822, https://github.com/aws/aws-sdk-kotlin/pull/1826, and https://github.com/aws/aws-sdk-kotlin/pull/1830.

hll/hll-codegen/src/main/kotlin/aws/sdk/kotlin/hll/codegen/rendering/BuilderRenderer.kt
```diff
@@ -121,7 +121,5 @@ public class BuilderRenderer(
             }
             blankLine()
         }
-
-        // TODO add DSL methods for low-level structure members
```
ianbotsf, line 125: **Comment**: This `TODO` was added when contemplating future DDB APIs like `CreateTable`, which is out of scope for GA. None of the presently-supported operations (e.g., `DeleteItem`, `Query`, `TransactWriteItems`, etc.) have DSL functions which need to be carried forward from their low-level requests or responses.

luigi617 (approved):
Thanks for all the comments!

Issue #1005, opened 2023-07-28
Create a StaticRegionProvider


cloudshiftchris:
### Describe the feature

Create a `StaticRegionProvider` class to explicitly set a region to use.

### Is your Feature Request related to a problem?

Currently Kotlin SDK does not have a mechanism to explicitly set the region via a RegionProvider; this is useful for say development environments (outside of AWS) where the region can't be directly inferred and configuration is integrated with other application configuration (i.e. no using system properties or environment variables)

### Proposed Solution

```
public class StaticRegionProvider(private val region: String) : RegionProvider {
    override suspend fun getRegion(): String = region
}
```

### Describe alternative solutions or features you've considered

See proposed solution.

### Acknowledge

- [ ] I may be able to implement this feature request

### AWS Kotlin SDK version used

029.0-beta

### Platform (JVM/JS/Native)

JVM

### Operating System and version

MacOS

aajtodd:
Is this just a "nice to have" request or is this blocking something for you?

As far as I understand the request you can get this now by wrapping client creation in a way that you consult your own region provider and pass that to client construction.

e.g.

```kotlin
val resolvedRegion = regionProvider.getRegion()
val s3 = S3Client.fromEnvironment{
    region = resolvedRegion
}

```

aajtodd:
> Q: what are RegionProviders used for in Kotlin SDK? They aren't symmetrical with CredentialsProvider behaviour (provider set at construction time, queried for each API call). Kotlin SDK clients look to use a stringly-typed region parameter at construction time.

They're used in `fromEnvironment` when we need to try and auto detect the region that the client should be configured for. Unlike credentials, region is static once set and doesn't change through the lifetime of a client (e.g. region doesn't expire, your EC2 instance isn't going to magically be in a different region, etc). If you need to talk to a different region you would create a new client or use the `withConfig` extension (see developer guide [here](https://docs.aws.amazon.com/sdk-for-kotlin/latest/developer-guide/override-client-config.html)).

RanVaknin:
Hi @cloudshiftchris,

After reviewing this thread with the team, we need more clarification about your use case. 

Since region is static and can be directly set during client construction (e.g. region = "us-east-1"), could you explain why a StaticRegionProvider would be necessary? What specific scenario requires this that can't be handled by the current region parameter?

This would help us better understand if there's a gap in functionality that needs to be addressed.

Thanks.
Ran~

[...]

PR #1926, opened 2026-06-23
feat: implemented StaticRegionProvider class that always returns a static, pre-configured region.

ashishdhingra:
## Issue \#
#1005 

## Description of changes
Implemented `StaticRegionProvider` class that always returns a static, pre-configured region. This is useful for cases where users build custom RegionProviderChain instances and need a composable building block for explicitly-set regions (e.g., development environments where region can't be inferred from the environment)

**NOTE:**
- Region validation is done in the generated client; e.g. from generated `S3Client.kt`:
  ```Kotlin
  override val region: String? = (builder.region ?: runBlocking { builder.regionProvider?.getRegion() ?: resolveRegion() })?.let { validateRegion(it) }
  ```
- `StaticRegionProvider` is not registered in the list of providers in `aws-runtime\aws-config\jvm\src\aws\sdk\kotlin\runtime\region\DefaultRegionProviderChainJVM.kt`. The reason is that `DefaultRegionProviderChain` is meant for auto-detecting region from the environment (e.g. system props, env vars, profile, IMDS). A static provider doesn't fit that chain because it requires a user-supplied value at construction time and would always "win" if placed first, making the other providers pointless. This will be used by users who build their own `RegionProviderChain` and want a `StaticRegionProvider` as one of the composable building blocks, so this is not about modifying the default chain.

___

### DIFF
--- a/aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/region/StaticRegionProviderTest.kt
+++ b/aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/region/StaticRegionProviderTest.kt
@@ -1,0 +1,24 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package aws.sdk.kotlin.runtime.region
+
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class StaticRegionProviderTest {
+    @Test
+    fun providesRegion() = runTest {
+        val provider = StaticRegionProvider("us-west-2")
+        assertEquals("us-west-2", provider.getRegion())
+    }
+
+    @Test
+    fun providesCustomRegion() = runTest {
+        val provider = StaticRegionProvider("eu-central-1")
+        assertEquals("eu-central-1", provider.getRegion())
+    }
+}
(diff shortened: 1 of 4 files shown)


aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/region/StaticRegionProviderTest.kt
```diff
@@ -0,0 +1,24 @@
...
+        val provider = StaticRegionProvider("us-west-2")
+        assertEquals("us-west-2", provider.getRegion())
+    }
+
+    @Test
+    fun providesCustomRegion() = runTest {
+        val provider = StaticRegionProvider("eu-central-1")
+        assertEquals("eu-central-1", provider.getRegion())
+    }
+}
```
ianbotsf, line 24: **Comment**: Thanks for adding new tests!

ianbotsf (changes requested):
The code changes look good but the CI is failing because we updated the public API of a module but didn't persist a new API manifest. Please run:

```sh
./gradlew :aws-runtime:aws-config:apiDump
```

And then add the changed **aws-runtime/aws-config/api/aws-config.api** file to this PR.

For reference, we use [Binary compatibility validator](https://github.com/kotlin/binary-compatibility-validator) to track our public API and to catch any breaking changes we may accidentally slip in. Fortunately, in this case your changes aren't breaking but we still need to maintain the API manifest file and keep it in sync with our code.

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/region/StaticRegionProviderTest.kt
ashishdhingra: Thanks. Done in latest commit.

ianbotsf (approved):

PR #1932, opened 2026-06-26
fix: unblock Signin minimal client from building

ianbotsf:
## Issue \#

Workaround for https://github.com/smithy-lang/smithy/issues/3190

## Description of changes

This change adds a dummy operation to the minimal `SigninClient` generated for `LoginCredentialsProvider`. This is necessary because an upcoming model change adds new endpoint ruleset parameters but the minimal client removes all operations which reference them, causing a failure during Smithy validation as part of codegen. This change can be reverted when https://github.com/smithy-lang/smithy/issues/3190 is resolved.

### DIFF
--- a/aws-runtime/aws-config/build.gradle.kts
+++ b/aws-runtime/aws-config/build.gradle.kts
@@ -213,13 +213,16 @@ smithyBuild {
                 }
             }
 
+            // FIXME `IntrospectOAuth2TokenWithIAM` is not needed for any credentials provider and is only added because
+            //  of https://github.com/smithy-lang/smithy/issues/3190
             transforms = listOf(
                 """
             {
                 "name": "awsSmithyKotlinIncludeOperations",
                 "args": {
                     "operations": [
-                        "com.amazonaws.signin#CreateOAuth2Token"
+                        "com.amazonaws.signin#CreateOAuth2Token",
+                        "com.amazonaws.signin#IntrospectOAuth2TokenWithIAM"
                     ]
                 }
             }


aws-runtime/aws-config/build.gradle.kts
```diff
@@ -213,13 +213,16 @@ smithyBuild {
...
+            // FIXME `IntrospectOAuth2TokenWithIAM` is not needed for any credentials provider and is only added because
+            //  of https://github.com/smithy-lang/smithy/issues/3190
             transforms = listOf(
                 """
             {
                 "name": "awsSmithyKotlinIncludeOperations",
                 "args": {
                     "operations": [
-                        "com.amazonaws.signin#CreateOAuth2Token"
+                        "com.amazonaws.signin#CreateOAuth2Token",
```
ashishdhingra, line 224: Just curious on why are we removing operations from service client! Is test the design choice to selectively exclude operations from credentials service clients? What if customer wants to reference service package and invoke operations manually (Signin client is exception where it provides operations for underlying tooling)?

Or are there full fledged service clients in addition to minimal service clients?

luigi617 (approved):

aws-runtime/aws-config/build.gradle.kts
ianbotsf: The Smithy projections referenced in this build script are used to generate minimal clients for credentials providers only. Because they're bundled in **aws-config** we need to keep them as small as possible, which we accomplish by filtering out all operations except the ones essential for specific credentials providers (e.g., `LoginCredentialsProvider` uses the `CreateOAuth2Token` operation). These **aws-config** clients are `internal`, not `public`, since they're implementation details of the credentials providers. Users cannot replace them with their own instances nor access them directly. These are completely separate from our `public` service clients, which are not bundled in **aws-config**.

ashishdhingra: Sounds fair. Now I recall you mentioning about these minimal clients.

PR #1952, opened 2026-07-15
Unsigned payload over HTTPS

luigi617:
## Summary

- Skip SHA-256 payload hashing for S3 requests over HTTPS by defaulting to `UNSIGNED-PAYLOAD`, consistent with AWS Java SDK v2 and Go SDK v2 behavior
- Add `payloadSigningEnabled` client config option for customers who need to force payload signing

## Motivation

Profiling S3 upload throughput benchmarks (256KiB objects, c256 concurrency) revealed that **SHA-256 payload hashing is the single largest CPU consumer at 11% of total samples**. For every `PutObject` request, `DefaultCanonicalizer.calculateHash()` reads and hashes the entire request body to produce the `x-amz-content-sha256` header value — even though the request is sent over HTTPS where TLS already provides transport-level integrity.

All other major AWS SDKs default to unsigned payload for S3 over HTTPS:
- **Java SDK v2**: `PAYLOAD_SIGNING_ENABLED = false` for all S3 operations ([source](https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/plugins/S3OverrideAuthSchemePropertiesPlugin.java))
- **Go SDK v2**: `dynamicPayloadSigningMiddleware` uses `UnsignedPayload` over HTTPS for upload operations ([source](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/signer/v4/middleware.go))
- **Python (botocore)**: `S3SigV4Auth` skips payload signing for streaming uploads over HTTPS ([source](https://github.com/boto/botocore/blob/develop/botocore/auth.py))

### What changed

**smithy-kotlin (runtime):**

`AwsSigningAttributes` gains a new `PayloadSigningEnabled` attribute key. `AwsHttpSigner.sign()` reads this attribute: when `false` and the request uses HTTPS, it sets `hashSpecification = HashSpecification.UnsignedPayload` instead of computing SHA-256 of the body. The default when the attribute is absent is `true` (always sign), ensuring non-S3 services are unaffected.

**aws-sdk-kotlin (S3 service codegen):**

`ClientConfigIntegration` adds a `payloadSigningEnabled: Boolean` config property (default: `false`). `S3SigningConfig` wires `config.payloadSigningEnabled` into the signing attributes via `putIfAbsent`. This means S3 operations use `UNSIGNED-PAYLOAD` over HTTPS by default, and customers can opt back into payload signing with:

```kotlin
S3Client {
    payloadSigningEnabled = true
}
```

### Cross-SDK references

| SDK | Default for S3 | Customer override | Source |
|---|---|---|---|
| Java v2 | `PAYLOAD_SIGNING_ENABLED = false` | `S3OverrideAuthSchemePropertiesPlugin.enablePayloadSigningPlugin()` | [S3OverrideAuthSchemePropertiesPlugin.java](https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/java/software/amazon/awssdk/services/s3/internal/plugins/S3OverrideAuthSchemePropertiesPlugin.java) |
| Go v2 | `UnsignedPayload` over HTTPS (automatic) | No client-level option | [v4/middleware.go](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/signer/v4/middleware.go) |
| Python | Unsigned for streaming uploads over HTTPS | `Config(s3={'payload_signing_enabled': True})` | [botocore/auth.py](https://github.com/boto/botocore/blob/develop/botocore/auth.py) |

**Safety guarantees:**
- HTTP (non-TLS) requests are **always** signed regardless of this setting — the `Scheme.HTTPS` check ensures payload integrity when TLS is not protecting the transport
- Operations with explicit `@unsignedPayload` trait or `HashSpecification` set in context take precedence (no behavior change)
- The `x-amz-content-sha256: UNSIGNED-PAYLOAD` header is still emitted (required by S3) because `SignedBodyHeader = X_AMZ_CONTENT_SHA256` remains set as a service default

## Benchmark Results

All benchmarks run on m7i.2xlarge instances with the following configuration:
- **Operations per batch:** 100,000
- **Warmup batches:** 3
- **Measurement batches:** 5

### E2E Throughput (S3)

15 paired instances (base and target on same host).

#### S3 Upload
| Concurrency | Base (Gbps) | Target (Gbps) | Change | Base CPU% | Target CPU% |
|---|---|---|---|---|---|
| 16 | 4.02 | 4.25 | **+5.77%** | 26.0 | 22.1 |
| 32 | 7.13 | 7.51 | **+5.32%** | 61.4 | 57.6 |
| 64 | 8.79 | 9.37 | **+6.64%** | 94.6 | 87.8 |
| 128 | 8.53 | 8.91 | **+4.43%** | 96.5 | 90.8 |
| 256 | 8.15 | 8.57 | **+5.20%** | 97.5 | 94.3 |

#### S3 Download
| Concurrency | Base (Gbps) | Target (Gbps) | Change | Base CPU% | Target CPU% |
|---|---|---|---|---|---|
| 16 | 6.17 | 6.14 | -0.52% | 44.9 | 45.7 |
| 32 | 8.66 | 8.70 | +0.46% | 86.6 | 86.2 |
| 64 | 9.57 | 9.40 | -1.81% | 87.2 | 87.5 |
| 128 | 9.62 | 9.81 | +1.98% | 87.2 | 84.7 |
| 256 | 9.56 | 9.37 | -1.94% | 88.0 | 88.2 |

**Throughput observations:**
- S3 Upload shows statistically significant improvement at all concurrency levels (**+5-7%**, t > 3.5 at c16–c64) — directly from eliminating SHA-256 computation of the 256KiB body on every request
- S3 Download shows no meaningful change (expected — GET requests have empty bodies, so payload hashing was already trivial)
- CPU usage is lower in target for upload, consistent with eliminating the hashing work
- No memory regression

### E2E Latency (DynamoDB)

This optimization does not affect DynamoDB — DDB does not set `PayloadSigningEnabled` in its signing attributes, so payloads remain signed (the attribute defaults to `true` when absent). No latency change expected or observed.

---


.changes/e2c31d13-0a8f-4022-b6f9-e0b25ac0f443.json
```diff
@@ -0,0 +1,6 @@
+{
+  "id": "e2c31d13-0a8f-4022-b6f9-e0b25ac0f443",
+  "type": "feature",
+  "description": "Add `payloadSigningEnabled` client config option to S3 (default: `false`) to skip SHA-256 payload hashing over HTTPS, consistent with other AWS SDK behavior",
```
ianbotsf: **Nit**: "SHA-256 payload hashing" is too generic and could be confused with request/response checksumming. I suggest "SigV4 signing" to be extra explicit.

codegen/aws-sdk-codegen/src/main/kotlin/aws/sdk/kotlin/codegen/customization/s3/ClientConfigIntegration.kt
```diff
@@ -66,6 +66,18 @@ class ClientConfigIntegration : KotlinIntegration {
...
+            name = "payloadSigningEnabled"
+            useSymbolWithNullableBuilder(KotlinTypes.Boolean, "false")
+            documentation = """
+                Flag to control whether request payloads are signed (SHA-256 hash computed) before sending.
+                When `false` (default), payload signing is automatically disabled for HTTPS requests to improve
+                performance, since TLS already provides integrity protection. Payloads are always signed for
+                HTTP (non-TLS) requests regardless of this setting. Set to `true` to force payload signing
+                even over HTTPS.
+            """.trimIndent()
+        }
```
ianbotsf, line 82: **Nit**: The wording of this documentation needs some work. In particular:
* This flag does not _fully_ control whether requests are signed. There are other factors too such as `AwsHttpSigner.Config.isUnsignedPayload`.
* The first sentence mentions nothing about HTTPS/TLS and therefore implies the flag is scheme-independent. Readers must get to sentence 3 before discovering this is an HTTPS-only setting.
* Payloads are not _always_ signed over HTTP. Factors such as the auth scheme of the service/operation, the `unsignedPayload` trait, etc. are also taken into account.

ianbotsf (changes requested):
**Correctness**: Please add unit tests.

.changes/e2c31d13-0a8f-4022-b6f9-e0b25ac0f443.json
luigi617: Good point on the checksumming ambiguity, I'd suggest "SigV4 payload siging" as what we are skipping is the hashing they payload and the request is still SigV4-signed

ianbotsf (approved):

PR #1953, opened 2026-07-15
Fix jackson dependency

luigi617:
## Description of changes

Dokka 2.2.0 bundles jackson-databind 2.15.3 contains vulnerabilities. This PR upgrades Jackson to 2.22.1 across both the buildSrc classpath and Dokka's isolated generator runtime.

The previous workaround (`enforcedPlatform` on the root buildscript classpath) was a no-op — nothing on that classpath actually depended on Jackson, and it didn't propagate to Dokka's generator process. This replaces it with a fix that targets the actual affected classpaths:

  1. `buildSrc/build.gradle.kts` — `platform(libs.jackson.bom)` overrides Jackson on the buildSrc classpath where dokka-gradle-plugin pulls it in transitively.
  2. `dokka-convention.gradle.kts` — `resolutionStrategy.eachDependency` overrides Jackson on all Dokka configurations (including the isolated generator process). A `platform()`/`enforcedPlatform()` BOM doesn't work here because Dokka 2.x uses custom Gradle attributes on its configurations that prevent BOM constraints from applying.

  The version catalog accessor is made available in the precompiled script plugin via the [gradle/gradle#15383](https://github.com/gradle/gradle/issues/15383) workaround, using `rootProject.the<LibrariesForLibs>()` since the extension isn't registered on subprojects at the time the convention plugin is applied cross-project.

  This is safe per the [Dokka maintainer's confirmation](https://github.com/Kotlin/dokka/issues/3472#issuecomment-3179718926) that Dokka 2.1.0+ supports updated Jackson versions.


buildSrc/src/main/kotlin/dokka-convention.gradle.kts
```diff
@@ -2,13 +2,16 @@
...
+// https://github.com/gradle/gradle/issues/15383
 import aws.sdk.kotlin.dokka.TrimNavigation
+import org.gradle.accessors.dm.LibrariesForLibs
 
 plugins {
     id("org.jetbrains.dokka")
 }
 
+val libs = rootProject.the<LibrariesForLibs>()
+
```
ianbotsf: **Nit**: Please move the comment to the actual call site at L13.

ianbotsf (approved):
Thanks for adding all the comments, they're really useful when trying to go back and understand/refactor unfamiliar code.

Issue #229, opened 2021-07-08
Kotlin Native Support


aajtodd:
Support for Kotlin Native targets

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue, please leave a comment

## Describe the Feature

Add support for using generated AWS service clients on Kotlin Native targets.
Requires:
* CRT support from aws-crt-kotlin
* Runtime support from smithy-kotlin (notably JSON/XML serde support)
* KMP project generation: https://github.com/awslabs/smithy-kotlin/issues/437

napperley:
There is a typo in the description, **Add support for using generated AWS service clients on Kotlin JS targets**. Should be **Kotlin Native** instead of **Kotlin JS**.

napperley:
Which targets will the Kotlin Native support cover? Does it include **linuxX64**, **linuxArm32Hfp**, and **linuxArm64**?

aajtodd:
@napperley It's still a bit early to know for sure which targets we would initially support. I don't think any of the ones you mentioned are particularly worrisome though.

aajtodd:
To anyone following this issue, we ask that you please take some time to respond to the [discussion](https://github.com/awslabs/aws-sdk-kotlin/discussions/464) on KMP use cases. Thank-you!

AlbertoSH:
Hey, I'd love to see native support for the SDK. That would be a huge game changer when talking about microservices development (my main pain point is that yes, I can go for a native backend framework with Kotlin but I still need the clients for several services to be native)
As you were mentioning, [previous memory manager model has changed](https://blog.jetbrains.com/kotlin/2020/07/kotlin-native-memory-management-roadmap/). A preview of the new model[ was released last year](https://blog.jetbrains.com/kotlin/2021/08/try-the-new-kotlin-native-memory-manager-development-preview) and now [it's the default one for Kotlin 1.7.20](https://kotlinlang.org/docs/native-memory-manager.html) (as of today, latest version)

[...]

PR #1988, opened 2026-09-03
chore: merge main into kn-main

ianbotsf:
## Issue \#

Related to #229 

## Description of changes

This PR merges the latest commits from **main** into **kn-main** in preparation for release of Kotlin/Native support in the SDK. Apologies for the length of the review. All of the base commits from **main** have obviously been reviewed but were a handful of merge conflicts I needed to resolve carefully. I recommend paying extra attention to:
* **tests/e2e-test-util** `withAllEngines` KMP contract
* **aws-config** `DefaultRegionProviderChainTest` common suite
* **aws-config** `runRegionProviderChainTestSuite` helper
* S3 e2e **S3TestUtils.kt**


ashishdhingra:
> * **tests/e2e-test-util** `withAllEngines` KMP contract
> * **aws-config** `DefaultRegionProviderChainTest` common suite
> * **aws-config** `runRegionProviderChainTestSuite` helper
> * S3 e2e **S3TestUtils.kt**

There files look fine to me. I will try to glance at other files as much as possible.

services/s3/e2eTest/src/ConnectionResetTest.kt
```diff
@@ -25,29 +23,32 @@ import kotlin.time.Duration.Companion.seconds
  * Reproduces "unexpected end of stream" errors as seen in https://github.com/aws/aws-sdk-kotlin/issues/1214
  * and ensures they are resolved by OkHttp's retryOnConnectionFailure option
  */
-@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@TestInstance(TestLifecycle.PER_CLASS)
 class ConnectionResetTest {
```
luigi617: This class is declared twice, one in `jvmMain` (this one) and one in `commonMain`. The latter is using some non-existing function, e.g. `getOrCreateSharedBucket`. This is why e2e-tests is failing

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3PresignerTest.kt
```diff
@@ -63,7 +57,7 @@ class S3PresignerTest {
             }
             val presignedPutRequest = client.presignPutObject(unsignedPutRequest, 60.seconds)
 
-            responseCodeFromPut(engine, presignedPutRequest, contents)
+            S3TestUtils.responseCodeFromPut(presignedPutRequest, contents)
```
luigi617: this is also failing for e2e-tests

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3ChecksumTest.kt
```diff
@@ -14,38 +14,38 @@ import aws.smithy.kotlin.runtime.hashing.crc32
...
+    private fun testKey(suffix: String): String = "test-object-$suffix"
 
-        @AfterAll
-        @JvmStatic
-        fun cleanup() = runBlocking {
-            S3TestUtils.deleteSharedBucket(client)
-            client.close()
-        }
+    @BeforeAll
+    private fun setUp(): Unit = runBlocking {
```
luigi617: correctness: `@BeforeAll` should not be used with private function. I think this will be ignored (https://docs.junit.org/5.5.1/api/org/junit/jupiter/api/BeforeAll.html)
The same comment applies to `@AfterAll`.

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/region/DefaultRegionProviderChainTest.kt
```diff
@@ -83,6 +151,20 @@ private const val REGION_PROVIDER_CHAIN_TEST_SUITE = """
...
+            "env": {
+                "AWS_REGION": "us-east-2"
+            },
+            "props": {},
+            "fs": {}
+        },
+        "imds": {
+            "/latest/meta-data/placement/region": "us-east-1"
+        },
+        "region": "us-east-2"
```
luigi617: Correctness: this test is basically the same as the next one "environment variables are favored over imds" and it is not testing what it claims.

Do we really need this test? doen't the test in `DefaultRegionProviderChainTestJVM.kt` already cover it?

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/region/DefaultRegionProviderChainTest.kt
```diff
@@ -17,6 +18,73 @@ class DefaultRegionProviderChainTest {
...
+    val targets: List<String> = emptyList(),
+)
+
+/**
+ * Parse a region-provider-chain test suite from a JSON string and run its assertions.
+ *
+ * This is a shared, reusable test runner so that both the common test suite and platform-specific
+ * test suites (e.g. JVM) can exercise the same machinery without duplicating logic.
+ */
+fun runRegionProviderChainTestSuite(testSuiteJson: String) = runTest {
```
luigi617: nit: this function had visibility `internal`, any reason why its `public` now?

tests/e2e-test-util/jvm/src/kotlin/aws/sdk/kotlin/testing/UtilsJvm.kt
```diff
@@ -5,23 +5,22 @@
...
-        OkHttp4Engine(),
-        CrtHttpEngine(),
+/**
+ * Run the [block] with each supported engine
+ */
+actual suspend fun withAllEngines(block: suspend (HttpEngineContext) -> Unit) {
+    val engines: List<Pair<String, CloseableHttpClientEngine>> = listOf(
+        "Default" to DefaultHttpEngine(),
+        "CRT" to CrtHttpEngine(),
     )
```
luigi617, line 23: Question: The new list drops Okhttp4, is there any reason we are excluding Okhttp4?

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3IntegrationTest.kt
```diff
@@ -18,35 +18,31 @@ import aws.smithy.kotlin.runtime.io.use
...
-        @AfterAll
-        @JvmStatic
-        fun cleanup(): Unit = runBlocking {
-            S3TestUtils.deleteSharedBucket(client)
-            client.close()
-        }
+    @AfterAll
+    fun cleanup() = runBlocking {
+        S3TestUtils.deleteBucket(client, testBucket)
     }
```
luigi617, line 50: Nit: We should close `client` when cleanup

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/MultiRegionAccessPointTest.kt
```diff
@@ -4,94 +4,107 @@
...
+            }
         }
+        print(mrapManifest)
+
+        S3TestUtils.deleteBucket(s3West, usWestBucket)
+        S3TestUtils.deleteBucket(s3East, usEastBucket)
+
+        s3West.close()
+        s3East.close()
+        s3Control.close()
```
luigi617: Any error that happens before it will throw and `deleteBucket` won't be called

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/MultiRegionAccessPointTest.kt
```diff
@@ -4,94 +4,107 @@
...
-            testMultiRegionAccessPointOperation(signer)
+                val regions = accessPoint.regions.orEmpty()
+                appendLine("  * Regions (${regions.size}):")
+
+                regions.forEach { region ->
+                    appendLine("    * ${region.region}: ${region.bucket} (account ID ${region.bucketAccountId})")
+                }
+            }
         }
+        print(mrapManifest)
```
luigi617: Nit: is this print a leftover?

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/MultiRegionAccessPointTest.kt
```diff
@@ -4,94 +4,107 @@
...
+        s3West.close()
+        s3East.close()
+        s3Control.close()
     }
 
-    private suspend fun testMultiRegionAccessPointOperation(signer: AwsSigner) {
-        println("Testing multi-region access point operations with $signer")
+    @Test
+    fun testMultiRegionAccessPointOperation(): Unit = parameterized(
+        listOf(DefaultAwsSigner, CrtAwsSigner),
```
luigi617: Nit: previously we have `distinct()`, now that it is removed, this can be ran twice in K/N

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3TestUtils.kt
```diff
@@ -19,161 +18,106 @@ import aws.smithy.kotlin.runtime.http.complete
...
-            .buckets
-            ?.mapNotNull { it.name }
-            ?.firstOrNull { it.startsWith(TEST_BUCKET_PREFIX) && it.endsWith(S3_EXPRESS_DIRECTORY_BUCKET_SUFFIX) }
-
-        if (testBucket == null) {
-            // Adding S3 Express suffix surpasses the bucket name length limit... trim the UUID if needed
-            testBucket = TEST_BUCKET_PREFIX +
-                Uuid.random().toString().subSequence(0 until (S3_MAX_BUCKET_NAME_LENGTH - TEST_BUCKET_PREFIX.length - suffix.ensurePrefix("--").length)) +
-                suffix.ensurePrefix("--")
-
```
luigi617, line 148: Nit: This check is removed in the new version (in `createDirectoryBucket`), I think this is still needed or we might need a length check.

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/region/DefaultRegionProviderChainTest.kt
ianbotsf: Overly eager agent. Restored to `public`.

aws-runtime/aws-config/common/test/aws/sdk/kotlin/runtime/region/DefaultRegionProviderChainTest.kt
ianbotsf: Good point, we cannot test JVM system properties in **common**. Removed test.

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/MultiRegionAccessPointTest.kt
ianbotsf: It's debugging info for the MRAP operation. When creating MRAP endpoints fails, it can be tricky to determine what the state was without it.

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/MultiRegionAccessPointTest.kt
ianbotsf: Will fix.

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/MultiRegionAccessPointTest.kt
ianbotsf: Will fix.

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3ChecksumTest.kt
ianbotsf: Overeager agents again. Reverting to `public`.

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3IntegrationTest.kt
ianbotsf: Will do!

tests/e2e-test-util/jvm/src/kotlin/aws/sdk/kotlin/testing/UtilsJvm.kt
ianbotsf: Nope we should've kept it. Reverting.

services/s3/e2eTest/src/ConnectionResetTest.kt
ianbotsf: Yep, missed this one in my refactor. Removing the old one in favor of the new one.

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3TestUtils.kt
ianbotsf: Readding a length check.

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3PresignerTest.kt
ianbotsf: Fixed!

services/dynamodb/e2eTest/src/PaginatorTest.kt
```diff
@@ -0,0 +1,122 @@
+/*
```
ashishdhingra: **[medium] Resurrected orphan file - this test is never compiled.**

This file comes back via a modify/delete conflict resolution. `kn-main` had moved this test to `services/dynamodb/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/dynamodb/PaginatorTest.kt`; `main` modified it at the old path, and the merge kept both copies.

`services/build.gradle.kts` registers only `kotlin.srcDir("e2eTest/src/commonMain")` for the e2e source set, so nothing under `e2eTest/src/*.kt` is compiled or run. Concretely: `main`'s improvements here (`@TestInstance(TestLifecycle.PER_CLASS)`, `runTest(timeout = 20.seconds)`, `Random.nextInt()`-based table name) silently never reach the test that actually executes, which still uses `Uuid.random()` and `@BeforeTest`/`@AfterTest`. And because the file never compiles, it will rot with no build signal.

This is the exact analogue of `services/s3/e2eTest/src/ConnectionResetTest.kt`, which was correctly deleted in the follow-up commit - this one was missed. Suggest deleting this file and porting `main`'s changes into the `commonMain` copy.

services/kinesis/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/kinesis/KinesisSubscribeToShardTest.kt
```diff
@@ -17,181 +22,159 @@ import kotlin.time.Duration.Companion.seconds
...
 
 private val TEST_DATA = "Bees, bees, bees, bees!"
 
 /**
  * Tests for Kinesis SubscribeToShard (an RPC-bound protocol)
  */
+@TestInstance(TestLifecycle.PER_CLASS)
 class KinesisSubscribeToShardTest {
-    private lateinit var client: KinesisClient
+    private val client = KinesisClient { region = "us-east-1" }
```
ashishdhingra: **[medium] `client` is now never closed - `client.close()` was lost in the restructure.**

The previous version created the client in `@BeforeTest` and closed it in `@AfterTest cleanUp()` (`client.close()` was the last statement). The rewrite moves the client to a `@TestInstance(PER_CLASS)` property but adds no `@AfterAll`, so the `KinesisClient` and its HTTP engine / coroutine scope are never released.

Scenario: the e2e suite finishes and the engine's connection pool and dispatcher threads are still live. On Kotlin/Native this leaks CRT resources and can produce a leak warning or a hang at test-process exit - precisely the platform this branch is adding support for.

Suggest an `@AfterAll fun cleanUp() { client.close() }`, or wrapping the test body in `client.use { }`.

ashishdhingra (commented):
Reviewed the conflict resolutions in this merge (`git show --cc` on the merge commit) plus the follow-up fix commit, skipping generated `codegen/sdk/aws-models/*`, `CHANGELOG.md`, and `endpoints.json`. Three things worth a look before merging:

**1. `services/dynamodb/e2eTest/src/PaginatorTest.kt` is an orphan resurrected by the merge.** `kn-main` had moved this test to `e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/dynamodb/PaginatorTest.kt`; `main` modified the old path, and the modify/delete conflict resolution kept both. `services/build.gradle.kts` only registers `kotlin.srcDir("e2eTest/src/commonMain")`, so this copy is never compiled or run — `main`'s improvements (`PER_CLASS` test lifecycle, `runTest(timeout = 20.seconds)`, `Random.nextInt()` table name) never reach the live test, which still uses `Uuid.random()` and `@BeforeTest`/`@AfterTest`. Same case as `services/s3/e2eTest/src/ConnectionResetTest.kt`, which was deleted in the follow-up commit. Suggest deleting this file and porting `main`'s changes into the commonMain copy.

**2. `KinesisSubscribeToShardTest` no longer closes its client.** In `services/kinesis/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/kinesis/KinesisSubscribeToShardTest.kt`, `client` became a `PER_CLASS` property but the `@AfterTest cleanUp` that called `client.close()` was dropped with no `@AfterAll` replacement. The client and its HTTP engine / coroutine scope are never closed, which on Kotlin/Native can leak CRT resources and hang or warn at test-process exit.

**3. `testPutObjectAwsChunkedEncoded` no longer exercises chunked encoding.** In `services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3ChecksumTest.kt`, the conflict resolution replaced the temp-file / `asByteStream` body with `ByteStream.fromString(testBody)`, making the test byte-for-byte identical to `testPutObject` above it. `aws-chunked` streaming-payload signing and trailing checksums are left without e2e coverage.

Things I looked at and believe are fine: the follow-up commit correctly restores OkHttp/OkHttp4/CRT coverage in `withAllEngines` that the merge had collapsed; the `ImdsCredentialsProviderTestJvm` filename/class mismatch is harmless (`ktlint_standard_filename` is disabled and there's no redeclaration); and the removed common "jvm system properties are favored over imds" region case is now covered properly by the JVM suite with real system properties.

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3ChecksumTest.kt
```diff
@@ -84,8 +84,8 @@ class S3ChecksumTest {
     }
 
     @Test
-    fun testPutObjectAwsChunkedEncoded() = runBlocking {
-        val testKey = testKey()
+    fun testPutObjectAwsChunkedEncoded(): Unit = runBlocking {
```
ashishdhingra: **[medium] `testPutObjectAwsChunkedEncoded` no longer tests aws-chunked encoding - it is now identical to `testPutObject`.**

The conflict resolution dropped the `File.createTempFile` / `FileInputStream` / `file.asByteStream(...)` setup and replaced the body with `body = ByteStream.fromString(testBody)`. That makes this test byte-for-byte the same request as `testPutObject` immediately above: a small, fully-buffered, known-length payload, which is *not* sent with `aws-chunked` encoding.

Result: `aws-chunked` streaming-payload signing and trailing checksums are no longer covered by any e2e test, so a regression there passes CI green while the test name claims otherwise. Restoring coverage needs a streaming body of unknown/streaming length (e.g. `RandomTempFile(...)` exposed as a byte stream via the multiplatform file APIs available on this branch).

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3TestUtils.kt
```diff
@@ -217,6 +165,18 @@ object S3TestUtils {
         }
     }
 
+    private suspend fun deleteMultiPartUploads(client: S3Client, bucketName: String) {
```
ashishdhingra: **[low] `listMultipartUploads` is not paginated - bucket teardown can leave uploads behind.**

`client.listMultipartUploads { bucket = bucketName }.uploads?.forEach { ... }` returns at most 1000 uploads in one page. Multipart tests now run once per engine under `s3WithAllEngines` and abandon parts whenever a test fails midway, so uploads accumulate.

Scenario: a bucket with >1000 in-progress uploads keeps the overflow after this call, the following `client.deleteBucket` fails with `BucketNotEmpty`, and the test bucket leaks - permanently, since nothing else reaps it, and each subsequent run adds another.

Suggest `client.listMultipartUploadsPaginated { bucket = bucketName }` (or honouring `isTruncated` / `nextKeyMarker`).

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3TestUtils.kt
```diff
@@ -217,6 +165,18 @@ object S3TestUtils {
...
+        }.uploads?.forEach { upload ->
+            client.abortMultipartUpload {
+                bucket = bucketName
+                key = upload.key
+                uploadId = upload.uploadId
+            }
+        }
+    }
+
     suspend fun responseCodeFromPut(engine: HttpClientEngine, presignedRequest: HttpRequest, content: String): Int {
```
ashishdhingra: **[low] Failure diagnostics dropped: the response body is no longer surfaced.**

The JVM version this replaces did `error("request failed: ${connection.errorStream?.bufferedReader()?.readText()}")` on a non-2xx response. The rewrite returns only `call.response.status.value`.

Scenario: a presigned PUT fails with `SignatureDoesNotMatch` or `XAmzContentSHA256Mismatch`. The callers in `S3ChecksumTest` only do `assertTrue(responseCode in 200..299)`, so CI reports a bare "expected value to be true" - no status code, no S3 error XML. Debugging a presigner regression then requires reproducing locally.

Suggest including the status code and body in the assertion message, or throwing with the body when the status is not 2xx.

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3ExpressTest.kt
```diff
@@ -16,34 +16,41 @@ import aws.smithy.kotlin.runtime.http.request.HttpRequest
...
+
+    @BeforeAll
+    fun setup(): Unit = runBlocking {
+        // create a few test buckets to test the credentials cache
+        testBuckets = (1..3).map { index ->
+            S3TestUtils.createTestDirectoryBucket(client, AVAILABILITY_ZONE, index.toString())
         }
     }
 
+    @AfterAll
```
ashishdhingra: **[low] `@AfterAll` dereferences a `lateinit` property with no guard, masking setup failures and skipping `client.close()`.**

If `@BeforeAll` throws - e.g. `createTestDirectoryBucket` fails, or the `require(... <= S3_MAX_BUCKET_NAME_LENGTH)` guard added in the follow-up commit trips - `testBuckets` is never initialised. `@AfterAll` then throws `UninitializedPropertyAccessException`, which replaces the real setup error in the report and aborts before `client.close()`.

Same shape in `S3ChecksumTest`'s `@AfterAll` (`testBucket`). Note that `MultiRegionAccessPointTest` was given `try`/`finally` + `runCatching` for exactly this reason, and `S3BucketOpsIntegrationTest` got `try { deleteBucket(...) } finally { client.close() }` - worth applying the same pattern here.

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3ChecksumTest.kt
```diff
@@ -121,21 +122,23 @@ class S3ChecksumTest {
         val fileBytes = file.readBytes()
```
ashishdhingra: **[low] The 8 MB temp file is read fully into memory twice, and the `assertEquals` below has its arguments swapped.**

1. `file.readBytes()` already ran a few lines up to compute `expectedChecksum`; this reads the same 8 MB again. Hoisting a single `val fileBytes = file.readBytes()` and deriving `expectedChecksum` from it halves the allocation.
2. Further down in this test: `assertEquals(actualChecksum, expectedChecksum)`. `assertEquals` takes `(expected, actual)`, so on a checksum mismatch the failure message labels the two values backwards and points the reader at the wrong side. (Pre-existing line, but adjacent to what changed here.)

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3IntegrationTest.kt
```diff
@@ -175,6 +175,7 @@ class S3IntegrationTest {
 
             val fileBytes = file.readBytes()
```
ashishdhingra: **[low] The 8 MB temp file is read fully into memory twice.**

`file.readBytes()` already ran above to compute `expectedSha256`; this reads the same 8 MB again. Because `testMultipartUpload` now runs once per engine via `s3WithAllEngines` (OkHttp, OkHttp4, CRT), that is ~48 MB of avoidable allocation per run instead of ~24 MB. Suggest hoisting a single `val fileBytes = file.readBytes()` and computing `expectedSha256` from it.

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3TestUtils.kt
```diff
@@ -19,161 +18,110 @@ import aws.smithy.kotlin.runtime.http.complete
...
+        region: String = client.config.region!!,
+    ): String = createBucket(client, TEST_BUCKET_PREFIX, suffix, region)
 
-    suspend fun deleteSharedBucket(client: S3Client) {
-        sharedBucket?.let { bucket ->
-            deleteBucketAndAllContents(client, bucket)
-            sharedBucket = null
-        }
-    }
+    suspend fun createBucket(
```
ashishdhingra: **[low] `LocationConstraint` is sent unconditionally, which breaks `us-east-1`.**

`createBucket` always emits `createBucketConfiguration { locationConstraint = BucketLocationConstraint.fromValue(region) }`. S3 rejects an *explicit* `us-east-1` location constraint with `InvalidLocationConstraint` - it must be omitted for that region.

Today every caller uses `us-west-2` / `us-east-2`, so this is latent. But `us-east-1` is already the region used by `PollyPresignerTest` and `KinesisSubscribeToShardTest`, so the first S3 test that follows suit fails at bucket creation with an error that looks unrelated to the region argument. Suggest skipping the configuration when `region == "us-east-1"`.

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3TestUtils.kt
```diff
@@ -19,161 +18,110 @@ import aws.smithy.kotlin.runtime.http.complete
...
-import kotlinx.coroutines.sync.Mutex
-import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withTimeout
 import kotlin.time.Duration.Companion.seconds
 
 object S3TestUtils {
-
     const val DEFAULT_REGION = "us-west-2"
 
     // The E2E test account only has permission to operate on buckets with the prefix "s3-test-bucket-"
```
ashishdhingra: **[low] Comment and constant no longer agree.**

The comment says the E2E account only has permission for the prefix `"s3-test-bucket-"` (trailing hyphen), but `TEST_BUCKET_PREFIX` is now `"s3-test-bucket"`. It happens to still work because names are assembled as `"$prefix-$testRunId-$suffix"`, which re-inserts the hyphen.

The trap is the `prefix` parameter of `createBucket`: someone reading this comment will pass `"s3-test-bucket-"` and get `"s3-test-bucket--<runId>-..."`, or will assume another prefix is fine and get an `AccessDenied` that this comment does not explain. Worth aligning the comment with the constant and noting that the hyphen is added by the name builder.

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/MultiRegionAccessPointTest.kt
```diff
@@ -79,19 +97,17 @@ class MutliRegionAccessPointTest {
...
-    private suspend fun testMultiRegionAccessPointOperation(signer: AwsSigner) {
-        println("Testing multi-region access point operations with $signer")
+    fun testMultiRegionAccessPointOperation(): Unit = parameterized(
+        // .distinct() deduplicates DefaultAwsSigner and CrtAwsSigner, which are the same on Native
+        listOf(DefaultAwsSigner, CrtAwsSigner).distinct(),
+    ) { signer ->
+        runBlocking {
+            println("Testing multi-region access point operations with $signer")
+
+            val s3SigV4a = s3West.withConfig {
```
ashishdhingra: **[low] `s3SigV4a` is never closed.**

`s3West.withConfig { ... }` returns a new client that owns its own resources; every other `withConfig` in this PR is wrapped in `.use { }` (`S3IntegrationTest`, `S3ExpressTest`, `KinesisSubscribeToShardTest`). Here it leaks for the lifetime of the test process - and once per signer, since this runs inside the `listOf(DefaultAwsSigner, CrtAwsSigner)` loop. Pre-existing, but the surrounding cleanup was reworked in this diff, so it is cheap to fix while here.

services/polly/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/polly/PollyPresignerTest.kt
```diff
@@ -29,17 +31,16 @@ class PollyPresignerTest {
             text = "hello world"
         }
 
-        val presignedRequest = PollyClient { region = "us-east-1" }.use { client ->
-            client.presignSynthesizeSpeech(unsignedRequest, 10.seconds)
-        }
+        val client = PollyClient { region = "us-east-1" }
```
ashishdhingra: **[low] The client is never closed.**

This changed from `PollyClient { region = "us-east-1" }.use { client -> ... }` to a bare `val client = ...`, and the `aws.smithy.kotlin.runtime.io.use` import was removed alongside it, so nothing releases the client's HTTP engine or coroutine scope.

Scenario: the test passes, then the engine's connection pool and dispatcher threads stay alive for the rest of the test process. On Kotlin/Native this leaks CRT resources and can produce a leak warning or hang at exit - same issue as the Kinesis test in this PR. Restoring the `use { }` wrapper (or adding an explicit `close()`) fixes it.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/config/retries/ResolveRetryStrategy.kt
```diff
@@ -20,34 +21,111 @@ import aws.smithy.kotlin.runtime.retries.StandardRetryStrategy
...
     }
+
+    enableLongPollingBackoff = true
+
+    delayProvider {
+        initialDelay = defaultInitialDelay ?: STANDARD_INITIAL_DELAY
+        scaleFactor = STANDARD_SCALE_FACTOR
+    }
+
+    tokenBucket {
```
ashishdhingra: **[low] `STANDARD_THROTTLING_RETRY_COST` is assigned to `timeoutRetryCost` - one of the two is wrong.**

```kotlin
tokenBucket {
    retryCost = STANDARD_RETRY_COST                    // 14
    timeoutRetryCost = STANDARD_THROTTLING_RETRY_COST  // 5
}
```

A constant named `..._THROTTLING_RETRY_COST` is being used for the *timeout* retry cost, and no throttling-specific knob is configured at all. Either the constant is misnamed or the wrong builder property is being set.

If it is the latter, this is a behaviour bug for every client with `AWS_NEW_RETRIES_2026` enabled: the values are inverted relative to the AWS retry standard (retry cost 5, timeout retry cost 10), so an ordinary retry drains the token bucket ~3x faster than intended. Scenario: a client hitting sustained non-timeout retryable errors exhausts capacity and starts failing fast with `RetryCapacityExceeded` well before the configured `maxAttempts`. Worth double-checking against the New Retry Behavior spec.

ianbotsf:
Thanks for the review, @ashishdhingra. I've addressed each of your comments:

> 1. `services/dynamodb/e2eTest/src/PaginatorTest.kt` is an orphan resurrected by the merge.

Good catch, removed.

> 2. `KinesisSubscribeToShardTest` no longer closes its client.

Restored client closing.

3. `testPutObjectAwsChunkedEncoded` no longer exercises chunked encoding.

Swapped from using `ByteStream.fromString` (ineligible for `aws-chunked`) to `ByteStream.SourceStream` (eligible for `aws-chunked`).

services/s3/e2eTest/src/commonMain/kotlin/aws/sdk/kotlin/services/s3/S3TestUtils.kt
```diff
@@ -19,161 +18,110 @@ import aws.smithy.kotlin.runtime.http.complete
...
 
-    suspend fun getOrCreateSharedBucket(client: S3Client, region: String = DEFAULT_REGION): String = sharedBucket ?: bucketMutex.withLock {
-        sharedBucket ?: getTestBucket(client, region).also { sharedBucket = it }
+    val testRunId by lazy {
+        Instant
+            .now()
+            .format(TimestampFormat.ISO_8601_CONDENSED)
+            .lowercase()
+            .also { println("Starting test run ID $it") }
     }
```
luigi617, line 49: Question: if K/N and JVM start at the same time (same second, rare but can happen), then the runId would be the same for both test and if one finishes early, the bucket will be deleted before the other ends. It feels like we might need to add some random number to it to guarantee uniqueness.

luigi617 (approved):
Left a comment, everything else LGTM

ashishdhingra:
Re-checked at `009a4008`. Thanks for the fixes — the `PaginatorTest.kt` deletion/port and the Kinesis `client.close()` both look right. One follow-up on the third one:

**`testPutObjectAwsChunkedEncoded` still doesn't take the aws-chunked path.** The new body is a `ByteStream.SourceStream` over an 11-byte array with `override val isOneShot: Boolean = false`. The eligibility gate in smithy-kotlin 1.7.9 (pinned at `gradle/libs.versions.toml:16`) is:

```kotlin
public val HttpBody.isEligibleForAwsChunkedStreaming: Boolean
    get() = (this is HttpBody.SourceContent || this is HttpBody.ChannelContent) &&
        contentLength != null &&
        (isOneShot || contentLength!! > AWS_CHUNKED_THRESHOLD)   // 1 MiB
```

With `isOneShot = false` and 11 bytes, both disjuncts are false, so the body isn't eligible and behaviour is the same as before the change: `AwsHttpSigner` falls through to `HashSpecification.CalculateFromPayload` (plain SigV4), and `FlexibleChecksumsRequestInterceptor` takes the precomputed-header branch rather than `calculateAwsChunkedStreamingChecksum`, so trailing checksums stay uncovered too. The code compiles and the test passes — there's no build error to chase.

`ByteStream.isOneShot` already defaults to `true`, and `true` alone satisfies the gate, so dropping the `isOneShot` override should be enough. A payload over 1 MiB would also work. Flagging it mainly because the new comment above the body now states the request *is* chunk-eligible, which would discourage anyone from re-checking.

ianbotsf:
> `testPutObjectAwsChunkedEncoded` still doesn't take the aws-chunked path

You're correct, I'm switching back to `isOneShot = true` with 11 bytes to get the chunking behavior.

ashishdhingra:
## Re-review at `797fd3f5` - aws-chunked fix confirmed correct

`isOneShot = true` satisfies the gate: `SourceContent` && `contentLength != null` && `(isOneShot || ...)` -> eligible. Both consumers now take the chunked path - `AwsHttpSigner.kt:174` selects `StreamingAws4HmacSha256PayloadWithTrailers`, and `FlexibleChecksumsRequestInterceptor.kt:69` routes to `calculateAwsChunkedStreamingChecksum`. `setAwsChunkedHeaders()` will set `Content-Encoding: aws-chunked` / `Transfer-Encoding: chunked` / `X-Amz-Decoded-Content-Length` and drop `Content-Length`. The rewritten comment is accurate on every claim it makes. All three of my original medium findings are now resolved.

Below is everything still open, grouped by severity, with an explicit read on whether it is mandatory and whether it leaks - since that was the question that came up.

**Bottom line: nothing still open is merge-blocking.** Every remaining item except one lives in `e2eTest` source sets, which are test-only compilations that never ship in the published artifacts, and all of it compiles and passes today. The sole exception is `ResolveRetryStrategy.kt:123`, which is shipped production code - but it came in from `main` rather than from this merge, so it should be a separate ticket rather than a gate on this branch.

---

### Medium - cumulative consequences, worth fixing

These leak **AWS resources**, not just memory, and get worse every CI run.

**1. `S3TestUtils.kt:168` - unpaginated `listMultipartUploads` leaks buckets permanently.**
`client.listMultipartUploads { bucket = bucketName }.uploads?.forEach { ... }` returns at most 1000 uploads per page. Multipart tests now run once per engine under `s3WithAllEngines` and abandon parts on mid-test failure, so past 1000 in-progress uploads the overflow survives, the subsequent `deleteBucket` fails with `BucketNotEmpty`, and the bucket is never reclaimed - nothing else reaps it, so each affected run adds another. Worth noting the lifecycle rule `createBucket` installs (1-day expiration) empties objects but does **not** delete the bucket shell, and S3 caps buckets per account, so this accumulates toward a hard limit. Suggest `listMultipartUploadsPaginated`.

**2. `S3ExpressTest.kt:46` and `S3ChecksumTest.kt:41` - a partial `@BeforeAll` failure leaks both buckets and the client.**
`@AfterAll` dereferences `lateinit` `testBuckets` / `testBucket` with no guard, so if setup throws partway (e.g. the `S3_MAX_BUCKET_NAME_LENGTH` `require` trips), cleanup throws `UninitializedPropertyAccessException` *before* reaching either the bucket deletion or `client.close()` - and that exception also replaces the real setup error in the report. Worse for S3 Express specifically: `createDirectoryBucket` installs no lifecycle rule at all (only `createBucket` does), so leaked directory buckets retain their objects indefinitely rather than self-emptying after a day. `MultiRegionAccessPointTest` and `S3BucketOpsIntegrationTest` already use the `try`/`finally` pattern that fixes this.

**3. `ResolveRetryStrategy.kt:123` - the only shipped-code item, and the only one I would call genuinely uncertain.**
`tokenBucket { retryCost = STANDARD_RETRY_COST /* 14 */; timeoutRetryCost = STANDARD_THROTTLING_RETRY_COST /* 5 */ }` assigns a constant named for *throttling* to the *timeout* retry cost, and sets no throttling-specific knob at all. Either the constant is misnamed or the wrong property is being configured; if the latter, the values are inverted relative to the AWS retry standard (retry cost 5, timeout retry cost 10), so with `AWS_NEW_RETRIES_2026` enabled an ordinary retry drains the token bucket ~3x faster than intended and a client hitting sustained retryable errors would trip `RetryCapacityExceeded` well before its configured `maxAttempts`. This is pre-existing from `main`, so it is explicitly **not** a regression in this PR - it just needs a check against the New Retry Behavior spec by someone on that work.

### Low - real but bounded

**4. `PollyPresignerTest.kt:34` - client never closed (in-process leak).**
Changed from `PollyClient { ... }.use { client -> ... }` to a bare `val client = ...` with the `aws.smithy.kotlin.runtime.io.use` import removed. The client owns a managed default engine, so its connection pool and coroutine scope leak outright. Now inconsistent with `KinesisSubscribeToShardTest`, which was just fixed for exactly this.

**5. `MultiRegionAccessPointTest.kt:107` - unclosed derived client keeps the *parent's* engine alive.**
`s3SigV4a = s3West.withConfig { ... }` is never closed. Generated clients register the engine in an `SdkManagedGroup` and `share()` it at init, releasing via `unshareAll()` in `close()`, and `SdkManagedBase.unshare()` only tears the resource down when the count reaches zero. Since this child never closes, its `share()` is never released and the count stays above zero - so `s3West`'s engine is never actually torn down even when `s3West.close()` runs, twice over given the `listOf(DefaultAwsSigner, CrtAwsSigner)` loop. Harmless at JVM process exit; on Kotlin/Native, unreleased CRT resources are exactly what produce leak warnings or exit hangs.

**6. `S3ChecksumTest.kt:89` - the chunked test still cannot detect its own regression.**
It only round-trips the object and compares body content, which is identical whether or not aws-chunked was used - which is precisely why both earlier attempts at this fix compiled and passed while covering nothing. If smithy-kotlin's eligibility rule changes (plausible: the `isOneShot` disjunct looks more like an implementation artifact than the intent, while the 1 MiB threshold looks deliberate) this silently reverts to plain SigV4 and stays green. The repo already has the pattern - `S3ExpressTest.kt:192` `S3ExpressInvocationTrackingInterceptor`, `:202` `CRC32ChecksumValidatingInterceptor`, `S3IntegrationTest.kt:266` `WriteGetObjectResponseHostInterceptor`. A few-line `HttpInterceptor` asserting `Content-Encoding: aws-chunked` and the presence of `x-amz-trailer` on the transmitted request would make the coverage self-verifying.

**7. `S3TestUtils.kt:52` - unconditional `LocationConstraint` breaks `us-east-1`.**
S3 rejects an *explicit* `us-east-1` constraint with `InvalidLocationConstraint`. Latent today (all callers use `us-west-2`/`us-east-2`), but `us-east-1` is already the region the Polly and Kinesis e2e tests use, so the first S3 test to follow suit fails at bucket creation with an error that looks unrelated to the region argument.

**8. `S3TestUtils.kt:180` - `responseCodeFromPut` drops the error body.**
The JVM version it replaced did `error("request failed: ${connection.errorStream?...readText()}")`. Callers only assert `responseCode in 200..299`, so a `SignatureDoesNotMatch` or `XAmzContentSHA256Mismatch` failure reports a bare "expected value to be true" with no status code and no S3 error XML, making a presigner regression undiagnosable from CI logs alone.

**9. `services/dynamodb/e2eTest/src/DynamoDbTestUtils.kt` - now the last dead file at the non-compiled path.**
Its `internal suspend fun DynamoDbClient.tableExists` was inlined as a `private` helper into the live `PaginatorTest.kt:127`, so nothing references it and nothing compiles it. Deleting it finishes the sweep the `PaginatorTest` fix started.

### Nits - no functional impact

**10. `S3TestUtils.kt:32`** - comment says the account only permits prefix `"s3-test-bucket-"` (trailing hyphen) while `TEST_BUCKET_PREFIX` is now `"s3-test-bucket"`; it only works because names are built as `"$prefix-$testRunId-$suffix"`. The trap is the `prefix` parameter of `createBucket`.
**11. `S3ChecksumTest.kt:135` and `S3IntegrationTest.kt:176`** - the 8 MB `RandomTempFile` is read fully into memory twice (once for the expected checksum/hash, again for chunking). Transient allocation, so not a leak; in `S3IntegrationTest` it runs per engine, ~48 MB instead of ~24 MB. Hoist one `val fileBytes = file.readBytes()`.
**12. `S3ChecksumTest.kt:172`** - `assertEquals(actualChecksum, expectedChecksum)` has the arguments swapped (`assertEquals` takes `(expected, actual)`), so a mismatch labels the two values backwards.
**13. `S3ChecksumTest.kt:104`** - FYI rather than a request: `isOneShot = true` makes this request non-retryable (`RetryMiddleware.kt:144` is `isRetryable get() = !body.isOneShot`, and it branches to a single attempt when false), so a transient S3 503 now fails the test instead of retrying. Defensible, since real streaming uploads are one-shot too. If you would rather keep retries, `isOneShot = false` with a payload over 1 MiB satisfies the *other* disjunct and additionally exercises multi-chunk framing instead of a single 11-byte chunk.

---

### Suggested triage

Items 1, 2, 4, 5 and 9 remove every leak and the last dead file - roughly twenty lines total, and they are the only items with consequences that compound over time. Item 6 I would recommend on its merits, since it is what would have caught the two earlier misses. Item 3 belongs in its own ticket. Everything else is safe to drop if you would rather not churn a merge PR this size.

ianbotsf:
Thanks for the detailed analysis. I've addressed @ashishdhingra's items 1, 2, 4, 5, 6, 9, and 12, as well as @luigi617's `runId` race condition.

PR #1994, opened 2026-09-08
feat(auth): cache default-chain credentials with refresh resiliency

ashishdhingra:
## Issue \#

No public GitHub issue. Tracked internally as **SDK-KT-1223**, implementing phase 1 of the cross-SDK
credential-refresh design specification (internal document — reviewers on the team have the link; ask if you need it).

**Depends on `smithy-kotlin`:** every new type used here is added by the companion PR
smithy-lang/smithy-kotlin#1714, on that repo's `feature/credential-refresh-resiliency-phase1` branch.

> [!IMPORTANT]
> **Release sequencing — this PR is not mergeable on its own.**
>
> 1. Merge smithy-lang/smithy-kotlin#1714.
> 2. **Release smithy-kotlin manually** — the changes in this PR reference types that only exist in that release.
> 3. **Add a commit to this PR** bumping `smithy-kotlin-version` in `gradle/libs.versions.toml` from `1.7.11` to the
>    released version, with a `.changes` entry alongside it (see `chore: version bump smithy-kotlin-version to 1.7.6`
>    for the shape of both).
> 4. CI on this PR only goes green once that bump is in place; until then it fails resolving the new symbols.
>
> The bump is deliberately not in this PR yet, because the pin cannot be validated locally: this workspace resolves
> smithy-kotlin as a Gradle composite build, which substitutes the sibling project regardless of the declared version.
> Only CI exercises the published artifact.

### Why

The default credentials chain is wrapped in `CachedCredentialsProvider`, which refreshes on a fixed 10-second buffer
before expiration and has no handling for a refresh that fails — the call either succeeds or the customer's request
fails. A brief credential-service disruption therefore becomes a client-side outage even while the credentials in hand
are still valid for another 20 minutes, and once refreshes start failing the SDK re-attempts on every request, putting
the most load on a struggling credential source exactly when it can least absorb it.

This PR moves the default chain onto the refresh lifecycle the other AWS SDKs already implement: refresh early, keep
serving what we have when a refresh fails and it is safe to do so, rate-limit retries, and fail fast on errors that
retrying cannot fix.

### What changes for customers

Only the **default** chain. A customer who explicitly configured a provider gets exactly today's behavior; nothing here
wraps or rewrites a caller-supplied provider.

| | Before | After |
|---|---|---|
| Refresh timing | 10s before expiration | advisory window 5 / 15 / 60 min by lifetime; mandatory at 1 min |
| Refresh fails, credentials still valid | request fails | cached credentials served |
| Refresh fails, credentials expired | request fails | served if the source is AWS-managed (static stability); otherwise fails |
| After a failed refresh | retried on every request | no attempt for a jittered 300–600s |
| Unrecoverable error (bad profile, expired SSO token) | absorbed, retried | propagated immediately |
| Service rejects the credentials | served again until scheduled refresh | refreshed on the next resolution |

One leaf-visible change: `ProcessCredentialsProvider` no longer reports `Instant.MAX_VALUE` as the expiration for a
credential process that reports none. A standalone caller reading `Credentials.expiration` now sees `null`. That value
was wrong — an expiration in the year 275760 suppressed refresh entirely — and it has its own `bugfix` changelog entry.

### The changes

**1. The chain switch** — `DefaultChainCredentialsProvider` now wraps its chain in
`ResilientCachingCredentialsProvider` instead of `CachedCredentialsProvider`, and implements
`RefreshAwareCredentialsProvider` so that a generated client wrapping this provider cannot stack a second cache on it.
Its KDoc is updated to match.

`close()` intentionally does **not** also call `chain.close()`: `ResilientCachingCredentialsProvider.close()` already
calls `source.closeIfCloseable()`, so adding it would double-close the chain.

**2. Refresh-behavior declarations** — each provider states how its credentials may be refreshed, which is what lets one
cache in front of the whole chain still scope behavior per provider:

| Declaration | Providers |
|---|---|
| `RefreshableWithStaticStability` | ECS, IMDS, SSO, Login, STS assume-role, STS web-identity |
| `NonRefreshable` | environment, JVM system property, static (builder path only) |
| *nothing* → `CachingOnly` | credential process, and any customer-provided provider |

`StaticCredentialsProvider` declares `NonRefreshable` **only on the builder path**, not on the primary constructor that
accepts a caller's `Credentials` — a customer's own object is returned exactly as given rather than having attributes
added to it.

**3. Non-recoverable error marking** — errors that cannot succeed on retry are marked so the cache propagates them
immediately instead of treating the failure as "keep using the cache":

- `ProviderConfigurationException` and the invalid SSO / login token exceptions are marked in an `init` block. Doing it
  there rather than via a new supertype is what keeps these classes out of the `.api` diff.
- The non-recoverable STS error codes are marked at the throw site in the two STS providers, and
  `UnauthorizedException` in `SsoCredentialsProvider`, via an internal `nonRecoverable()` helper.

Note one deviation from the design document: it called for marking 3 of the 4 `OAuth2ErrorCode` branches in
`LoginTokenProvider` and leaving `AuthcodeExpired` recoverable. In the actual code all four branches funnel through
`throwTokenExpired(...)` → `InvalidLoginTokenException`, so the `init` block covers all four, and `AuthcodeExpired`
becomes non-recoverable too. That is the right outcome — the exception's own message is "Please reauthenticate using
'aws login'", and no in-process retry can obtain a fresh authorization code — so the per-branch edit the document
sketched is not needed at all.

**4. Credential invalidation** — `CredentialsInvalidationInterceptor` (new, registered on every generated client via
`CredentialsInvalidationMiddleware`) tells the identity provider when a target service rejected the credentials that
signed the request, so the next resolution refreshes rather than serving them again.

- Matches `ExpiredToken`, `ExpiredTokenException`, `InvalidToken`. **`AccessDenied` is deliberately excluded** — it
  usually means the credentials are valid and the principal lacks permission, so refreshing would not help.
- Reads the code from the deserialized exception, not a response header: `x-amzn-ErrorType` is not populated for all
  protocols, and query/XML services carry the code in the body.
- Does not modify the response and does not retry the failed request; the refresh happens on the next resolution.
- A failure inside `invalidate` is logged and swallowed, so invalidation bookkeeping can never replace the service error
  the caller needs to see.

**5. The IMDS gate** — `ImdsCredentialsProvider` skips its own refresh lifecycle when the new
`CallerOwnsCredentialsRefresh` attribute says a cache above it owns one, and keeps every line of its current behavior
when held directly. `resolve` splits into `resolveUncached` and `resolveSelfPaced` so that phase 2 deletes a function
rather than untangling a body; the two do not share a fetch because today's lifecycle is interleaved with the requests
at five separate points.

### API compatibility

`aws-config.api` is **+39 / −1** and additive. The single deletion is `DefaultChainCredentialsProvider`'s class
declaration line being rewritten to add the `RefreshAwareCredentialsProvider` supertype. The bulk of the additions are
synthetic `invalidate` bridge methods that appear on every provider implementing `IdentityProvider` / `TokenProvider`,
a consequence of the new default method upstream rather than of anything written here. `apiCheck` passes.

### Testing

`./gradlew :aws-runtime:aws-config:jvmTest :codegen:aws-sdk-codegen:test :aws-runtime:aws-config:apiCheck` passes
(209 tests), as does `./gradlew ktlint`.

**Codegen validated end to end, not just compiled:** bootstrapped STS from a deleted `generated-src`
(`:codegen:sdk:bootstrap -Paws.services=+sts`), confirmed the interceptor registers at all 11 operations with the
import present, then compiled the generated client (`:services:sts:compileKotlinJvm`). This is the check that would
catch a wrong runtime-type symbol or namespace.

**12 new tests:** 7 for the interceptor (including the `AccessDenied` non-invalidation and the swallowed-failure path),
3 for the IMDS gate, 2 for the chain.

**A note on the ~46 test diffs.** Adding a refresh-behavior declaration puts a new entry in `Credentials.attributes`,
which participates in `CredentialsImpl` data-class equality, so every `assertEquals` against an expected `Credentials`
had to state it too. Rather than hand-editing ~30 attribute literals, two helpers in `CredentialsTestUtils.kt`
(`withRefreshBehavior` / `withoutRefreshBehavior`) build on the existing public `Credentials.copy`.

`DefaultChainCredentialsProviderTest` is the exception: its `test-case.json` fixtures are shared with other AWS SDKs
and do not model refresh behavior, and which provider wins (hence which declaration appears) varies per case. Rather
than edit shared fixtures or encode a per-case map, the attribute is sanitized off the *actual* value — mirroring the
`providerName` / `expiration` sanitization already there — and a dedicated `testRefreshBehaviorSurvivesTheChain`
asserts the declaration reaches the caller end to end.

#### Local testing (using test app)
- Uploaded test application [KotlinCredentialsRefreshTest.zip](https://github.com/user-attachments/files/32020148/KotlinCredentialsRefreshTest.zip) which publishes changes to `mavenLocal` (refer `Readme.md` for details)
- Executed tests locally:
  ```
  Picked up JAVA_TOOL_OPTIONS: -Dlog4j2.formatMsgNoLookups=true

  > Task :run
  Picked up JAVA_TOOL_OPTIONS: -Dlog4j2.formatMsgNoLookups=true

  ====================================================================================================
    Scenario 1: the default chain is refresh-resilient
  ====================================================================================================

  -- resolve through the default chain
  [main] [TRACE] ResilientCachingCredentialsProvider - refreshed credentials from SystemProperties; expiration=none, advisory=2026-09-09T17:45:49.968604Z, mandatory=none
      providerName      = SystemProperties
      accessKeyId       = AKIDFROMSYSTEMPROPERTY
      expiration        = none
      refreshBehavior   = NonRefreshable
    [PASS] the chain declares itself refresh-aware, so a client cannot stack a second cache on it
    [PASS] the system property provider declared NonRefreshable  <was NonRefreshable>

  -- resolve again
    [PASS] the second resolve was served from the cache, not re-resolved  <same object identity>

  ====================================================================================================
    Scenario 2: refresh timing, static stability, backoff, and error classification
  ====================================================================================================

  -- an AWS-managed source (declares RefreshableWithStaticStability), 15m lifetime, 5m backoff
  [main] [TRACE] ResilientCachingCredentialsProvider - refreshed credentials from FakeImds; expiration=2026-09-09T17:59:50.262040Z, advisory=2026-09-09T17:54:50.262040Z, mandatory=2026-09-09T17:58:50.262040Z
    [PASS] T+0m  initial resolve went to the source once  <calls=1>
    [PASS] T+0m  the source saw CallerOwnsCredentialsRefresh, so a self-pacing provider resolves straight through
    [PASS] T+1m  outside the advisory window, no refresh attempted  <calls=1>
  [main] [WARN] ResilientCachingCredentialsProvider - Credential refresh failed: simulated credential service outage. The SDK will continue using cached credentials. A refresh of these credentials will be attempted again after 300 seconds.
  FakeOutage: simulated credential service outage
    [PASS] T+10m advisory refresh was attempted  <calls=2>
    [PASS] T+10m the failed advisory refresh still served the cached credentials
    [PASS] T+11m the 5m backoff suppressed a second attempt (today's cache would retry on every request)  <calls=2>
  [main] [WARN] ResilientCachingCredentialsProvider - Credential refresh failed: simulated credential service outage. The SDK will continue using cached credentials. A refresh of these credentials will be attempted again after 300 seconds.
  FakeOutage: simulated credential service outage
    [PASS] T+21m static stability served credentials that expired at T+15m
  [main] [TRACE] ResilientCachingCredentialsProvider - refreshed credentials from FakeImds; expiration=2026-09-09T18:26:50.262040Z, advisory=2026-09-09T18:21:50.262040Z, mandatory=2026-09-09T18:25:50.262040Z
    [PASS] T+27m the source recovered and the credentials were replaced  <AKID-1 -> AKID-4>

  -- a provider that declares nothing (credential process, custom provider): cached, but never static-stable
  [main] [TRACE] ResilientCachingCredentialsProvider - refreshed credentials from CustomProvider; expiration=2026-09-09T17:59:50.280842Z, advisory=2026-09-09T17:54:50.280842Z, mandatory=2026-09-09T17:58:50.280842Z
  [main] [WARN] ResilientCachingCredentialsProvider - Credential refresh failed: simulated credential source outage. The SDK will continue using cached credentials. A refresh of these credentials will be attempted again after 60 seconds.
  FakeOutage: simulated credential source outage
    [PASS] T+16m expired credentials from an undeclared source are NOT served; the error is raised  <Credential refresh failed and the cached credentials for CustomProvider are expired>

  -- a non-recoverable error (malformed profile, expired SSO token) propagates instead of being absorbed
  [main] [TRACE] ResilientCachingCredentialsProvider - refreshed credentials from FakeSso; expiration=2026-09-09T17:59:50.283055Z, advisory=2026-09-09T17:54:50.283055Z, mandatory=2026-09-09T17:58:50.283055Z
    [PASS] T+11m the error reached the caller even though usable credentials were cached  <the SSO token is expired; run 'aws sso login'>
    [PASS] the error is briefly cached, so a retry loop does not hammer the source  <calls still 2>
  [main] [TRACE] ResilientCachingCredentialsProvider - refreshed credentials from FakeSso; expiration=2026-09-09T18:10:54.283055Z, advisory=2026-09-09T18:05:54.283055Z, mandatory=2026-09-09T18:09:54.283055Z
    [PASS] once the error TTL lapses the cache tries again

  -- invalidate(): a rejected credential is refreshed on the next resolution
  [main] [TRACE] ResilientCachingCredentialsProvider - refreshed credentials from FakeSts; expiration=2026-09-09T18:44:50.285190Z, advisory=2026-09-09T18:29:50.285190Z, mandatory=2026-09-09T18:43:50.285190Z
    [PASS] with 59m left, a plain resolve does not refresh  <calls=1>
  [main] [TRACE] ResilientCachingCredentialsProvider - credentials were rejected by the target service; a refresh is due on the next resolution
  [main] [TRACE] ResilientCachingCredentialsProvider - refreshed credentials from FakeSts; expiration=2026-09-09T18:45:50.285190Z, advisory=2026-09-09T18:30:50.285190Z, mandatory=2026-09-09T18:44:50.285190Z
    [PASS] after the rejection the next resolve refreshed, 58m before expiry  <AKID-1 -> AKID-2>
  [main] [TRACE] ResilientCachingCredentialsProvider - credentials were rejected by the target service; a refresh is due on the next resolution
    [PASS] a rejection naming already-replaced credentials is ignored  <still AKID-2>

  ====================================================================================================
    Scenario 3: a service rejection reaches the cache through a generated STS client
  ====================================================================================================

  -- the service returns ExpiredTokenException
  [main] [TRACE] ResilientCachingCredentialsProvider - refreshed credentials from FakeSource; expiration=2026-09-09T18:44:50.289137Z, advisory=2026-09-09T18:29:50.289137Z, mandatory=2026-09-09T18:43:50.289137Z
  [main] [TRACE] ResilientCachingCredentialsProvider - rpc=STS.GetCallerIdentity sdkInvocationId=4d3a00b1-dd51-416e-ad9f-ce61aa664f53 credentials were rejected by the target service; a refresh is due on the next resolution
  [main] [TRACE] ResilientCachingCredentialsProvider - refreshed credentials from FakeSource; expiration=2026-09-09T18:44:50.289137Z, advisory=2026-09-09T18:29:50.289137Z, mandatory=2026-09-09T18:43:50.289137Z
      client threw StsException (errorCode=ExpiredTokenException)
    [PASS] the rejected credentials were replaced on the next resolution  <AKID-1 -> AKID-2>
    [PASS] the source was called again  <calls=2>

  -- the service returns AccessDenied (deliberately not an invalidation trigger)
  [main] [TRACE] ResilientCachingCredentialsProvider - refreshed credentials from FakeSource; expiration=2026-09-09T18:44:50.650659Z, advisory=2026-09-09T18:29:50.650659Z, mandatory=2026-09-09T18:43:50.650659Z
      client threw StsException (errorCode=AccessDenied)
    [PASS] AccessDenied did NOT invalidate: the credentials are valid, the principal just lacks permission  <still AKID-1>
    [PASS] the source was not called again  <calls=1>

  ====================================================================================================
    22 checks, 0 failed
  ====================================================================================================

  BUILD SUCCESSFUL in 3s
  ```

### Suggested reading order

1. `DefaultChainCredentialsProvider.kt` — the switch itself, and the whole customer-visible surface of this PR.
2. The behavior declarations — mechanical, but worth confirming each provider's classification.
3. `internal/NonRecoverableErrors.kt` and the marking sites.
4. `CredentialsInvalidationInterceptor.kt` and its test, then the codegen registration.
5. `ImdsCredentialsProvider.kt` — the gate, and the two resolve paths.

### Deliberately not in this PR

Phase 2, on the next minor line: providers adopting `SelfManagedRefresh` so they cache themselves; four providers
gaining `close()`; `ProfileCredentialsProvider`; deleting IMDS's bespoke lifecycle; and the KDoc rewrites that currently
tell customers to hand-wrap with `CachedCredentialsProvider`. That last one cannot land yet — its replacement,
`resilientlyCached()`, is `@InternalApi`, so the rewritten text would point customers at a symbol they are told not to
call. Once the providers cache themselves, the replacement for hand-wrapping is *nothing*, and the text can say so.

### Changelog entries

Three, none requiring a minor version bump: two `feature` (the cache, and invalidation) and one `bugfix` (the process
expiration).
___


ashishdhingra:
[KotlinCredentialsRefreshTest.zip](https://github.com/user-attachments/files/32020148/KotlinCredentialsRefreshTest.zip)

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CredentialsInvalidationInterceptor.kt
```diff
@@ -0,0 +1,62 @@
...
+        if (identity != null && provider != null) {
+            try {
+                provider.invalidate(identity)
+            } catch (invalidationFailure: Exception) {
+                // never let invalidation bookkeeping replace the service error the caller needs to see
+                coroutineContext.trace<CredentialsInvalidationInterceptor> {
+                    "failed to invalidate rejected credentials: ${invalidationFailure.message}"
+                }
+            }
+        }
```
ianbotsf, line 64: **Nit**: I see you log a `TRACE` when we fail to invalidate credentials but we should never throw in `invalidate` by contract, right? So any exception thrown would indicate a bug. Catching it to prevent the bug from failing the user's request seems correct but we should log at `WARN` level instead to increase the chances the issue will be reported and fixed.

Additionally, log a `DEBUG` when we determine that `errorCode in INVALIDATION_ERROR_CODES`. That could be helpful for users debugging a failure scenario:

```kotlin
val errorCode = ex.sdkErrorMetadata.attributes.getOrNull(ServiceErrorMetadata.ErrorCode)
if (errorCode !in INVALIDATION_ERROR_CODES) return context.response

coroutineContext.debug<CredentialsInvalidationInterceptor> {
    "Credentials were rejected by the service with error code $errorCode"
}
```

ianbotsf (approved):

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/CredentialsInvalidationInterceptor.kt
ashishdhingra: Both taken in b531feeb.

`WARN` on the invalidation failure, and the throwable is now passed to the logger too so the stack trace survives — it was being dropped in favor of just `.message`, which would have made an actual bug here hard to chase. Wording of the comment updated to say why it is caught at all.

`DEBUG` line added right where you suggested, so it fires even when `ResolvedIdentity`/`ResolvedIdentityProvider` are absent and nothing ends up being invalidated — that distinction is exactly what someone debugging this would want to see.

One note on the contract: `IdentityProvider.invalidate` documents behavioral requirements and a no-op default, so nothing should throw — but it is a public interface method, so the implementation that throws may be a customer's rather than ours. `WARN` rather than `ERROR` felt right for that reason.

PR #1995, opened 2026-09-08
fix(dynamodb-mapper): restore support for nested types in annotation-driven schemas

ianbotsf:
## Issue \#

(none)

## Description of changes

This PR re-adds support for nested classes in generated schemas. Significant changes:
* Introduce a new annotation for nested types: `@DynamoDbMappable`. This annotation conveys the semantic meaning "generate an `ItemConverter` for this class".
* Apply `@DynamoDbMappable` as a meta-annotation to `@DynamoDbItem`. Now, `@DynamoDbItem` inherits the semantic meaning "generate an `ItemConverter` for this class" from `@DynamoDbMappable` and adds additional semantic meaning "enforce primary key requirements and generate an `ItemSchema` for this class".
* Rename `ItemToValueConverter` codegen symbol declaration to `ItemValueConverter`, commensurate with the actual runtime type.
* Explicitly detect and reject cycles in nested types (e.g., `data class Node(val children: List<Node>)`). We want to support these eventually but that's a larger change than supporting basic (non-cyclic) nesting.
* Add new schema-generator unit tests to verify new behavior.


hll/ddb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/HighLevelRenderer.kt
```diff
@@ -47,22 +64,66 @@ internal class HighLevelRenderer(
...
+     */
+    private fun findCyclicallyNestedClasses(): Set<String> {
+        val classesByName = annotatedClasses.associateBy { requireNotNull(it.qualifiedName).asString() }
+
+        // Build adjacency limited to edges between annotated (mappable) types.
+        val adjacency = classesByName.mapValues { (_, decl) ->
+            decl.getAllProperties()
+                .flatMap { it.type.resolve().referencedTypeNames() }
+                .filter { it in classesByName }
+                .toSet()
```
luigi617: Correctness: Here we don't exclude `private` and `@DynamoDbIgnore` properties like in _SchemaRenderer.kt:57-60_. This falsely rejecting valid schemas.

hll/ddb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -80,13 +79,30 @@ internal class SchemaRenderer(
...
             .annotations
-            .single { it.annotationType.resolve().declaration.qualifiedName?.asString() == DynamoDbItem::class.qualifiedName }
-            .arguments
-            .singleOrNull()
+            .firstOrNull { it.annotationType.resolve().declaration.qualifiedName?.asString() == markerQualifiedName }
+            ?.arguments
+            ?.singleOrNull()
             ?.value
-            ?.let { it as KSType }
+            ?.let { it as? KSType }
```
luigi617: In what case the argument is not `KSType`? I feel like if there is one, it must be `KSType`.

luigi617:
Missing unit tests:
- acyclic `@DynamoDbItem` nested inside another `@DynamoDbItem`
- explicit `converter = ...` on `@DynamoDbMappable`
- `@DynamoDbTtlSeconds` / `@DynamoDbCounter` on a nested type
- deep nesting such as `List<Map<String, Address>>`

hll/ddb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -80,13 +79,30 @@ internal class SchemaRenderer(
...
 
     init {
+        // @DynamoDbMappable targets ANNOTATION_CLASS so it can meta-annotate @DynamoDbItem, which means the Kotlin
+        // compiler alone won't stop a user from incorrectly applying it to their own annotation class. Fail fast with a
+        // located diagnostic in that case.
+        if (classDeclaration.classKind == ClassKind.ANNOTATION_CLASS) {
+            val msg = "@DynamoDbMappable cannot be applied to annotation classes " +
+                "(${classDeclaration.qualifiedName?.asString()}); it is only valid on regular or data classes."
+            ctx.logger.error(msg, classDeclaration)
+            error(msg)
```
luigi617, line 86: This throws and the `SchemaRenderer` is used `HighLevelRenderer.render()`'s `forEach` and nothing catches it. And the `IllegalStateException` escapes `SymbolProcessor.process()`. KSP treats an exception out of a processor as a processor crash rather than a diagnostic: it aborts the round, reports it as an internal KSP error with a stack trace, and never returns to the `forEach`.

luigi617 (changes requested):

hll/ddb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/HighLevelRenderer.kt
ianbotsf: You're right, I'll centralize the logic for deriving mapped properties and delegate from `SchemaRenderer` and `HighLevelRenderer`.

hll/ddb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
ianbotsf: I forget what I was thinking here...reverting to `as KSType`.

hll/ddb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
ianbotsf: Throwing an uncaught exception is how we've historically handled critical errors in user annotations within KSP (this file is replete with examples). I can see now that logging the error but continuing to process as much as possible provides better info to the user but I'd like to scope that change to a dedicated PR, not this one.

ianbotsf:
> Missing unit tests:
> 
> * acyclic `@DynamoDbItem` nested inside another `@DynamoDbItem`
> * explicit `converter = ...` on `@DynamoDbMappable`
> * `@DynamoDbTtlSeconds` / `@DynamoDbCounter` on a nested type
> * deep nesting such as `List<Map<String, Address>>`

Great suggestions, added tests for each.

hll/ddb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/HighLevelRenderer.kt
```diff
@@ -24,9 +26,24 @@ internal class HighLevelRenderer(
...
+        val cyclicClasses = findCyclicallyNestedClasses()
+
         annotatedClasses.forEach { annotated ->
             logger.info("Processing annotation on ${annotated.simpleName}")
 
+            val qualifiedName = annotated.qualifiedName?.asString()
+            if (qualifiedName != null && qualifiedName in cyclicClasses) {
+                logger.error(
+                    "Cyclic nesting detected involving type '$qualifiedName'. Nesting " +
+                        "@DynamoDbItem/@DynamoDbMappabletypes in a reference cycle (including self-references) is " +
```
luigi617: Nit: missing a space between "@DynamoDbMappable" and "types"

hll/ddb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -225,8 +283,23 @@ internal class SchemaRenderer(
...
+                if (decl.containingFile == null) {
+                    ctx.logger.error(
+                        "Nested type '${decl.qualifiedName?.asString()}' is defined in another module. Nesting " +
+                            "@DynamoDbMappable types across modules is not yet supported; the nested type must be " +
+                            "declared in the same module as the type that references it.",
+                        decl,
+                    )
+                }
+
+                writeInline("#T", TypeRef(nestedConverterPackage(decl), "${decl.simpleName.asString()}ValueConverter"))
```
luigi617, line 302: I guess this might be intentional to make the code clearer, but just want to confirm in case that `decl.containingFile == null`, the `writeInline` below is essentially useless because the build will eventually fail, right?

luigi617, line 302: Question: I don't fully understand an edge case. Let's say we have `A` that has property of `B`, then `B` has `B` as property. My understanding is `BValueConverter` is skipped because we currently do not support self-cycle. But when we handle `A`, `BValueConverter` is being used which will be raise an `unresolved reference` compiler error. Did you handle it somewhere?

hll/ddb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
```diff
@@ -80,11 +76,28 @@ internal class SchemaRenderer(
...
+            ctx.logger.error(msg, classDeclaration)
+            error(msg)
+        }
+
+        // Read the user-specified converter (if any) from whichever marker annotation applies to this type.
+        val markerQualifiedName = if (isTableItem) {
+            DynamoDbItem::class.qualifiedName
+        } else {
+            DynamoDbMappable::class.qualifiedName
+        }
```
luigi617, line 94: Nit: I think we already have unit test for `DynamoDbItem`, but not for `DynamoDbMappable`. A unit test similar to `testDynamoDbItemConverter` would help

luigi617 (changes requested):

hll/ddb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
ianbotsf: Yes, we can probably short-circuit out of the method after determining `decl.containingFile == null`. Note that even so, we may have already written lines to this same file which will also be essentially useless.

ianbotsf: In this scenario, we'll have logged an `ERROR` message which will cause the KSP task to fail. The generated code won't compile but that's okay because the whole build will fail until the user fixes their annotations.

hll/ddb-mapper/dynamodb-mapper-schema-codegen/src/main/kotlin/aws/sdk/kotlin/hll/dynamodbmapper/codegen/annotations/rendering/SchemaRenderer.kt
ianbotsf: `testNestedItems` already verifies the creation of an `ItemConverter` for a `@DynamoDbMappable` type. Are there any improvements you can suggest to that test?

luigi617: oh you're right

luigi617 (approved):

PR #1997, opened 2026-09-09
Add AWS_LOGIN_CACHE_DIRECTORY env var support

### LINKED ISSUE #1993: Login Credentials Provider uses the wrong cache env var
### Describe the bug

The Login Credentials Provider does not respect the `AWS_LOGIN_CACHE_DIRECTORY` environment variable to set the location of the login cache, like other SDKs. Instead, it uses `AWS_LOGIN_IN_CACHE_DIRECTORY` variable (with an extra `IN_`).

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected behavior

When I set the `AWS_LOGIN_CACHE_DIRECTORY` env var, the Kotlin SDK should read the loaded credentials that other SDKs put there, and put credentials there for other SDKs to read.

### Current behavior

Credentials get written to and read from the default `~/.aws/login/cache` location instead.

### Steps to Reproduce

1. Point `AWS_LOGIN_CACHE_DIRECTORY` at a scratch directory:
`export AWS_LOGIN_CACHE_DIRECTORY=/tmp/login-cache`
2. Write a valid, non-expired login token in that directory. The filename is `sha256(<login session name>) + '.json'`. For `my-login-session`, that's `85a19150f6728d225f1da22f304c53c9f1ca13d715083b33bc8df22081280d15.json`. An example token:
```json
{
  "accessToken": {
    "accessKeyId": "AKIDEXAMPLE",
    "secretAccessKey": "examplesecret",
    "sessionToken": "examplesessiontoken",
    "expiresAt": "2999-01-01T00:00:00Z",
    "accountId": "012345678910"
  },
  "refreshToken": "examplerefresh",
  "clientId": "exampleclient",
  "dpopKey": "unused-because-the-token-is-not-expired"
}
```
3. Resolve credentials for that session:
```kotlin
val provider = LoginCredentialsProvider(loginSession = "my-login-session", region = "us-east-1")
provider.use { runBlocking { println(it.resolve(emptyAttributes()).accessKeyId) } }
```

The result is a `ProviderConfigurationException`. Do the same, but with `AWS_LOGIN_IN_CACHE_DIRECTORY`, and it works.

### Possible Solution

Support the `AWS_LOGIN_CACHE_DIRECTORY` env var in addition to `AWS_LOGIN_IN_CACHE_DIRECTORY`.

### Context

_No response_

### AWS SDK for Kotlin version

1.8.45

### Platform (JVM/JS/Native)

JVM

### Operating system and version

macOS 26.6.2

luigi617:
## Issue \#

Resolve https://github.com/aws/aws-sdk-kotlin/issues/1993

## Description of changes

Add AWS_LOGIN_CACHE_DIRECTORY env var support per https://docs.aws.amazon.com/sdkref/latest/guide/feature-login-credentials.html


aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/LoginCredentialsProvider.kt
```diff
@@ -86,7 +86,7 @@ public class LoginCredentialsProvider public constructor(
     }
 }
 
-internal fun resolveCacheDir(platformProvider: PlatformProvider) = platformProvider.getenv("AWS_LOGIN_IN_CACHE_DIRECTORY") ?: platformProvider.filepath("~", ".aws", "login", "cache")
+internal fun resolveCacheDir(platformProvider: PlatformProvider) = platformProvider.getenv("AWS_LOGIN_CACHE_DIRECTORY") ?: platformProvider.getenv("AWS_LOGIN_IN_CACHE_DIRECTORY") ?: platformProvider.filepath("~", ".aws", "login", "cache")
```
ashishdhingra, line 89: I guess old variable is still supported for backwards compatibility so that existing users are not affected.

ashishdhingra (approved):

ashishdhingra:
Please add change log file.

aws-runtime/aws-config/common/src/aws/sdk/kotlin/runtime/auth/credentials/LoginCredentialsProvider.kt
luigi617: yes, correct

PR #1998, opened 2026-09-10
feat(telemetry-cloudwatch): publish SDK metrics to CloudWatch

ashishdhingra:
## Issue \#

Related to aws/aws-sdk-kotlin#984 - CloudWatch `MetricsProvider`. Referenced rather than auto-closing, so the issue stays open until both halves ship.

## Description of changes

Adds `telemetry-cloudwatch`, a `CloudWatchMetricExporter` that publishes SDK operational metrics - call duration, retry counts, error rates - to CloudWatch with `PutMetricData`. Over the wire rather than through EMF, so it needs no CloudWatch agent, no OpenTelemetry collector, and no ambient log ingestion: it works on EC2 without the agent, on-premises, in containers, on desktop JVM, on Android, and on Kotlin/Native. The backend-agnostic aggregation half is smithy-lang/smithy-kotlin#1716; this module converts and publishes.

- Maps aggregated values to `MetricDatum`: scalars to `Value`, histograms to a `StatisticSet`, or to `Values`/`Counts` pairs when percentiles are wanted.
- Chunks datums to CloudWatch's per-request limits and bounds requests per cycle with `maxCallsPerUpload`. Excess datums are dropped with a warning naming the knob that fixes it, rather than silently raising the cap.
- Forces telemetry off on the publishing client - otherwise each `PutMetricData` emits metrics of its own, collected next cycle, producing another `PutMetricData`: a loop that does not decay and bills for every turn.
- Always owns the client it publishes with. A caller-supplied client is copied via `withConfig` and the copy closed on shutdown, leaving the original theirs.
- Defaults follow the Java v2 `CloudWatchMetricPublisher` - summary statistics for every metric, percentiles opt-in per instrument, `rpc.service` and `rpc.method` as the default dimensions.
- Built only when `:services:cloudwatch` is bootstrapped, following `:hll:ddb-mapper` on `:services:dynamodb`, since it depends on a generated service client.

### Merge ordering

**smithy-lang/smithy-kotlin#1716 must merge and release first.** `gradle/libs.versions.toml` here references `aws.smithy.kotlin:telemetry-metrics-aggregation` at the current `smithy-kotlin-version` (1.7.9), which predates the module, so CI cannot resolve it until smithy-kotlin 1.7.11 ships and the catalog is bumped. Local builds resolve it today only through the sibling `../smithy-kotlin` composite build.

### Testing

`:aws-runtime:telemetry-cloudwatch:build` passes on JVM and Kotlin/Native (mingwX64) - 10 tests per target, plus `apiCheck` against the committed baseline.

#### Local testing

[CloudWatchMatricsProviderTest.zip](https://github.com/user-attachments/files/32130653/CloudWatchMatricsProviderTest.zip) consumes both branches from source through a Gradle composite build; see its `README.md`. The default mode is offline and makes no billable calls. The billable modes published to `Test/CloudWatchMetricsProvider` in `us-west-2`, read back with `get-metric-statistics` and `list-metrics` to confirm what landed rather than what was logged:

| Behaviour | Verified |
|---|---|
| Delta temporality | `Calls` recorded 3 + 3 over two cycles returned `Sum=6, SampleCount=2`, not the `9` a cumulative counter would give |
| Gauges are absolute | `QueueDepth` readings of 7 then 3 returned `Max=7.0, Min=3.0`; over a 1-minute periodic run it traced a ramp (`Max=15.0, Min=0.0`) rather than oscillating around zero |
| Non-monotonic counters | `InFlight`, five `+1`/`-1` pairs, returned `Sum=0.0` |
| `detailedMetrics` -> percentiles | `Latency` returned `p50=24.36`, `p99=37.98`; unavailable with the default `StatisticSet` |
| Unit mapping | `ms` published as `Unit=Milliseconds` |
| Recursion guard | `list-metrics` on dimension `rpc.service=CloudWatch` returns 0 series, so the exporter's own calls cannot feed back |
| Dimension allowlist | An attribute outside `dimensions`, varied independently of a listed one, was dropped and its series merged rather than leaking into billable cardinality |
| SDK's own instrumentation | Real `S3Client` traffic produced 22 instruments per cycle, including the engine's four async series - which appear only when `httpClient { telemetryProvider = ... }` is set as well |
| Periodic reader | 5 cycles at a 10s interval, 26 datums in 1 request each, plus the final flush on `close()` |

Two findings went into the module README. The SDK's async instruments live on the HTTP engine, which reads `telemetryProvider` from its *own* config, so setting it on the client alone drops them silently. And the closing flush publishes 19 datums rather than 26 because `gauge.stop()` and `s3.close()` run before `provider.use { }` exits - collect-time sampling, not a defect.

Async instruments were exercised on JVM only; `mingwX64` coverage is the unit tests above.
___


ashishdhingra:
[CloudWatchMatricsProviderTest.zip](https://github.com/user-attachments/files/32130653/CloudWatchMatricsProviderTest.zip)

.changes/6df57a30-d612-4b6c-b5f9-d40799f711b0.json
```diff
@@ -0,0 +1,9 @@
+{
+    "id": "6df57a30-d612-4b6c-b5f9-d40799f711b0",
+    "type": "feature",
+    "description": "Add `telemetry-cloudwatch`, a `CloudWatchMetricExporter` that publishes SDK metrics to Amazon CloudWatch with `PutMetricData`. Requires no OpenTelemetry collector and no CloudWatch agent",
```
ianbotsf, line 4: **Correctness**:

> Add `telemetry-cloudwatch`, a `CloudWatchMetricExporter` that...

`telemetry-cloudwatch` is not "a `CloudWatchMetricExporter`", it's a module that provides a CloudWatch telemetry implementation

> that publishes SDK metrics to Amazon CloudWatch with `PutMetricData`

Users don't care about the backing API used to get their metrics to CloudWatch, that's an inner implementation detail

> Requires no OpenTelemetry collector and no CloudWatch agent

Reads like an advertisement. As requested in the other PR, please tune your agent to be more succinct and less  florid.

aws-runtime/telemetry-cloudwatch/common/src/aws/sdk/kotlin/runtime/telemetry/cloudwatch/CloudWatchMetricExporter.kt
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    public companion object {
+        /**
+         * Default namespace. Slash-separated to group under a vendor prefix in the console, matching
+         * the convention AWS services use (`AWS/EC2`, `AWS/Lambda`). Not `AWS/...` — that prefix is
+         * reserved for service-published metrics, and reusing it would mix SDK client-side metrics
+         * in with the service's own server-side view of the same calls.
+         */
+        internal const val DEFAULT_NAMESPACE = "AwsSdk/KotlinSdk"
```
ianbotsf, line 197: **Question**: Why would any user want their metrics namespaced under "AwsSdk/KotlinSdk"? That seems like a sensible default for SDK developers but not for SDK users. What do other SDKs use for a default namespace?

aws-runtime/telemetry-cloudwatch/common/src/aws/sdk/kotlin/runtime/telemetry/cloudwatch/CloudWatchMetricExporter.kt
```diff
@@ -0,0 +1,314 @@
...
+     * Throws rather than logs, unlike every other error path in this class. The distinction: a failed
+     * publish loses one interval of data, whereas a shared exporter produces numbers that are
+     * *wrong but plausible* indefinitely. The former is worth surviving, the latter is not.
+     */
+    override fun onAttach() {
+        check(attached.compareAndSet(expect = false, update = true)) {
+            "CloudWatchMetricExporter instances cannot be shared between SdkTelemetryProvider " +
+                "instances. Construct one exporter per provider."
+        }
+    }
```
ianbotsf, line 119: **Question**: I don't see the point of this attachment mechanism. The implementing class takes no special action other than checking it hasn't already been attached, but has no idea what it's been attached to and, as far as I can see, nothing in this class cares whether it's been attached or not. For instance, `export` doesn't check whether it's attached, it simply exports. Why does this multi-attachment functionality belong in `CloudWatchMetricExporter` or even in `MetricExporter`s in general?

aws-runtime/telemetry-cloudwatch/common/src/aws/sdk/kotlin/runtime/telemetry/cloudwatch/CloudWatchMetricExporter.kt
```diff
@@ -0,0 +1,314 @@
...
+         */
+        internal fun validate() {
+            require(namespace.isNotEmpty() && namespace.length <= CloudWatchLimits.MAX_NAMESPACE_LENGTH) {
+                "namespace must be 1..${CloudWatchLimits.MAX_NAMESPACE_LENGTH} characters, got ${namespace.length}"
+            }
+            require(maxCallsPerUpload > 0) { "maxCallsPerUpload must be positive" }
+            require(storageResolution == 1 || storageResolution == 60) {
+                "storageResolution must be 1 or 60, got $storageResolution"
+            }
+        }
```
ianbotsf, line 299: **Nit**: Validation of finalized values belongs in the built object (i.e., `CloudWatchMetricExporter`) rather than the builder. Move this to an `init` block.

ianbotsf, line 299: **Question**: Why are we validating service inputs here? Normally our stance is to validate only the basic information our SDK requires and then allow the service to enforce its own contract. This logic effectively duplicates the service's internal constraints and would have to be changed if CloudWatch introduced support for additional metric resolutions.

aws-runtime/telemetry-cloudwatch/common/src/aws/sdk/kotlin/runtime/telemetry/cloudwatch/MetricDatumChunker.kt
```diff
@@ -0,0 +1,188 @@
...
+     *
+     * CAUTION — the scope of this limit is unresolved. `MetricDatum.Values` describes it as 150 unique
+     * values "in each `PutMetricData` action", and `PutMetricData` as 150 "per metric with one
+     * `PutMetricData` request"; both phrasings bind the *request* rather than the individual datum,
+     * whereas [MetricDatumChunker.toDatums] splits across datums within one request. This needs an
+     * empirical check before release: if the limit binds the request, the chunking strategy changes.
+     *
+     * @see <a href="https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricDatum.html#ACW-Type-MetricDatum-Values">MetricDatum.Values</a>
+     */
+    const val MAX_VALUES_PER_DATUM = 150
```
ianbotsf, line 44: **Correctness**: Resolve the questions about the scope of this limit before shipping this PR.

ianbotsf (changes requested):

aws-runtime/telemetry-cloudwatch/common/src/aws/sdk/kotlin/runtime/telemetry/cloudwatch/CloudWatchMetricExporter.kt
ashishdhingra: Java v2 does. `CloudWatchMetricPublisher` defaults to `AwsSdk/JavaSdk2` — [declared here](https://github.com/aws/aws-sdk-java-v2/blob/c216cedfdf81ab009ed921078ad49647b53bbbbd/metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisher.java#L178) and [applied when the builder value is unset](https://github.com/aws/aws-sdk-java-v2/blob/c216cedfdf81ab009ed921078ad49647b53bbbbd/metric-publishers/cloudwatch-metric-publisher/src/main/java/software/amazon/awssdk/metrics/publishers/cloudwatch/CloudWatchMetricPublisher.java#L276). `AwsSdk/KotlinSdk` mirrors that.

Happy to drop the default and require an explicit namespace instead if you would rather users be forced to name their own. The KDoc currently only recommends setting one per application.

aws-runtime/telemetry-cloudwatch/common/src/aws/sdk/kotlin/runtime/telemetry/cloudwatch/CloudWatchMetricExporter.kt
ashishdhingra: Removed. `onAttach` is gone from `MetricExporter`, and the override plus the `attached` field are gone from here.

You're right that nothing reads the flag: it was written once at install and never consulted again, so it was a one-shot guard in the clothes of a lifecycle callback — and it sat on the interface, where every future exporter implementer has to read and dismiss it for a problem only this one exporter has.

**What it protected against:** constructing one exporter and handing it to two `AggregatingTelemetryProvider`s. The concrete failure is `shutdown()`. This exporter owns its CloudWatch client and closes it there; the call is idempotent but not reference-counted, so the first provider to close silences the other, which then logs a publish warning every cycle and publishes nothing for the rest of the process. Second-order, two collection loops offset in time reporting one gauge into one namespace read as instability rather than as two sources. The old KDoc's "adds together delta sums from separate aggregators" was the weakest part of the case and I've dropped it — each provider has its own registry, so a single `record()` was never double-counted, and CloudWatch merging two `StatisticSet`s is arithmetically fine.

**What we lose:** failing fast at configuration time, with the stack trace at the caller's own code. I looked for a way to keep that without the SPI method and didn't find a good one — a check inside `export` can't throw, since exporters are contractually non-throwing, so it could only log, and only on the first collection cycle; a module-level identity set of installed exporters would catch it for every exporter but is a static leak and awkward on native. Given the misuse requires deliberately sharing an exporter across two providers and its worst outcome is degraded metrics rather than wrong data in the caller's application, that trade seems right.

**The KDoc now carries it on both sides.** `MetricExporter`: "Install one exporter into one reader. An exporter that owns a transport closes it in `shutdown`, so a shared instance is silenced for the remaining readers as soon as the first one closes." `CloudWatchMetricExporter`: "Construct one per provider. This exporter owns its CloudWatch client and closes it in `shutdown`, so a shared instance stops publishing for every provider but the first to close."

The one test that covered the guard is removed; nothing else asserted attachment.

.changes/6df57a30-d612-4b6c-b5f9-d40799f711b0.json
ashishdhingra: Reworded to describe the module rather than advertise it: "Add `telemetry-cloudwatch`, a module that publishes SDK metrics to Amazon CloudWatch".

aws-runtime/telemetry-cloudwatch/common/src/aws/sdk/kotlin/runtime/telemetry/cloudwatch/CloudWatchMetricExporter.kt
ashishdhingra: Moved — `Builder.validate()` is gone and the exporter has `init { require(builder.maxCallsPerUpload > 0) { "maxCallsPerUpload must be positive" } }`.

ashishdhingra: Agreed — the namespace-length and `storageResolution` checks are removed. Duplicating the service contract only means changing this module the day CloudWatch supports another resolution.

aws-runtime/telemetry-cloudwatch/common/src/aws/sdk/kotlin/runtime/telemetry/cloudwatch/MetricDatumChunker.kt
ashishdhingra: Settled empirically against real CloudWatch (every payload used values `1..N`, so an exact `Sum` of `N(N+1)/2` proves nothing was dropped). The decision-relevant case: two datums of 150 values each sharing name, dimensions, and timestamp in one request came back as SampleCount 300 / Sum 45,150 — splitting within a request merges correctly, so the chunker stands as written.

Separately, single datums over the documented 150 were accepted rather than rejected — 151 (Sum 11,476), 200, 500, 1,000, and 1,500 (Sum 1,125,750) all round-tripped intact. That's undocumented tolerance, which is a reason to keep chunking at the documented limit, not to stop.

aws-runtime/telemetry-cloudwatch/common/src/aws/sdk/kotlin/runtime/telemetry/cloudwatch/CloudWatchMetricExporter.kt
ianbotsf: Oh wow. Well I guess that settles that then!

ianbotsf (approved):

